PR #1, opened 2016-11-08
Rewrite call_standardise()

lionel-:
and add some tools for pairlists.


R/call.R
```diff
@@ -49,9 +216,13 @@ call_new <- function(f, ..., .args = list()) {
...
 #' # Add an explicit missing argument
 #' call_modify(call, list(na.rm = quote(expr = )))
-call_modify <- function(call, new_args, env = parent.frame()) {
-  stopifnot(is.call(call), is.list(new_args))
-
+#'
+#' # If the call is missing, the parent frame is used instead.
+#' f <- function(bool = TRUE) call_modify(new_args = list(bool = FALSE))
+#' f()
+call_modify <- function(call = NULL, new_args, env = NULL) {
```
hadley: It's a bit weird to have an optional arg, then a required arg, then an optional arg.

R/call.R
```diff
@@ -64,13 +235,43 @@ call_modify <- function(call, new_args, env = parent.frame()) {
...
+#' call_fn_name(~foo$bar(baz))
+call_fn_name <- function(call = NULL) {
+  call <- call %||% call_frame(2)
+  call <- get_call(call)
+  stopifnot(is.call(call))
+  fn <- call[[1]]
 
-  f <- eval(call[[1]], env)
-  if (is.primitive(f)) return(call)
+  # Handle namespaced calls: foo::bar(), foo@bar(), foo$bar()
```
hadley: Why does that not trigger in other scenarios?

tests/testthat/test-call.R
```diff
@@ -42,3 +90,16 @@ test_that("new args replace old", {
...
+  f <- function() call_modify(new_args = list(bool = FALSE))
+  expect_identical(f(), quote(f(bool = FALSE)))
+})
+
+# Utils --------------------------------------------------------------
+
+test_that("call_fn_name() handles namespaced calls", {
+  expect_equal(call_fn_name(quote(foo::bar())), "bar")
+  expect_equal(call_fn_name(quote(foo$bar())), "bar")
+  expect_equal(call_fn_name(quote(foo::bar@baz())), "baz")
```
hadley: Handle `foo[["bar"]]` too?

R/call.R
```diff
@@ -25,11 +25,178 @@ call_new <- function(f, ..., .args = list()) {
...
+#'   \item Always report dotted arguments as such: \code{call(..1, ..2)}.
+#'         In comparison, \code{match.call()} inlines literals:
+#'         \code{call("foo", 10)}.
+#'   \item Provides an argument \code{enum_dots} to impose
+#'         enumerated names on dotted arguments. This produces
+#'         \code{call(..1 = x, ..2 = ..3)} instead of
+#'         \code{call(foo = x, ..3)}.
+#'   \item Does not sort arguments according to the order of
+#'         appearance in the function definition.
+#'   \item Does not partial-match arguments.
```
lionel-: oh right, this could create problems for arg introspection if we don't partial match.

R/call.R
hadley: Yeah, that's what I was thinking. It seems like you should explicitly test that `f` has length 3, and the first element is one of `@`, `$`, `::`, `:::` or `[[`.

R/call.R
hadley: Two options:

1. Added some special behaviour for when `call_modify()` is missing
2. Default `new_args` to `list()`.

R/call.R
```diff
@@ -25,11 +25,178 @@ call_new <- function(f, ..., .args = list()) {
...
+#'   arguments. If \code{TRUE}, this produces calls such as
+#'   \code{f(..1 = ..2)} instead of \code{f(..2)}. The first form is
+#'   mostly intended for code analysis tools as it makes it easier to
+#'   climb arguments through nested calls. The latter form (the
+#'   default) is more faithful to the actual call and is ready to be
+#'   evaluated.
+#' @export
+call_standardise <- function(call = NULL, env = NULL, fn = NULL,
+                             enum_dots = FALSE) {
+  call <- call %||% call_frame(2)
```
hadley: Wouldn't you almost always want to access the call with `call_standardise()` though?

lionel-: But `call_fn()` would not return the call but the function used in a call or formula.

hadley: Oh got it, sorry

R/call.R
```diff
@@ -25,11 +25,178 @@ call_new <- function(f, ..., .args = list()) {
...
+    })
+
+    # Replace dots symbol with actual dots and remaining args
+    set_cdr(d, dots)
+  }
+
+  call
+}
+
+call_validate_args <- function(call, env, fn) {
```
hadley: Ok, I get it now

R/call.R
hadley: You should also think about what to do for calls like `f()()()` or `(f())()`

lionel-: yup this returns `NULL` as well.

lionel-:
@hadley I addressed your comments and added in the process:

- `fn_args_defaults()` (equivalent of `formals()`) and `fn_args_names()`
- `call_args()` and `call_args_names()`
- `prim_name()` to get the name of a primitive function

maybe `fn_args_defaults()` should be `fn_args()` for symmetry though. I chose a different name because formals and actuals have slight syntactic differences:

```r
 # here a is an arg name
function(a) NULL

# here a is an arg
fn(a)
```

lionel-:
Also `call_modify()` now accepts dots, as with `call_new()`.

hadley:
Maybe use `fn_formals` instead of `fn_args_defaults`? That ties to the existing convention of formal arguments vs. actual arguments.

R/call.R
```diff
@@ -194,61 +183,117 @@ call_match <- function(call, fn, enum_dots) {
...
+    eval(info$call[[1]], info$env)
   }
 }
 
-#' Function name of a call
+#' Extract function name of a call
 #'
 #' @inheritParams call_standardise
-#' @return A string with the function name.
+#' @return A string with the function name, or \code{NULL} if the
```
hadley: I wonder if returning `NA_character()` would be better - it slightly simplifies the interface, but still conveys the same idea.

lionel-: hmm I'm not sure. I think `NA`s make sense as an option type mostly in a vectorised context, and taking the name of a function feels like a scalar operation. E.g. I don't like testing with `is.na()` in a if branch because it is a vector function. Also `NA` has length 1 so you cannot use `length()` as a test like you can with `NULL`. Of course someone might want to use the function with `map_chr()`, in which case `NA` is better. It depends on the actual use-cases I guess.

lionel-:
I used `fmls` so that it aligns with `args` ;)

Is it ok to depend on `stringi`? I use it to match arguments that might contain special characters: https://github.com/hadley/rlang/pull/1/commits/7f006418cc010c62b54d2b5cbc0bc27c92028a17. Ideally we'd just compare character by character and stop early like a regex starting with `^` would, but I'm not sure how to handle encodings to do it manually.

hadley:
Unfortunately we can't rely on stringi because it's too heavy as a dependency. You'll have to use base R functions instead - I think you can take advantage of the fact that symbols are always stored in the native encoding, so just call `enc2native()` on the comparison string.

PR #2, opened 2016-11-08
arg introspection

lionel-:
Includes #1


lionel-:
Ready for review @hadley

lionel-:
cc @kalibera

src/expr.c
```diff
@@ -65,12 +65,12 @@ SEXP base_promise_env(SEXP promise, SEXP env) {
   return env;
 }
 
-SEXP expr_find_(SEXP name, SEXP env) {
+SEXP arg_find_(SEXP name, SEXP env) {
```
lionel-: oops yes

lionel-:
@hadley `arg_info()` is now exported and documented.

R/arg.R
```diff
@@ -1,85 +1,59 @@
...
-#' @param x A promise (function argument)
-#' @export
-#' @examples
-#' # Unlike substitute(), arg_expr() finds the original expression
-#' f <- function(x) g(x)
-#' g <- function(y) h(y)
-#' h <- function(z) list(substitute(z), arg_expr(z))
+#' @param x An argument of a function.
+#' @return A list containing:
+#' \itemize{
```
hadley: From R-exts

> Description of the function’s return value.
> 
> If a list with multiple values is returned, you can use entries of the form
>
>\item{comp_i}{Description of comp_i.}
>for each component of the list returned. Optional text may precede93 this list (see for example the help for rle). >Note that \value is implicitly a \describe environment, so that environment should not be used for listing >components, just individual \item{}{} entries.

So you can drop the `\itemize`, and use the `\item{expr}{...}` form

hadley:
Does this also fix https://github.com/hadley/lazyeval/issues/91?

R/arg.R
```diff
@@ -10,32 +10,26 @@
...
+#'
+#'   \item{eval_frame}{The frame providing the scope for \code{expr},
+#'     which should normally be evaluated in \code{eval_frame$env}.
+#'     This is either the original calling frame, or the execution
+#'     frame of the function that was called if the argument was
+#'     missing. The difference reflects the evaluation rules of R,
+#'     where default arguments are scoped within the called function
+#'     rather than the calling frame.}
+#'
+#'   \item{caller_frame}{The original calling frame. This is the same
```
hadley: Maybe document `caller_frame()` first? i.e. something like "The original calling frame. This is the environment in which `expr` will be evaluated, unless the argument is missing (see below)" ?

R/arg.R
```diff
@@ -214,12 +210,79 @@ arg_text_ <- function(x, width = 60L, nlines = Inf) {
...
+#'
+#' # Note that accessing a missing argument contained in a list does
+#' # not trigger an error:
+#' l[[1]]
+#' is.null(l[[1]])
+#'
+#' # But if the missing argument is assigned in the current
+#' # environment, it is no longer possible to touch it. The following
+#' # lines would all return errors:
+#' #> x
```
hadley: Better to put in a `\dontrun{}`?

R/stack.R
```diff
@@ -92,10 +104,26 @@ is_frame <- function(x) {
...
+}
+
+#' @rdname stack
+#' @export
 eval_frame <- function(n = 1) {
   stopifnot(n > 0)
   pos <- sys.nframe() - n
-  if (pos < 1) {
+
+  if (pos == 0) {
```
hadley: I think this would be slightler clear as:

```
if (pos < 0L) { 

} else if (pos == 0L) {

} else {

}
```

R/arg.R
lionel-: I've had a CRAN maintainer run the dontrun examples in vdiffr and complain they were throwing errors. Maybe that was only because most of the examples were dontrun{}'d?

hadley: Are you sure you weren't thinking of `\dontest{}`?

hadley: Oh wow!

PR #4, opened 2016-11-18
Fix Recall()

lionel-:
Includes #3

This patch adds post-processing in `call_stack()` to restore continuity in the stack when `Recall()` is involved. Call arguments and caller positions need to be adjusted.

This is one step in making `arg_info()` compatible with the next version of magrittr in which `freduce()` applies the function sequence recursively with `Recall()`. To make magrittr fully compatible requires two additional upstream changes:

*  Break the chain of passed arguments by renaming `value` to another name before passing it to elements of the function sequence.

*  Pass to these functions an argument called `.` so that

   ```r
   mtcars %>% lazyeval::lazy_dots()
   ```

   returns `lazy$expr` = `.`. For instance `dplyr::do()` relies on this.

E.g.

```{r}
freduce <- function(value, function_list) {
  . <- value
  if (length(function_list) == 1L)
    function_list[[1L]](.)
  else
    Recall(function_list[[1L]](.), function_list[-1L])
}
```

We could make these changes to magrittr, release a new version, and have lazyeval or rlang depend on that magrittr version to make sure we are compatible. cc @smbache

I noticed one more incompatibility in that `mtcars %>% lazy()` returns `.` as lazy expr with the new rlang backend, while it inlines mtcars with the old promise-based backend. But I don't think we can do anything on that front. I'm rerunning the revdep checks, hopefully that won't be an issue.


lionel-:
@smbache The reason we are changing lazyeval's backend from climbing promises to climbing the call stack is that the former approach doesn't play well with the bytecode compiler.

smbache:
The next version of magrittr will not use freduce at all. Would you consider having a look at the 'update' branch to see how this works for you?

lionel-:
Merging this for now as I need some of the fixes to call stack code in other branches.

@smbache Will do. We are still thinking about exactly what kind of semantics we want for lazy evaluation, and still discussing with Luke Tierney what guarantees we can expect from R core.

PR #6, opened 2016-11-28
Move type predicates from purrr to rlang

lionel-:
I also added a discussion of closures vs primitives in `?is_function`. Maybe that should be a vignette?


hadley:
LGTM.

I'd say merge, and then move detailed discussion to a vignette.  

If you rebase the env pr, I'll comment on that next

lionel-:
env PR is now rebased.

PR #7, opened 2016-11-28
Environment functions

lionel-:
Includes #6 

On hindsight, maybe whether to look in parent environments should be an argument like in base R. Currently this is being handled in different verbs. E.g. `env_sees()` rather than `env_has(fetch = TRUE)`.


lionel-:
cc @hadley

R/formula.R
```diff
@@ -164,3 +166,73 @@ f_list <- function(...) {
 as_f_list <- function(x) {
   .Call(lhs_name, x)
 }
+
+#' Coerce quoted expressions to quoted formula.
```
hadley: I'm not sure about the name of this function, but everything else looks ok, so I'd say merge.

This may be another candidate to live in the lazyeval package, since it's in support of a specific way of handling NSE/quoted args

PR #8, opened 2016-12-03
Condition handling

lionel-:
Includes #6 and #7.

Also some functions for non-local returns since they are related to restarts.


R/stack.R
```diff
@@ -391,3 +402,187 @@ sys_frame <- function(n) {
...
+
+  i <- 1
+  for (env in envs) {
+    if (identical(env, frame)) {
+      return(length(envs) - i)
+    }
+    i <- i + 1
+  }
+
+  stop("this environment is not on the stack", call. = FALSE)
```
hadley: `` "`frame` not found on call stack" ``

R/stack.R
```diff
@@ -391,3 +402,187 @@ sys_frame <- function(n) {
...
+}
+
+#' Compute the distance between a frame and the current context.
```
hadley: Should this be documented with `frame_position()`? 

I also wonder if you could make the names more symmetric by using `top` and `bottom` suffixes?

R/stack.R
```diff
@@ -391,3 +402,187 @@ sys_frame <- function(n) {
...
+#' # site:
+#' identity(identity(
+#'   stack_trim(identity(identity(eval_stack())))
+#' ))
+#'
+#' # It is especially useful when used within a function that needs to
+#' # inspect the evaluation stack but should nonetheless be callable
+#' # within nested calls without side effects:
+#' stack_util <- function() {
+#'   # n = 2 means that two layers of intervening calls should be removed
```
hadley: How do you know that this should be 2 for this case?

R/stack.R
```diff
@@ -391,3 +402,187 @@ sys_frame <- function(n) {
...
+#' These functions should only be used rarely. These sort of non-local
+#' gotos can be hard to reason about in casual code, though they can
+#' sometimes be useful. Also, consider to use the condition system to
+#' perform non-local jumps.
+#'
+#' @param frame An environment, a frame object, or any object with an
+#'   \code{\link{env}()} method. The environment should be an
+#'   evaluation environment currently on the stack.
+#' @param value The return value.
+#' @export
```
hadley: This really feels like it needs an example

R/env.R
```diff
@@ -0,0 +1,702 @@
+#' Get an environment.
+#'
+#' Environments are objects that create a scope for evaluation of R
+#' code. Reification of scope is one of the most powerful feature of
```
hadley: Is reification the right word here?

R/env.R
```diff
@@ -0,0 +1,702 @@
...
+#' the R language: it allows you to change what objects a function or
+#' expression sees when it is evaluated. In R, scope is hierarchical:
+#' each environment is defined with a parent environment. An
+#' environment and its grand-parents form together a linear
```
hadley: grandparents

R/env.R
```diff
@@ -0,0 +1,702 @@
...
+#'
+#' \code{env()} is a s3 generic. Methods are provided for functions,
+#' formulas and frames. If called with a missing argument, the
+#' environment of the current evaluation frame (see
+#' \code{\link{eval_stack}()}) is returned. If you call \code{env()}
+#' with an environment, it acts as the identity function and the
+#' environment is simply returned (this helps simplifying code when
+#' writing generic functions).
+#'
+#' \code{env_new()} creates a new environment. \code{env_next()}
```
hadley: Why not `env_parent()`?

R/env.R
```diff
@@ -0,0 +1,702 @@
...
+#'
+#' # Create a new environment with a particular scope by setting a
+#' # parent. When inheriting from the empty environment, the
+#' # environment will have no object in scope at all:
+#' env <- env_new(env_empty())
+#' env_sees(env, "lapply")
+#'
+#' # The base package environment is often a good default choice for a
+#' # parent environment because it contains all standard base
+#' # functions:
```
hadley: But note that this will never inherit from packages the user has loaded (and maybe add a note about why it's impossible to get that?)

R/env.R
```diff
@@ -0,0 +1,702 @@
...
+#' @rdname env
+#' @export
+env.environment <- function(env = env_caller()) {
+  env
+}
+#' @rdname env
+#' @export
+env.default <- function(env = env_caller()) {
+  # Default argument env_caller() gets dispatched here:
+  if (is_env(env)) {
```
hadley: Why do we need this switch?

R/env.R
```diff
@@ -0,0 +1,702 @@
...
+#' @param x An object with an \code{env_set} method.
+#' @param value The new environment.
+#' @export
+`env<-` <- function(x, value) {
+  env_set(x, value)
+}
+
+#' @rdname env
+#' @export
+env_new <- function(parent = env_caller(), dict = list()) {
```
hadley: I wonder if it would be better to default to `parent = empty_env()`?

R/env.R
```diff
@@ -0,0 +1,702 @@
...
+}
+
+#' @rdname env
+#' @export
+env_next <- function(env = env_caller(), n = 1) {
+  env_ <- rlang::env(env)
+
+  while (n > 0) {
+    if (identical(env_, env_empty())) {
+      stop("not enough environments in scope", call. = FALSE)
```
hadley: My convention is to start error messages with a capital letter

R/env.R
```diff
@@ -0,0 +1,702 @@
...
+    eval.env = f_env(f),
+    assign.env = rlang::env(env)
+  )
+  do.call("delayedAssign", args)
+}
+
+#' Bury bindings and define objects in new scope.
+#'
+#' \code{env_bury()} is like \code{env_bind()} but it creates the
+#' bindings in a new child environment. Note that this function does
```
hadley: Instead of the last sentence, maybe: "This function does not modify its inputs"

R/env.R
```diff
@@ -0,0 +1,702 @@
...
+  env_bind(env_, dict)
+  env_set(env, env_)
+}
+
+#' Remove bindings from an environment.
+#'
+#' \code{env_unbind()} is the complement of
+#' \code{\link{env_bind}()}. Like \code{env_has()}, it ignores the
+#' parent environments of \code{env}. On the other hand,
+#' \code{env_eliminate()} is akin to \code{env_sees()} and will track
```
hadley: The pairs of names don't seem obviously symmetric to me - now I think an `inherits` argument would be better, but we should default it to `FALSE`.

R/env.R
```diff
@@ -0,0 +1,702 @@
...
+#' @examples
+#' parent <- env_new(env_empty(), list(foo = "foo"))
+#' env <- env_new(parent, list(bar = "bar"))
+#'
+#' # This throws an error because `foo` is not directly defined in env:
+#' # env_grab(env, "foo")
+#'
+#' # However `foo` can be fetched in the parent environment:
+#' env_fetch(env, "foo")
+env_grab <- function(env = env_caller(), nm) {
```
hadley: Why not make this `env_get()`?

R/env.R
```diff
@@ -0,0 +1,702 @@
...
+#' f <- with_env(env, ~new_formula())
+#' identical(f_env(f), env)
+#'
+#' # Or functions with a given enclosure:
+#' fn <- with_env(env, function() NULL)
+#' identical(env(fn), env)
+#'
+#'
+#' # Unlike eval() it doesn't create duplicates on the evaluation
+#' # stack. You can thus use it e.g. to create non-local returns:
```
hadley: Cool!!

R/language.R
```diff
@@ -20,8 +20,13 @@
 #' q3 <- quote(x + 1)
 #' is_lang(q3)
 #' is_call(q3)
+#'
+#' # A literal is either a scalar atomic vector or a name:
```
lionel-: From parsing. I'm not sure it is quite right to call literal something that has been parsed and stored in a variable however, since they have become constructed objects at that point. Maybe this should just be removed.

lionel-: Or should it have another name?

The concept that I was trying to represent is that these objects are self-quoting. E.g. if you quote a vector, you get a call to `c()`. But if you quote a scalar atomic vector, you get the scalar literally.

R/env.R
lionel-: I think so. The idea is that environments are scopes reified as first-class objects.

R/env.R
lionel-: I think I was in a "environments are linked lists" mood ;). Also it's shorter. But it's true `parent` is the canonical term.

lionel-: Also I find parent slightly more confusing because it suggests there can be several parents, whereas linked lists and environments are a purely linear sequence. Otoh the `inherit` terminology is a useful analogy and only makes sense with parents.

R/env.R
lionel-: Missing arguments get dispatched to `.default`, so we check for the default environment.

R/env.R
lionel-: I thought about it, but it seemed that it is rarer to inherit from the empty env. I thought the most common inheriting environments are: the current env, the base package, and the empty env, in that order. That's only based on my experience though. Should this be switched to the empty env?

R/env.R
lionel-: Because it would be misleadingly symmetric to `env_set()`. Should it be `env_get()` anyway?

At first `env_set()` was `set_env()` for that reason (setting the env of an object rather than setting something in the env) but it seemed wrong so I changed it.

R/stack.R
lionel-: One problem with top and bottom is that it's always ambiguous what they mean, because a call at top level is at the bottom of the call stack.

R/language.R
hadley: Maybe `is_leaf()` indicating that it's a leaf of the call tree? 

Do you also need to include `NULL` here?

R/stack.R
hadley: I'd just prefer something a bit more symmetric. Begin vs. end?

R/language.R
lionel-: good idea. Yes `NULL` should be there.

R/env.R
hadley: Ah got it

R/env.R
hadley: I'd prefer it was to make it more explicit when you want to inherit from the current

lionel-: ok. And `env_new(env())` is quite fast to type.

lionel-:
@hadley now rebased on master

R/conditions.R
```diff
@@ -0,0 +1,815 @@
+
+# conditions ---------------------------------------------------------
```
hadley: Still needed?

R/conditions.R
```diff
@@ -0,0 +1,815 @@
...
+
+#' Create a condition object.
+#'
+#' These constructors make it easy to create subclassed conditions.
+#' Conditions are objects powering the error system in R. They can
```
hadley: objects that

R/conditions.R
```diff
@@ -0,0 +1,815 @@
...
+#' # condition.
+cnd_signal <- function(.cnd, ..., .msg = NULL, .call = FALSE,
+                       .mufflable = TRUE) {
+  call <- if (.call) sys.call(-1) else NULL
+  cnd <- make_cnd(.cnd, ..., .msg = .msg, .call = call)
+  cnd_signal_(cnd, base::signalCondition, .mufflable)
+}
+#' @rdname cnd_signal
+#' @export
+cnd_panic <- function(.cnd, ..., .msg = NULL, .call = FALSE,
```
hadley: Why not just switch between `signal` and `panic` based on whether or not the condition class contains "error"?

How do you imagine people will usually use these? By creating a condition first or by using one of these functions?

R/conditions.R
```diff
@@ -0,0 +1,815 @@
...
+                      .mufflable = FALSE) {
+  call <- if (.call) sys.call(-1) else NULL
+  cnd <- make_cnd(.cnd, ..., .msg = .msg, .call = call)
+  cnd_signal_(cnd, base::stop, .mufflable)
+}
+
+make_cnd <- function(.cnd, ..., .msg, .call) {
+  if (is_scalar_character(.cnd)) {
```
hadley: Or `inherits(.cnd, "condition")` ? I feel like I'm missing something here

R/conditions.R
```diff
@@ -0,0 +1,815 @@
...
+}
+
+#' Signal an error, warning, or message.
+#'
+#' These functions are equivalent to base functions
+#' \code{\link[base]{stop}()}, \code{\link[base]{warning}()} and
+#' \code{\link[base]{message}()}, but the \code{type} argument makes
+#' it easy to create subclassed conditions. They also don't include
+#' call information by default. This saves you from typing
+#' \code{call. = FALSE} to make error messages cleaner within package
```
hadley: I do wonder if the right thing to do is keep the call information but not print it. But I guess we don't have control over that.

R/conditions.R
```diff
@@ -0,0 +1,815 @@
...
+#' normally after that. See \code{\link{rst_muffle}()} to jump to a
+#' muffling restart, and the \code{muffle} argument of
+#' \code{\link{inplace}()} for creating a muffling handler.
+#'
+#' @param msg A message to display.
+#' @param type Subclass of the condition to signal.
+#' @param call Whether to display the call.
+#'
+#' @export
+panic <- function(msg, type = NULL, call = FALSE) {
```
hadley: Panic seems a bit extreme ;)  How about throw?

R/conditions.R
```diff
@@ -0,0 +1,815 @@
...
+inform <- function(msg, type = NULL, call = FALSE) {
+  call <- if (call) sys.call(-1) else NULL
+  cnd <- cnd_message(type, .msg = msg, .call = call)
+  message(cnd)
+}
+
+
+# restarts -----------------------------------------------------------
+
+#' Restarts utilities.
```
hadley: Maybe best in its own file?

R/conditions.R
```diff
@@ -0,0 +1,815 @@
...
+#' @seealso \code{\link{rst_jump}()}, \code{\link{panic}()} and
+#'   \code{\link{cnd_panic}()}.
+#' @export
+#' @examples
+#' # The `abort` restart is a bit special in that it is always
+#' # registered in a R session. You will always find it on the restart
+#' # stack because it is established at top level:
+#' rst_list()
+#'
+#' # You can use the `above` restart to jump to top level without
```
hadley: above -> abort

R/conditions.R
```diff
@@ -0,0 +1,815 @@
...
+#'   on the stack.
+#' @param ...,.restarts Named restart functions. The name is taken as
+#'   the restart name and the function is executed after the jump.
+#' @param .env The environment in which to evaluate a captured
+#'   \code{expr}.
+#' @seealso \code{\link{return_from}()} and \code{\link{return_to}()}
+#'   for a more flexible way of performing a non-local jump to an
+#'   arbitrary call frame.
+#' @export
+#' @examples
```
hadley: Might be better to make this a little shorter and shift the rest to a vignette?

R/conditions.R
```diff
@@ -0,0 +1,815 @@
...
+#' There are two types of condition handlers: exiting handlers, which
+#' are thrown to the place where they have been established (e.g.,
+#' \code{\link{with_handlers}()}'s evaluation frame), and local
+#' handlers, which are executed in place (e.g., where the condition
+#' has been signalled). \code{thrown()} and \code{inplace()} create
+#' handlers suitable for \code{\link{with_handlers}()}.
+#'
+#' A subtle point in the R language is that conditions are not thrown,
+#' handlers are. \code{\link[base]{tryCatch}()} and
+#' \code{\link{with_handlers}()} actually catch handlers rather than
```
hadley: Oh maybe this is why you use `panic()` above

R/conditions.R
```diff
@@ -0,0 +1,815 @@
...
+#'   muffling restart. Otherwise, an error will be issued.
+#' @seealso \code{\link{with_handlers}()} for examples,
+#'   \code{\link{restarting}()} for another kind of inplace handler.
+#' @export
+thrown <- function(handler) {
+  structure(handler, class = c("thrown", "handler"))
+}
+#' @rdname thrown
+#' @export
+inplace <- function(handler, muffle = FALSE) {
```
hadley: Since the handler functions only ever have one argument, should we provide a formula/expression helper?

```R
inplace(~ print(.))
thrown(~ 10)
```

R/conditions.R
```diff
@@ -0,0 +1,815 @@
...
+#' handlers suitable for \code{\link{with_handlers}()}.
+#'
+#' A subtle point in the R language is that conditions are not thrown,
+#' handlers are. \code{\link[base]{tryCatch}()} and
+#' \code{\link{with_handlers}()} actually catch handlers rather than
+#' conditions. When a critical condition signalled with
+#' \code{\link[base]{stop}()} or \code{\link{panic}()}, R inspects the
+#' handler stack and looks for a handler that can deal with the
+#' condition. If it finds a throwable handler, it throws it to the
+#' function that established it (\code{\link{with_handlers}()}). If R
```
hadley: What does it mean to throw a handler? I don't quite follow what's happening here.  This might need more explanation in a vignette

R/conditions.R
```diff
@@ -0,0 +1,815 @@
...
+#' rst_bar <- function(df, nms) {
+#'   stats::setNames(df, nms)
+#' }
+#'
+#' # This restart is simpler and does not take arguments
+#' rst_baz <- function() "baz"
+#'
+#' # Signalling a condition parameterised with a data frame
+#' fn <- function() {
+#'   with_restarts(cnd_signal("foo", foo_field = mtcars),
```
hadley: Should we provide defer/on.exit style too?

R/conditions.R
lionel-: I was thinking creating a condition manually would be somewhat exceptional. So you'd use `cnd_signal()` and `cnd_panic()` to create a condition on the fly:

```r
# Create a condition of type `foo` and signal it:
cnd_signal("foo")
```

We could still switch within a single `cnd_signal()` function by accepting a chr vector:

```r
cnd_signal(c("my_type", "error"))
```

There is the issue that by default signalling a serious error should be done without a muffle restart. We could take `.mufflable = NULL`.

R/conditions.R
lionel-: This is for creating a typed condition on the fly with the signal function.

lionel-: e.g.

```r
cnd_signal("loop_tick")
```

R/conditions.R
lionel-: We could store the call in a `.call` field (instead of `call` which gets printed) when `FALSE`?

R/conditions.R
lionel-: That's how I called it at first until I understood that conditions are signalled in place rather than thrown in R. If the condition isn't handled (doesn't jump out of the signal function), then R jumps to the abort restart.

oh, how about `abort()`?

R/conditions.R
lionel-: I'm not sure anymore that throw is the right term. I was trying to convey that when R finds an exiting handler, it jumps to the stack frame where the handler was defined and calls the handler there. The stack is unwound before calling thrown handlers, similarly to exception systems in other languages.

Other ideas:

- `exiting(function(c) NULL)`? Or `jumping()`. But exiting/jumping where? The throw analogy has the advantage that it implies a target.

- `unwinding(function(c) NULL)`? Probably too technical.

- `rewinding(function(c) NULL)` might be more suggestive than unwinding (to people that had tapes when they were young anyway).


Maybe throw is best, as the distinction between thrown conditions and thrown handlers is a good way of describing how R is different from other languages. It may need more explanation that it really means a jump.

R/conditions.R
lionel-: good idea!

R/conditions.R
hadley: But what about `cnd_signal(cnd_new("foo"))`?

R/conditions.R
hadley: That's not a bad idea.

R/conditions.R
hadley: Yeah, `abort()` would be better

R/conditions.R
hadley: `aborting()` ? `terminating()`? `escaping()` ?

The distinction between the two types is about where executing continues after the handler, right? For `local()` handlers, execution proceeds from where it just left; for `thrown()` handlers, execution proceeds from where the handler was declared.

R/conditions.R
hadley: See the local variants in https://github.com/jimhester/withr

i.e. you could write:

```R
fn <- function() {
  local_restarts(
    rst_bar = rst_bar,
    rst_baz = rst_baz
  ) 
  cnd_signal("foo", foo_field = mtcars)
}
```

(I'm not sure that's the possible)

R/conditions.R
lionel-: That's right. Exiting handlers are thrown to `with_handlers()`. Do you prefer `exiting()` or `thrown()`?

PR #10, opened 2016-12-06
Switch to new capture approach

lionel-:
Includes #7 

* Rename `f_capture()` to `arg_capture()`. `arg_capture()` and `dots_capture()` now work only across one call level.
* `arg_expr()`, `arg_label()` work across one call level as well. Remove `arg_env()` since it would be synonymous with `env_caller()` / `parent.frame()`.
* Rename `arg_info()` and `dots_info()` to `arg_inspect()` and `dots_inspect()`. These functions use the call stack for introspection and their limitations are documented.


lionel-:
cc @hadley

R/arg.R
```diff
@@ -1,14 +1,195 @@
...
+#'   most immediate call site is captured. For this reason, named
+#'   arguments should be captured by an NSE function at the outermost
+#'   level, and then passed around to SE versions that handle
+#'   pre-captured arguments. See \code{\link{arg_inspect}()} for another
+#'   approach to introspecting arguments with which it is possible to
+#'   capture expressions at the outermost call site. This approach may
+#'   be harder to reason about and has some limitations.
+#'
+#'   Dots are different from named arguments in that they are
+#'   implicitely forwarded. Forwarding dots does not create a new call
```
hadley: 'implicitly'

lionel-:
Merging this for now so I can test with direct capture of named arguments. But we may need to switch to direct capture of dots as well, cf. discussion on Slack.

PR #11, opened 2016-12-08
Potpourri of new functions

lionel-:
I particularly like the new implementation of `invoke()`. It creates clean call stacks. Does it make sense to have `invoke()` in rlang, and keep the invoke-map versions in purrr? Maybe it'd be better to rename `.args` to `.x` for cross-package consistency, though `.x` feels more like purrr than rlang.

cc @hadley


R/attr.R
```diff
@@ -0,0 +1,174 @@
...
+  if (any(nms == "" | is.na(nms))) {
+    return(FALSE)
+  }
+
+  TRUE
+}
+
+#' @rdname is_named
+#' @export
+is_dict <- function(x) {
```
hadley: `dict` seems a bit too short a name for this.

R/attr.R
```diff
@@ -0,0 +1,174 @@
...
+has_names <- function(x) {
+  nms <- names(x)
+  if (is.null(nms)) {
+    rep(FALSE, length(x))
+  } else {
+    !(is.na(nms) | nms == "")
+  }
+}
+
+#' Set names of a vector.
```
hadley: Do you imagine this will replace the equivalent in purrr?

R/attr.R
```diff
@@ -0,0 +1,174 @@
...
+
+  names(x) <- nm
+  x
+}
+
+#' Get names of a vector.
+#'
+#' This names getter always returns a character vector, even when an
+#' object does not have a \code{names} attribute. In this case, it
+#' returns a vector of empty names \code{""}.
```
hadley: Should it also convert `NA` to `""`?

R/attr.R
```diff
@@ -0,0 +1,174 @@
...
+#'
+#' @param x A vector.
+#' @export
+#' @examples
+#' names2(letters)
+names2 <- function(x) {
+  names(x) %||% rep("", length(x))
+}
+
+#' Get length of an object.
```
hadley: This might be better in lobstr where it would be `prim_length()`

R/call.R
```diff
@@ -425,3 +425,17 @@ call_args_names <- function(call = NULL) {
   call <- call_info(call, NULL)$call
   names2(call_args_lsp(call))
 }
+
+#' Inspect a call.
+#'
+#' This function is useful for quick testing and debugging when you
+#' manipulate expressions and calls. It lets you check that a function
+#' is called with the right arguments. This can be useful in unit
+#' tests for instance.
```
hadley: Probably should note that this is just a wrapper around `match.call()` that makes the semantics clearer

R/env.R
```diff
@@ -193,6 +204,78 @@ env_tail <- function(env = env_caller()) {
 }
 
 
+#' Coerce to an environment.
+#'
+#' This is a s3 generic. The default method coerces named vectors
```
hadley: S3

R/env.R
```diff
@@ -193,6 +204,78 @@ env_tail <- function(env = env_caller()) {
...
+    warning("`parent` ignored for empty environment", call. = FALSE)
+  }
+  env_empty()
+}
+
+#' @rdname as_env
+#' @export
+as_env.environment <- function(x, parent = NULL) {
+  if (!is_null(parent)) {
+    env_set_parent(x, parent)
```
hadley: Should this copy `x`?

R/env.R
```diff
@@ -193,6 +204,78 @@ env_tail <- function(env = env_caller()) {
...
+  }
+  x
+}
+
+#' @rdname as_env
+#' @export
+as_env.character <- function(x, parent = NULL) {
+  if (!is_scalar_character(x)) {
```
hadley: I think this would be slightly clearer as 

```R
if (length(x) > 1) {
  return(as_env.default(...))
}
```

But it's not clear that length is the right check. What should `as_env(c(a = "b")` return?

R/language.R
```diff
@@ -21,36 +22,115 @@
...
 #' is_call(q3)
 #'
-#' # A literal is either a scalar atomic vector or a name:
-#' is_literal("string")
-#' is_literal(letters)
-#' is_literal(quote(name))
+#'
+#' # Atomic language objects are the terminating nodes of a call
+#' # tree. They are either a scalar atomic vector, a name, or NULL:
+#' is_atomic_lang("string")
```
hadley: I like this terminology :)

R/read-eval.R
```diff
@@ -0,0 +1,174 @@
...
+#'
+#' # The versions prefixed with f_ return formulas:
+#' f_read("foo %>% bar()")
+#' f_read_list("1; 2; mtcars")
+#'
+#' # The env argument is passed to as_env(). It can be e.g. a string
+#' # representing a scoped package environment:
+#' f_read("identity(letters)", env = env_empty())
+#' f_read_list("identity(letters); mtcars", env = "base")
+read <- function(x) {
```
hadley: `read` is a bit too short and too similar to readr.

R/attr.R
lionel-: yes, I think we have used purrr as both a functional programming library and a stdlib. Does it make sense to move all stdlib stuff from pryr, purrr, etc to rlang?

R/attr.R
lionel-: since `names2()` is usually not implemented that way, it should probably be an option. `regularise_na = FALSE`?

R/env.R
lionel-: hmm yes, it'd make sense to also check `x` has a valid name

R/read-eval.R
lionel-: This could be `read_expr()` and `read_exprs()`, and they both could accept connections and could be used like this: `read_exprs(file("src.R"))`?

lionel-:
> But maybe the consistent of the argument names isn't that important?

I'd think so. Is there a good deprecation strategy when exported functions are moved to another package? Or are we stuck reexporting the functions?

R/attr.R
hadley: I think we should try and keep rlang fairly focussed on low-level language issues. I'm not sure `set_name()` is in keeping with the rest of the functions. 

I think pryr should definitely go away as it's a hodgepodge of code that should live in rlang, lobstr, and purrr.

R/read-eval.R
hadley: What about `parse_expr()` and `parse_exprs()`?  

Accepting connections would be handy.

hadley:
I think it would be reasonable to re-export `invoke()` from purrr since it is useful in FP too.

R/read-eval.R
lionel-: ok. I liked `read()` because of the connection with REPL terminology, but I guess it's not really important.

R/attr.R
lionel-: I think setting names is as low level as checking the type of an object. The other place `set_names()` could go is vctrs, but my feeling is that it's an important function that should be easily accessible to all packages. It's like having `setNames()` in stats:: rather than base::, it always seemed to be out of place to me.

lionel-: how about `rep_along()`, rlang or vctrs? :)

lionel-:
@hadley 

all functions now accept connections, and will close them automatically so it's handy to type `parse_exprs(file("src.R"))`. I wonder if we should use conditions and restarts to prevent closing connections:

```r
retaining_connections({
  exprs <- parse_exprs(my_conn)
  exprs2 <- parse_exprs(my_conn2)
})
```

The internal functions would look like this:

```r
conn_close <- function(conn) {
  with_restarts(conn_retain = function() NULL, {
    cnd_signal("conn_close")
    close(conn)
  })
}

retaining_connections <- function(expr) {
  with_handlers(expr, conn_close = restarting("conn_retain"))
}
```

R/attr.R
hadley: Ok, that makes sense.  We'll just need to also re-export these functions in a package which we expect data analysis users to load.

R/language.R
```diff
@@ -21,36 +22,115 @@
...
+#' # However you can use the fact that as_name(quote(base::list()))
+#' # extracts the function identifier as is, and thus returns the call
+#' # base::list:
+#' is_call(~base::list(baz), ~base::list(), 1)
+is_call <- function(x, name = NULL, n = NULL) {
+  if (is_formula(x)) {
+    x <- f_rhs(x)
+  }
+
+  if (!typeof(x) == "language") {
```
hadley: `!=` ?

R/language.R
```diff
@@ -96,27 +181,31 @@ as_name.formula <- function(x) {
...
-  parse(text = x)[[1]]
+  parse_expr(x)
 }
-
 #' @export
 as_call.formula <- function(x) {
   as_call(f_rhs(x))
 }
+
+is_prefixed_name <- function(x) {
```
hadley: Maybe we should make `is_call_to(x, c("::", ":::", "$", "@"))` work?

R/read-eval.R
```diff
@@ -0,0 +1,176 @@
+#' Parse R code.
+#'
+#' These functions parse_expr and transform text into R expressions.
```
hadley: Something is wrong with this sentence

R/read-eval.R
```diff
@@ -0,0 +1,176 @@
...
+#' This is the first step of interpreting or evaluating a piece of
+#' code written by a programmer.
+#'
+#' \code{parse_expr()} returns one expression. If the text contains
+#' more than one expression (separated by colons or new lines), an
+#' error is issued. On the other hand \code{parse_exprs()} can handle
+#' multiple expressions. It always returns a list of expressions
+#' (compare to \code{\link[base]{parse}()} which returns an obsolete
```
hadley: Might be better to use a kinder word than obsolete

R/read-eval.R
```diff
@@ -0,0 +1,176 @@
...
+#' @export
+#' @examples
+#' # parse_expr() can parse_expr any R expression:
+#' parse_expr("mtcars %>% dplyr::mutate(cyl_prime = cyl / sd(cyl))")
+#'
+#' # A string can contain several expressions separated by ; or \n
+#' parse_exprs("NULL; list()\n foo(bar)")
+#'
+#' # The versions prefixed with f_ return formulas:
+#' f_parse_expr("foo %>% bar()")
```
hadley: Shouldn't it be `parse_expr_f()`?  (Since we usually put return types on the end)

R/read-eval.R
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  exprs[[1]]
+}
+#' @rdname parse_expr
+#' @export
+parse_exprs <- function(x) {
+  if (inherits(x, "connection")) {
+    exprs <- parse(file = x)
+    close(x)
```
hadley: I think the usually philosophy with connections is only to close it if you opened it (and only open it, if it's closed)

R/language.R
lionel-: by the way, I think I understand better the concept of literal now. IIRC, literal refers to anything that's self-evaluable. So `quote(c(1, 2))` is not a literal but the SEXP that results from evaluating this in the base environment is a literal. Another way to view this is that anything that you can evaluate in the empty environment is a literal.

The tricky part is that literal vectors are represented in R in the same way as their language representation.

Since literal vectors are definitely atomic language objects as well (and technically scalar and vectors have the same representation as a wrapped C array), I'm not sure anymore that this predicate represents the right concept. Should we unexport it for now?

hadley: Yeah, agreed.  

It's a literal in the sense that it's a leaf node in the parse tree.  Using code you can manipulate the tree to contain any R object, but you can't get there from parsing text.

R/read-eval.R
lionel-: You're right. It now occurs to me it would be more consistent if `f_new()` was renamed to `new_f()`, since it seems the constructors in the tidyverse use `new_` as prefix. What do you think?

lionel-: hmm... I see we have `env_new()`, `call_new()`, etc so I'm not sure where I got this idea. But we should decide a naming strategy nonetheless. Should all constructors be prefixed or suffixed with `new`?

R/read-eval.R
lionel-: I have added a `temporary()` function that add a `tmp = TRUE` attribute to its argument. Then `parse_exprs()` uses `maybe_close()` to close connections. It only closes connections with a temporary attribute.

It is used like this:

```r
parse_exprs(temporary(file(path)))
```

`temporary()` could be useful in other contexts. Is this a good strategy? Should `temporary()` have another name? Maybe `transient()`? `deathbed()` if we want to get dramatic :)

R/language.R
```diff
@@ -33,10 +49,16 @@
 #' is_atomic_lang(quote(call()))
 #'
 #' # With the exception of names, atomic language objects have the
-#' # property of being self-quoting:
+#' # property of being self-quoting literals:
```
hadley: Maybe we should make this the definition of an atomic language object? (i.e. `identical(x, quote(x))` is true whenever `x` is an atomic lang object)

R/read-eval.R
```diff
@@ -68,7 +75,7 @@ parse_expr <- function(x) {
 parse_exprs <- function(x) {
   if (inherits(x, "connection")) {
     exprs <- parse(file = x)
```
hadley: This logic doesn't seem right to me. I think you can ditch the `temporary` stuff, and just do:

```R
if (!isOpen(x)) {
  open(x)
  on.exit(close(x))
}
```

That's the convention that most base functions follow

lionel-: Got it. I guess I still don't understand connections well :(

R/language.R
lionel-: hmm yes it would make sense from the perspective of computing on the language: symbols need to be inspected in a way that literals don't.

hadley: Maybe we should make it `is_literal()`? 

Then a node of the parse tree is either a literal, a name, or a call. (Or a pairlist in the special case of function arguments)

lionel-: So the first intuition was the right one :)

Now updated, along with some explanation of the role of pairlists in language objects.

PR #12, opened 2016-12-14
Feature redispatch

lionel-:
Includes #11 

Relevant commit: https://github.com/hadley/rlang/commit/487bd6c536dd890f404a76822613580113c9a42a

I tested redispatch performance with purrr by splitting `as_function()` into two generics, one for mapping and one for indexing. The following `map_()` has a 3ms redispatch cost compared to the original `map()`:

```r
map_ <- function(.x, .f, ...) {
  .f <- rlang::with_redispatch(as_function = "as_pluck",
    as_function(.f, ...)
  )
  .Call(map_impl, environment(), ".x", ".f", "list")
}
```


hadley:
Did you have a motivating example for this code? 

I don't quite follow the map example.

lionel-:
The idea was to refactor `purrr::as_function()` into two generics: one general for creating functions, and one specific for creating plucking functions. The `as_function()` generic would live in rlang and `as_pluck()` would live in purrr. We'd use the condition system to redispatch `as_function()` to `as_pluck()` if no method was found.

hadley:
I still don't get it, but I'll take another look once #11 is merged.

lionel-:
@hadley now rebased.

PR #14, opened 2017-01-03
Moving purrr's general functions to rlang

lionel-:
- This moves the rest of purrr's general-purpose functions to rlang.
- The handler constructors now use `as_function()` to accept lambda-formulas.
- `modify()` as you suggested on Slack. I called it like this to mirror `call_modify()`. Maybe it should be `lst_modify()` or `list_modify()`.
- `with_attributes()` is like `structure()` but accepts a list in addition to dots.
- The `_along()` family handles all vector types and is completed with `_until()` family. Should the latter be `_len()` as in `rep_len()`?

Also the vector constructors accept attributes. Not sure this is a good idea though:

```r
list_along(letters, class = "s3_class")
```

versus:

```r
with_attributes(list_along(letters), class = "s3_class")
```


R/vector.R
```diff
@@ -0,0 +1,213 @@
...
+  }
+}
+
+#' Modify a vector.
+#'
+#' This function merges a list of arguments into a vector. It always
+#' returns a list.
+#'
+#' @param .x A vector to modify.
+#' @param ...,.elts List of elements to merge into \code{.x}.
```
hadley: Named elements replace existing named elements. Unnamed elements are inserted at the ned.

R/vector.R
```diff
@@ -0,0 +1,213 @@
...
+#' rep_along(x, 1, class = "my_class")
+#' double_along(x, class = "my_class")
+#' @name along
+#' @seealso new-vectors
+NULL
+
+#' @export
+#' @rdname along
+logical_along <- function(.x, ..., .attrs = list()) {
+  with_attributes(vector("logical", length(.x)), ..., .attrs = .attrs)
```
hadley: I don't love the default values that `vector` fills in. I'd rather have a generic `na_along(x)` which would take the type from `x`, optionally overridden by a "type" argument (or _maybe_ `na_dbl_along(x)` etc).

R/vector.R
```diff
@@ -0,0 +1,213 @@
...
+#' # Add attributes, including the S3 class:
+#' integer_until(0, index = 1)
+#' double_until(10, class = "my_class")
+#' @name new-vectors
+#' @seealso along
+NULL
+
+#' @export
+#' @rdname new-vectors
+logical_until <- function(.n = 0, ..., .attrs = list()) {
```
hadley: I don't get the `_until()` suffix. Why not `_new()`?

lionel-: `_new()` felt like it would take actual contents rather than a length. Maybe that's just me?

hadley: Hmmmm, I think vectors are probably a special case.

lionel-:
- I have added aliases for typed NA's like `na_lgl`, `na_dbl` etc.
- I renamed the along family to use the usual type prefixes: `lgl_along()`.
- And the until family to new: `lgl_new()`
- Both the along() and new() families now create vectors filled with typed NA's.

Also the last commit adds the `:=` operator we talked about before the holidays.

hadley:
I like it!

PR #15, opened 2017-01-03
Refactor interp()

lionel-:
- Remove old lazyeval's interp generic
- Rename `f_interp()` to `interp()` and add interpolation of closures.
- Capitalise all unquoting functions.
- Let `UQS()` handle language objects. This is useful for interpolating bodies of closures.
- Add `!!`, `!!!`, and `` `!!`() `` shortcuts for unquoting and splicing.


R/arg.R
```diff
@@ -319,8 +319,8 @@ fml_default <- function(expr, fn) {
 #' @export
 #' @examples
 #' # The missing argument can be useful to generate calls
-#' f_interp(~f(x = uq(arg_missing())))
-#' f_interp(~f(x = uq(NULL)))
+#' interp(~f(x = UQ(arg_missing())))
```
hadley: Did you want to use `!!`?

lionel-: so should we promote `!!` as the default way of unquoting? I'd like that.

hadley: Yeah, I think so

R/interp.R
```diff
@@ -1,8 +1,10 @@
 #' Interpolate a formula
 #'
-#' Interpolation replaces sub-expressions of the form \code{UQ(x)} with
-#' the evaluated value of \code{x}, and inlines sub-expressions of
-#' the form \code{UQS(x)}.
+#' Interpolation replaces sub-expressions of the form \code{UQ(x)}
+#' with the evaluated value of \code{x}, and inlines sub-expressions
+#' of the form \code{UQS(x)}. Syntactic shortcuts are provided for
+#' unquoting and unquote-splicing by prefixing with \code{!!} and
+#' \code{!!!}.
```
hadley: I'm still not sure about `!!!`.

R/interp.R
```diff
@@ -39,12 +42,26 @@
 #' interp(f)
 #'
 #'
+#' # You can also unquote and splice syntactically with bang operators:
+#' interp(~mean(!!! args))
+#'
+#' # However you need to be a bit careful with operator precedence.
+#' # All arithmetic and comparison operators bind more tightly than `!`:
+#' interp(x ~ 1 +  !! (1 + 2 + 3) + 10)
```
hadley: Oh that's a major downside 😞

lionel-: yes :/

R/interp.R
lionel-: I really like it but it's your call ;)

hadley: Ok, let's just go with it. I can't see an option that's better (unless we use `!!.` or `!!?`)

PR #16, opened 2017-01-09
New semantics for interp()

lionel-:
Formulas are now treated as reified promises.

`find_data` and `complain` have been renamed to `data_source`: the former becomes a generic constructor for data sources and the latter is a set of methods for such sources.


R/env.R
```diff
@@ -562,6 +562,21 @@ env_get <- function(env = env_caller(), nm, inherit = FALSE) {
   get(nm, envir = env, inherits = inherit)
 }
 
+#' Clone an environment.
+#'
+#' This creates a new environment containing exactly the same objects.
```
hadley: ", optionally with a new parent"

R/f-eval.R
```diff
@@ -72,79 +49,148 @@ f_eval_lhs <- function(f, data = NULL) {
...
-  arg_env <- new.env(parent = env)
-  arg_env$.env <- complain(env, "Object '%s' not found in environment")
-  arg_env$.data <- complain(data, "Variable '%s' not found in data")
+  # Install fguards and fpromises. Make sure the latter propagate `data`.
+  eval_env$`_F` <- function(f) f
+  eval_env$`_P` <- function(f) f_eval(f, data)
 
-  eval(expr, data, arg_env)
-}
+  # Make unquoting and guarding operators available even when not imported
```
hadley: :+1:

R/f-eval.R
```diff
@@ -72,79 +49,148 @@ f_eval_lhs <- function(f, data = NULL) {
...
+  if (!has_binding(src, name)) {
+    abort(sprintf(attr(x, "lookup_msg"), name))
   }
-  x[[name]]
+  src[[name]]
 }
-
 #' @export
-`[[.complain` <- function(x, i, ...) {
+`[[.data_source` <- function(x, i, ...) {
```
hadley: If we're going to do this a little more formally, I think we should also define `length` and `names` methods (and whatever else is needed so the object is consistent).  (At some point I want to write up the methods that you need to override if you override one method - that's going to be quite important if we want to empower people to create new vector classes)

R/f-eval.R
```diff
@@ -72,79 +49,148 @@ f_eval_lhs <- function(f, data = NULL) {
...
 #' @rdname f_eval
+#' @param x An object for which you want to find associated data.
+#' @param lookup_msg An error message when your data source is
+#'   accessed inappropriately (by position rather than name).
 #' @export
-find_data <- function(x) UseMethod("find_data")
-
-#' @export
-find_data.NULL <- function(x) list()
+data_source <- function(x, lookup_msg = NULL) {
```
hadley: I think this is probably big enough that it can now go in its own file.

R/interp.R
```diff
@@ -79,25 +123,22 @@ interp <- function(f, data = NULL) {
...
 }
-
 #' @export
 #' @rdname interp
 UQF <- function(x) {
-  if (!is_formula(x)) {
-    abort("`x` must be a formula")
-  }
-  x
+  bquote(`_F`(.(x)))
```
hadley: Seems a little weird to be using `bquote()` here - maybe just construct the call directly?

PR #19, opened 2017-01-19
Self-evaluating formulas and unquote on capture

lionel-:
- There is now two types of unquoting: raw unquoting and promised unquoting. `expr_quote()` and `interp()` with a function or an expression do raw unquoting, while `f_quote()` and `interp()` with a formula do promised unquoting. Also `arg_capture()` does promised unquoting on capture.

  One-sided formulas are considered to be promises when unquoted. `expr_quote()` will inline the RHS while `f_quote()` inserts the formula as is. The trick is that `~` is given a special definition in the `f_eval()` evaluation environment so that the formula can evaluate itself.

- Two-sided formulas don't self-evaluate and so don't need to be guarded to be treated literally. In case a one-sided formula should be treated literally rather than as a promise, it should be unquoted with `UQF()`. This operator changes the call name of the formula from `~` to `_F()`, and `_F()` is defined in the `f_eval()` env to restore the name to `~`.

- `interp()` handles formulas containing unquote operators. It doesn't visit inner formulas so there is actually no need for a "visited" attribute. This is why fpromises are no longer inlined when their environment is not informative, and are unquoted as is instead. Also the consistent unquoting should make it more predictable to new users of tidyeval, as they won't have to learn why the fpromise inlines in some cases but not others.

- `f_eval()` is now more general and can deal with arbitrary expressions. This makes it possible to evaluate a raw expression with pronouns. While probably not very useful, this reflects the fact that tidyeval can deal with raw expressions.

The documentation has been updated to reflect these changes. I think tidy evaluation is now rather complete. A second PR with some file reorganisation and small changes is upcoming (I splitted in two PRs for readability). Now that `f_eval()` is more general, it should probably be an unprefixed verb just like `interp()`. `evaluate()` seems to be a good fit. Or maybe `tidyeval()`?


R/quote.R
```diff
@@ -0,0 +1,258 @@
...
+#' var <- quote(xyz)
+#' extra_args <- list(trim = 0.9, na.rm = TRUE)
+#' f_quote(mean(UQ(var) , UQS(extra_args)))
+#' interp(~mean(!!var , !!!extra_args))
+#'
+#'
+#' # Unquoting is especially useful for transforming a captured
+#' # expression:
+#' expr <- quote(foo(bar))
+#' expr <- expr_quote(inner(!! expr, arg1))
```
hadley: Maybe `expr_quote()` should go in another documentation file?  It's going to be used in a slightly different context.

R/quote.R
```diff
@@ -0,0 +1,258 @@
...
+#' # operator. You need to use UQE() for subsetting with dollar:
+#' var <- ~cyl
+#' f_quote(mtcars$UQE(var))
+#'
+#' # `!!`() is also treated as a shortcut. It is meant for situations
+#' # where the bang operator would not parse, such as subsetting with
+#' # $. Since that's its main purpose, we've made it a shortcut for
+#' # UQE() rather than UQ():
+#' var <- ~cyl
+#' f_quote(mtcars$`!!`(var))
```
hadley: For completeness, maybe mention `` `$`(mtcars, !!var)`` ?

R/arg.R
```diff
@@ -87,6 +87,8 @@ arg_capture <- function(x) {
 
   arg_expr <- substitute_(x_expr, x_env)
   arg_env <- env_caller(2)
+
+  arg_expr <- .Call(interp_, arg_expr, arg_env, TRUE)
```
hadley: I think this needs a line in the docs (i.e. this function will automatically interpolate)

R/f-eval.R
```diff
@@ -46,9 +46,9 @@
 #' f_eval(~ mean(cyl), mtcars)
 #' # How can you change the variable that's being computed?
 #' # The easiest way is "unquote" with !!
-#' # See ?interp for more details
+#' # See ?f_quote for more details
 #' var <- ~ cyl
-#' f_eval(~ mean( !!var ), mtcars)
+#' f_eval(f_quote(mean( !!var )), mtcars)
```
hadley: This is the key change, right? We now interpolate on capture; either automatically with `arg_capture()` or if the function takes formulas, the author needs to explicitly call `interp()` before `f_eval()`.  

That gives three possible options for functions: 

* accept raw expressions and call `arg_capture()`
* accept "quoted" formulas (i.e. expect that the user calls `f_quote()`
* accept formulas and immediately call `interp()`

lionel-:
@hadley I'm thinking that we should unexport all unquoting operators to reflect the fact that they are syntactic elements rather than usual function calls. What do you think?

R/f-eval.R
lionel-: I think the second case you mention should only be for internal functions, not user-exposed functions.

hadley: Right

hadley:
@lionel- hmmmmm, the problem is that they look like function calls and need to be documented like function calls, so I don't think there's a huge benefit to not exporting.

lionel-:
@hadley oh right I hadn't thought about documentation.

I'm not sure `tidy_eval_lhs()` and `tidy_eval_rhs()` do the right thing currently. They evaluate the lhs or rhs, but return the result as a one-sided formula. Maybe they should evaluate the lhs/rhs "in place" instead, and return a two-sided formula if the input is two-sided?

hadley:
Yeah, that sounds right.

PR #22, opened 2017-02-03
Ui tidyeval

lionel-:
New UI for tidyeval and some fixes:

- `tidy_quote()`, `tidy_interp()`, `tidy_eval()`

- `tidy_eval_lhs()` and `tidy_eval_rhs()` evaluate in place.

- `tidy_quote_expr()` returns an expression. It does not perform "raw unquoting" anymore (automatically taking the RHS). Doing so was making the code complicated and it was harder to document, especially with `tidy_interp()` which would have require a new argument to that effect.

- `tidy_eval_env()` is exported for packages that implement their own tidy evaluation scheme. It is paired with a cleanup function that removes the `~` self-evaluating definitions in case the environment leaks somewhere. For example if the evaluated expression captures the eval env by creating a closure.

- Unquote operators are now inlined during interpolation rather than made available in the interpolation env. This way unquoting has "frame consistent" semantics when the interpolation environment is leaked by creating a formula or closure.

New `expr` functions:

- `arg_label()` etc are renamed to `expr_label()` since they take an expression.

- New `expr_eval()` function for lightweight and "stack consistent" evaluation. It's a simple wrapper around `Rf_eval()` and is a better way than creating and forcing a promise. In the places where it's important to keep the call stack clean, like `invoke()`, `Rf_eval()` can be called directly via `.Call(rlang_eval, expr, env)`.


lionel-:
Would it be ok to have that signature in the formula ctor:

```r
new_f <- function(rhs, env = parent.frame(), lhs = NULL)
```

rather than:

```r
new_f <- function(rhs, lhs = NULL, env = parent.frame())
```

It seems much more common to specify an environment than a LHS.

hadley:
And new functions look good too.

PR #23, opened 2017-02-03
UI environments

lionel-:
Includes #22 

This PR updates the UI to reflect the prefix = input type and suffix = output type principle.

- Namespace functions have been renamed from `env_namespace()` to `ns_env()`. Named environments like `env_global()` have been renamed to `global_env()`. New `pkg_` and `scoped_` prefixes for functions about package environments and the search path.

- Renamed `as_quoted_f()` to `as_fquote()`. `parse_expr_f()` to `parse_f()` since it returns a formula. And `has_names()` to `has_name()` to reflect that it's vectorised.

- Renamed `dict` argument of env functions to `data`, since that is more in line with other similar arguments like in `tidy_eval()`. The concept of dictionary is still mentioned in the docs and I wonder if the `data_source` class should be renamed to `dictionary`. I think we should formally define the concept of dictionary as part of our reflections on S3 and generic programming and it seems like data sources for tidy evaluation are an instance of that, just like environment is an instance of dictionary (which is why data sources were modelled after environments).


R/call.R
```diff
@@ -302,7 +302,7 @@ call_modify <- function(.call = NULL, ..., .args = list(), .env = NULL) {
   call <- .call %||% call_frame(2)
   call <- call_standardise(call, .env)
 
-  if (!all(has_names(args))) {
+  if (!all(has_name(args))) {
```
hadley: This could just be `!is_named(args)` right?

R/expr.R
```diff
@@ -30,7 +30,7 @@
 #' @export
 #' @examples
 #' # expr_eval() works just like base::eval():
-#' env <- env_new(dict = list(foo = "bar"))
```
hadley: I wonder if it's worth having `as_env()`? Then this could be `env <- as_env(list(foo = "bar"))`

hadley: Or should we have `env(foo = "bar")` (with `.parent = emptyenv()`)?

R/call.R
lionel-: correct

R/expr.R
lionel-: we already have `as_env()`, which is indeed synonymous with `new_env()` in that case.

PR #24, opened 2017-02-03
UI ctors

lionel-:
Includes #22 and #23.

This simply renames all constructors to use `new` as prefix rather than suffix.


lionel-:
@hadley Now rebased.

lionel-:
I guess this one is straightforward enough so let's move to the next one ;)

PR #25, opened 2017-02-03
Compatibility with R 3.1.0

lionel-:
Includes #22, #23 and #24 

3.1.0 is not yet enabled on Travis because of a bug in devtools (and it's hard to install the dev version on R 3.1 because it lacks `https` support in download functions).


lionel-:
@hadley Added files `compat-purrr.R` and `compat-oldrel.R`. These files should serve as references for other packages.

lionel-:
@hadley now rebased

R/compat-oldrel.R
```diff
@@ -0,0 +1,21 @@
...
+if (utils::packageVersion("base") < "3.2.0") {
+
+  dir_exists <- function(path) {
+    !identical(path, "") && file.exists(paste0(path, .Platform$file.sep))
+  }
+  dir.exists <- function(paths) {
+    vapply_lgl(paths, dir_exists)
+  }
+
+  names.environment <- function(x) {
```
hadley: Will this work without being exported?

tests/testthat/test-call.R
```diff
@@ -241,6 +241,10 @@ test_that("call_args() and call_args_names()", {
 # call_stack() consolidation -----------------------------------------
 
 test_that("Recall() does not mess up call history", {
+  if (utils::packageVersion("base") < "3.3.0") {
+    skip("test for Recall() depends on internal implementation")
```
hadley: Can you file an issue on testthat to add a skip helper for this case?

R/compat-oldrel.R
lionel-: does the fix in the last commit look reasonable?

R/compat-oldrel.R
```diff
@@ -0,0 +1,25 @@
...
+if (utils::packageVersion("base") < "3.2.0") {
+
+  dir_exists <- function(path) {
+    !identical(path, "") && file.exists(paste0(path, .Platform$file.sep))
+  }
+  dir.exists <- function(paths) {
+    vapply_lgl(paths, dir_exists)
+  }
+
+  names <- function(x) {
```
hadley: I didn't know that the previous code didn't work, I just wasn't sure and it might've been ok.  So as long as the tests still pass, I'm ok.

PR #26, opened 2017-02-03
Tidy capture of dots

lionel-:
Includes #22, #23, #24 and #25

The `:=` operator is now called the pattern operator, with predicate function `is_pattern()`.

`tidy_dots()`:

- now interpolates on capture.
- accepts splice-unquote in dots: `tidy_dots(foo, !!! captured_dots)`
- evaluates the LHS of patterns (formulas of type `a := b`) and treats it as dot name, unless `patterned = FALSE`. If another name is supplied with a pattern (e.g. `var = a := b`, throws a warning).

`tidy_patterns()`:

- Returns patterns as is, with both LHS and RHS interpolated.
- Accepts names: `tidy_patterns(pat = a := b)`.


lionel-:
- Renamed `as_fquote()` to `as_tidy_dots()` and add `is_tidy_quote()`. A formula is not a tidy quote if it does not have environment information. `as_tidy_quote()` now only changes env info for such degenerate formulas.

- `tidy_dots()` now require explicit unquoting in the LHS of patterns, so it's more consistent with `tidy_quote()`. E.g. `tidy_dots(!!var := expr)`. Otherwise `:=` is a synonym of `=`.

- Added `tidy_dots_alt()` for DSLs like ggvis. The only difference is that dots are returned in a list with components `dots` and `alts`.

hadley:
Glad you made the `is_tidy_quote()` changes; I was just thinking about that.

lionel-:
@hadley now rebased

R/f.R
```diff
@@ -6,8 +6,8 @@
 #' @export
 #' @examples
 #' new_f(quote(a))
-#' new_f(quote(a), quote(b))
-new_f <- function(rhs, lhs = NULL, env = parent.frame()) {
+#' new_f(quote(a), lhs = quote(b))
+new_f <- function(rhs, env = caller_env(), lhs = NULL) {
```
hadley: This new argument order feels a bit weird. Can you comment on why you made the change?

R/f.R
```diff
@@ -52,7 +52,8 @@ f_rhs <- function(f) {
 #' @rdname f_rhs
 `f_rhs<-` <- function(x, value) {
   stopifnot(is_formula(x))
-  new_f(value, f_lhs(x), f_env(x))
+  f <- new_f(value, f_env(x), f_lhs(x))
+  maybe_pattern(f, x)
```
hadley: Maybe this would be better called `copy_lhs()`? But I'm a bit confused as to what it's doing here.

R/operators.R
```diff
@@ -49,9 +49,9 @@
   attr(x, name, exact = TRUE)
 }
 
-#' Quoting assignment operator.
+#' Definition operator.
 #'
-#' The colon-equals operator is typically used in DSL packages like
+#' The pattern operator is typically used in DSL packages like
```
hadley: definition?

R/tidy-capture.R
```diff
@@ -100,9 +90,198 @@ tidy_capture <- function(x) {
...
+#' expressions of the type \code{var := expr}.
+#'
+#'\describe{
+#'  \item{\code{tidy_dots()}}{
+#'    When \code{:=} definitions are supplied to \code{tidy_dots()},
+#'    they are treated as a synonym of argument assignment
+#'    \code{=}. On the other hand, they allow unquoting operators on
+#'    the left-hand side, which makes it easy to assign names
+#'    programmatically.}
+#'  \item{\code{tidy_dots_alt()}}{
```
hadley: Do we need to export this? I don't know if I want to promote ggvis style

R/tidy-quote.R
```diff
@@ -235,3 +235,64 @@ tidy_quote_expr <- function(expr) {
   expr <- substitute(expr)
   .Call(rlang_interp, expr, parent.frame())
 }
+
+
+#' Coerce expressions to a tidy formula quote.
+#'
+#' A tidy quote is a formula or pattern (see \code{\link{op-definition}})
```
hadley: "definition"

R/f.R
lionel-: I think that the most common case while working with tidyeval is to create tidy quotes. So this argument order reflects their respective importance.

R/f.R
lionel-: this is copying the call name, e.g. either `~` or `:=`. I'll rename to `copy_call_name()`.

R/f.R
hadley: Hmmm, I'd rather keep it as is, and force people to name `env` if needed.

R/tidy-capture.R
lionel-: I agree, it would be better to use a unified NSE style in the tidyverse.

R/f.R
lionel-: ok. Maybe have `new_quote(rhs, env)` in addition to `new_f(rhs, lhs, env)`? And then the latter can have the less confusing argument order `new_f(lhs, rhs, env)`

hadley: `new_tidy_quote()`? Otherwise, sounds good.

lionel-: hmm `new_tidy_quote()` feels a bit long for this common operation.

lionel-: `new_f()` and `new_f2()`?

hadley: Seem a bit short. They're not used that commonly are they?

lionel-: I think they will be mostly used for testing NSE functions, e.g. you'll mostly find them in testthat sections of packages. `_f` is the suffix that correspond to the `f_` prefix so it seems consistent. I can also just revert the commit so we stop over-bikeshedding this :)

PR #27, opened 2017-02-08
Replace is_numeric() with is_integerish()

lionel-:
- Remove `is_numeric()` as it is a confusing predicate (tests for types instead of concept). Keep `is_bare_numeric()` for testing for base numeric types

- Add `is_integerish()` to check whether a double is close enough to an integer, according to R tolerance rules. Useful when a UI function expects an integer input


lionel-:
Maybe logicals should be integerish?

hadley:
LGTM. Needs scalar version?

lionel-:
oh right

PR #28, opened 2017-02-10
Changes for tibble port

lionel-:
Added `.named` argument to ensure dots are named. The default width in rlang is 60, while tibble wants 500, so it accepts integers and forward them to `expr_text()`.


lionel-:
we can change it in tibble but I think being able to do `tidy_dots(..., .named = 20)` makes sense.

PR #34, opened 2017-02-16
Add functions to deal with character, string, and encoding

lionel-:
`symbol()` and `symbols()` allow symbols to be used within the tidyeval framework without losing the encoding information of strings. They always convert their inputs to the native encoding. This will allow e.g.

```r
select(data, !!! symbols(colnames))
```
cc @krlmlr 

I've added a few functions to help create strings and character vectors with specific encodings. They should make tests easier to write and read. The string functions are mostly equal to the chr_ versions, but check for scalar input.

```r
cafe <- string(c(0x63, 0x61, 0x66, 0xE9), "latin1")
str_bytes(cafe)
str_encoding(cafe)
```

I've documented how locales and encodings are dealt with in R in two places: `str_set_encoding()` discusses why it's important to consistently mark the encoding of strings, and `set_utf8_locale()` discusses the side effects of the `LC_CTYPE` locale.


R/types-chr.R
```diff
@@ -0,0 +1,295 @@
...
+    invisible(Sys.setlocale("LC_CTYPE", locale = "en_US.UTF-8"))
+  }
+}
+#' @rdname set_utf8_locale
+#' @export
+set_latin1_locale <- function() {
+  if (.Platform$OS.type == "windows") {
+    locale <- "English_United States.1252"
+  } else {
+    locale <- "en_US.ISO8859-1"
```
krlmlr: This is `"en_US.iso88591"` on my Ubuntu system. Does it make sense to use a system call to e.g. `locale -a` to find the right locale?

R/types-sym.R
```diff
@@ -0,0 +1,27 @@
...
+  }
+  if (!is_string(x)) {
+    abort("Only strings can be converted to symbols")
+  }
+  .Call(rlang_symbol, x)
+}
+#' @rdname symbol
+#' @export
+symbols <- function(x) {
+  map(x, symbol)
```
krlmlr: Should we define an S3 class here?

R/types-chr.R
lionel-: Could you try to set the locale with the all caps ISO please, to see if it accepts it? This is just testing tool so I'm not sure it's worth it trying a more complicated logic.

krlmlr: My bad -- `"en_US.ISO8859-1"` is also accepted, just not reported with `locale -a`.

R/types-sym.R
lionel-: We are trying to steer clear of S3 in rlang. I think it should be ok to keep it as a simple list.

lionel-:
>  I wonder if we should merge the utf8 package into rlang

I'm not sure. When is it useful to know whether a string is pure ASCII? What is "deep reencoding" by the way?

krlmlr:
`Encoding()` is not very helpful to detect if a character vector has "proper" encoding markers. Even after `enc2utf8()`, all ASCII elements are still reported as `"unknown"`.

Deep reencoding means traversing the entire object, which may be a list of lists of ... of character, and also the attributes, especially names.

It's just that your PR obliterates half of the utf8 package, why not add the other half ;-)

@hadley: Which other parts of `utf8` should live here?

lionel-:
The deep application should probably be handled with a purrr functional:

```r
l %>% map_leaves_if(is_string, as_utf8_string)
```

I'm not opposed to the idea of `str_encoding()` returning `"ASCII"` for ascii only input, but I don't think rewriting `Encoding()` to do it is the way forward (as we'll need to track upstream changes). Also it's probably not very helpful to know that a string is ASCII, since those can just be handled as if in the native encoding.

For the reading/writing functions, I think that's a case where they wouldn't belong in rlang but in a domain package, maybe readr?

R/types.R
```diff
@@ -129,6 +132,9 @@ is_scalar_double <- function(x) {
 is_scalar_character <- function(x) {
   is_character(x) && length(x) == 1
 }
+#' @export
+#' @rdname scalar-type-predicates
+is_string <- is_scalar_character
```
hadley: Should we add an encoding param?

R/types-chr.R
```diff
@@ -0,0 +1,295 @@
...
+  if (is_integerish(x)) {
+    x <- rawToChar(as.raw(x))
+  } else if (is_raw(x)) {
+    x <- rawToChar(x)
+  } else if (!is_string(x)) {
+    abort("`x` must be a string or raw vector")
+  }
+
+  if (!is_null(encoding)) {
+    x <- chr_set_encoding(x, encoding)
```
hadley: Maybe make `chr_set_encoding` do nothing is `encoding` is null? Would save a bit of duplication here

R/types-chr.R
```diff
@@ -0,0 +1,295 @@
...
+#' # to the current locale:
+#' \dontrun{
+#' bad <- set_names(str_set_encoding(latin1, "unknown"))
+#' set_utf8_locale()
+#'
+#' str_encoding(names(c(bad)))
+#' str_bytes(names(c(bad)))
+#' }
+chr_set_encoding <- function(x, encoding = c("unknown", "UTF-8", "latin1", "bytes")) {
+  Encoding(x) <- match.arg(encoding)
```
hadley: I wonder if this should warn if the encoding is unchanged

R/types-chr.R
```diff
@@ -0,0 +1,295 @@
...
+#' language and region of the locale. They have permanent side effects
+#' and should not be used in package examples, unit tests, or in the
+#' course of a data analysis.
+#'
+#' @export
+set_utf8_locale <- function() {
+  if (.Platform$OS.type == "windows") {
+    warn("UTF-8 is not supported on Windows")
+  } else {
+    cat("Locale codeset is now UTF-8\n")
```
hadley: `message()` ?

R/types-chr.R
```diff
@@ -0,0 +1,295 @@
...
+#'
+#' Note that these helpers are only provided for testing interactively
+#' the effects of changing locale codeset. They let you quickly change
+#' the default text encoding to latin1, UTF-8, or non-UTF-8 MBCS. They
+#' are not widely tested and do not provide a way of setting the
+#' language and region of the locale. They have permanent side effects
+#' and should not be used in package examples, unit tests, or in the
+#' course of a data analysis.
+#'
+#' @export
```
hadley: `@return The previous locale (invisibly)`

R/types-chr.R
```diff
@@ -0,0 +1,295 @@
...
+#' @return A raw vector for \code{str_bytes()} or a list of raw
+#'   vectors for \code{chr_bytes()}.
+#' @export
+str_bytes <- function(x) {
+  stopifnot(is_string(x))
+  charToRaw(x)
+}
+#' @rdname str_bytes
+#' @export
+chr_bytes <- function(x) {
```
hadley: Maybe this should be `as_bytes.character` ?

R/types-chr.R
lionel-: hmm I think that might cause more pain than good, but I'm not sure. `as_utf8_string()` should for sure be idempotent, but in this case maybe it's reasonable to assume that we should only set the encoding when we know it has been lost or was never set (in corner cases like manually created strings)?

R/types-chr.R
lionel-: hmm yes. We are going to settle on switchpatching for base-type coercion right? Also we should make base-type predicates non-generic? Currently `is_name()` and `is_call()` are still generics.

R/types-chr.R
lionel-: when I do it on my computer, it's returning the new locale, not the previous. But in the R CMD check scripts in `tools::`, they do use it as if it returned the old locale. Maybe a bug introduced at some point without anyone noticing?

hadley: Either way you should work around the bug with an explicit `Sys.getlocale()`

R/types-raw.R
```diff
@@ -15,3 +15,19 @@ bytes <- function(x) {
...
+#' Coerce to a raw vector.
+#'
+#' This currently only works with strings, and returns its hexadecimal
+#' representation.
+#'
+#' @param x A string.
+#' @return A raw vector of bytes.
+#' @export
+as_bytes <- function(x) {
+  if (is_bytes(x)) {
```
hadley: Maybe use switchpatch here just to be clear?

lionel-: do you mean using a literal `switch(typeof(x), ...)`? I'd prefer if switchpatch was taken as a figure of speech ;)

here we also dispatch on the size, so using switch would make the code slightly harder to read I think

hadley: I think this would be clearer as:

```R
switch(typeof(x),
   raw = x,
   character = charToRaw(x),
   stop("Failed to coerce")
)
```

lionel-: This is now:

```r
as_bytes <- function(x) {
  switch(typeof(x),
    raw = return(x),
    character = if (is_string(x)) return(charToRaw(x))
  )
  abort("`x` must be a string or raw vector")
}
```

lionel-: or do you prefer this way?

```r
as_bytes <- function(x) {
  x <- switch(typeof(x),
    raw = x,
    character = if (is_string(x)) charToRaw(x)
  )

  if (is_null(x)) {
    abort("`x` must be a string or raw vector")
  } else {
    x
  }
}
```

hadley: I like the first one better

lionel-:
I'm going to merge this PR for now but we can still change `symbol()` later once we've figured out what to do.

PR #38, opened 2017-02-20
C API for signaling conditions

jimhester:
Because R's C API does not have a direct way to signal a condition, this
implementation simply constructs a condition object and calls the R
`signalCondition` function.

This file is located in `inst/include` so that other packages can use it
by adding `LinkingTo: rlang` to their DESCRIPTION. The function is
static so it will be copied to each compilation unit, avoiding multiple
definition errors. `inline` is not part of the c89 standard, so cannot
be used without `-std=c99` or compiler extensions.

This API assumes the condition object will contain only only one (additional) class and no additional values apart from the message, we can discuss if we should provide a more flexible API or if this is sufficient. I am also not sure if we need to expose the evaluation env, or if always evaluating the  `signalCondition()` call in the global environment is sufficient.

### DIFF
--- a/inst/include/condition.h
+++ b/inst/include/condition.h
@@ -1,0 +1,37 @@
+#ifndef CONDITION_H_
+#define CONDITION_H_
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include <Rinternals.h>
+
+static SEXP signal_condition(const char * msg, const char * class_, SEXP env = R_GlobalEnv) {
+  SEXP condition, c, signalConditionFun, out;
+
+  const char *nms[] = {"message", ""};
+  PROTECT(condition = Rf_mkNamed(VECSXP, nms));
+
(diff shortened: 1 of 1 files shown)


lionel-:
I wonder if we should move this to `/src/` and just copy the file in other packages for the time being. So we can think about a more comprehensive and consistent C API for R.

lionel-:
Hi Jim,

rlang will feature a comprehensive C API in the future, and we would't want to commit to any public API for now. But we can host it in `/src/` if you'd like, this way packages that need this can just copy it from there. The same way rlang is hosting unexported compatibility functions (e.g. `/R/compat-purrr.R`), so they are all written and tested in the same place.

jimhester:
That is fine, we can just close this PR then.

I am using this code in odbc, so it can just stay there for now.

lionel-:
ok, let's revisit this once we get to the C API. But thanks for the PR!

PR #39, opened 2017-02-21
Fix symbol encoding

lionel-:
Includes #34 

Relevant commit: https://github.com/hadley/rlang/commit/8736dc5d3164bba881c43855d81ae1fc7793006d

With this version of `symbol()`, UTF-8 tagged strings are properly store in new symbols. The drawback is that if a symbol has been created beforehand with `as.name()`, the untagged string is cached in the symbol table and there is nothing `symbol()` can do anymore. Hopefully this doesn't cause too much problems, since users are more likely to create strings than symbols, and we'd advise to always use `symbol()` or `as_name()`.


lionel-:
This triggers a non-API call CRAN note :/

krlmlr:
Would you mind rebasing?

lionel-:
Here you go. According to Jim, CRAN accepts this kind of warning on old versions of R as long as it's for compatibility.

krlmlr:
Can we revert https://github.com/hadley/rlang/pull/34/commits/5156d92b645c02755a4751df58cffb9148a0a2c6 and provide a lossless conversion from `symbol` to `character`? I can think of two options for converting back from symbol to UTF-8 string:

- Interpret strings of the format `"<U+xxxx>"` as Unicode characters: This is cheap, but we should then warn or err on calls of the form `symbol("<U+xxxx>")`
- Maintain our own hash map for the backwards conversion: Whenever `symbol(x)` is called with a string not representable in the native encoding, we still perform a (lossy) conversion to the native encoding, but store the original data in our hash table

lionel-:
> Interpret strings of the format "<U+xxxx>" as Unicode characters

hmm what do you mean? This format is only the result of the C function `encodeString()`, called when code is printed to the console. We might as well just look at the encoding tag.

lionel-:
ah I think I get it, you want to check for unrepresentable characters? Maybe `iconv()` would provide a more direct way to do it?

lionel-:
you were right yesterday. Now I see that R translates strings to native when they are assigned in an environment, e.g. with `assign()`. The idea of having our own symbol table is interesting.

krlmlr:
`enc2native()` and `Rf_translateChar` convert not representable characters to `<U+xxxx>`.

lionel-:
Got it. Seems to be the same code that will format `UTF-8` strings in non-UTF8 context, which got me confused. That looks parsable.

PR #41, opened 2017-02-23
Update expr API and introduce switchpatch()

lionel-:
* Renames `is_lang()` to `is_expr()` and refactors it so it's the union of `is_parsable_literal()` and `is_symbolic()`. Updated documentation reflects the new distinctions.

* Also renames `as_name()` to `as_symbol()`, so we respect the `typeof()` nomenclature.

* Adds `as_name()` which extracts the name of named calls. So now a name is always a string which is consistent with the names attribute.

* Adds `switchpatch()` and `type_of()` for clearer dispatching on base types.


R/lang.R
```diff
@@ -1,104 +1,144 @@
...
-#' empty environment. For instance, \code{quote(c(1, 2))} is not a
-#' literal, but the result of evaluating it in
-#' \code{\link{base_env}()} is (in this case an atomic vector).
-#' Technically, this sort of literal objects can be inlined in
-#' language expressions. If your function accepts arbitrary
+#' equivalents. \code{is_expr()} tests for expressions, the set of
+#' objects that can be obtained from parsing R code. An expression
+#' can be one of two things: either a symbolic object (for which
+#' \code{is_symbolic()} returns \code{TRUE}), or a parsable literal
+#' (testable with \code{is_parsable_literal()}).
```
hadley: Need a note here to mention `expression()`, and how that's different

R/lang.R
```diff
@@ -1,104 +1,144 @@
...
+#'
+#' Literals are the set of R objects that evaluate to themselves when
```
hadley: I think you need to emphasise:

* that any R object can be placed in a call
* this doesn't happen normally (when parsing text)
* symbols and calls have special behaviour

R/lang.R
```diff
@@ -1,104 +1,144 @@
...
+#' \code{quote(c(1, 2))} is not a literal, but the result of
+#' evaluating it in \code{\link{base_env}()} is (in this case an
+#' atomic vector).  Technically, this sort of literal objects can be
+#' inlined in language expressions. If your function accepts arbitrary
 #' expressions, it should thus account for that possibility with a
 #' catch-all branch. On the other hand, if your function only gets
 #' expressions created from a parse, \code{quote()}, or
 #' \code{\link{tidy_capture}()}, then you can check for literals with
-#' \code{is_literal()}.
+#' \code{is_parsable_literal()}.
```
hadley: I just realised we should turn markdown on for roxygen

R/lang.R
```diff
@@ -206,76 +244,71 @@ is_binary_call <- function(x, name = NULL) {
...
     FALSE
   }
 }
+
+expr <- function(x) {
+  if (is_quote(x)) f_rhs(x) else x
+}
+
+# More permissive than is_tidy_quote()
+is_quote <- function(x) {
```
hadley: Maybe fquote and tquote for short?

R/types.R
```diff
@@ -331,3 +331,101 @@ is_bare_integerish <- function(x, n = NULL) {
...
+    "primitive"
+  } else {
+    typeof(x)
+  }
+}
+
+#' Dispatch on base types.
+#'
+#' This is equivalent to
+#' \code{\link[base]{switch}(\link[base]{typeof}(x, ...))}.
```
hadley: `type_of`

R/lang.R
lionel-: is it now clearer?

lionel-:
After a bit more use and thought:

- should we have `switch_type()` and `switch_lang()`

- The pattern `switch_type()` + `switch_lang()` is so nice that I wonder if we should collapse all atomic types in one `atomic`. Then we'd have a `switch_atomic()` to deal with those.

lionel-:
> I wonder if we should collapse all atomic types in one "atomic"

An alternative is to use `switch_atomic()` in the catch-all clause of `switch_type()`:

```r
switch_type(x, .to = "target",
  string = foo(x)
  list = bar(x),
  switch_atomic(x, .to = "target"
    logical = ,
    integer = ,
    double = baz(x)
  )
)
```

lionel-:
Those `.to` are nice to uniformise error messages but feel a bit verbose. I suggest having `switch_` and `coerce_` variants. The latter would have target type as second argument, e.g.

```r
switch_type <- function(.x, ...) {
  switch(type_of(.x), ...)
}
coerce_type <- function(.x, .to, ...) {
  switch(type_of(.x), ..., abort_coercion(.x, .to))
}
```

This would also make the intent a bit clearer. And they have the same name length, it's a good match ;)

R/lang.R
```diff
@@ -206,76 +253,71 @@ is_binary_call <- function(x, name = NULL) {
...
-  }
+#' as_symbol(~ f())
+as_symbol <- function(x) {
+  switchpatch(x, .to = "symbol",
+    symbol = x,
+    string = symbol(x),
+    quote = as_symbol(f_rhs(x)),
+    language =
+      if (is_prefixed_name(x)) {
+        x
```
hadley: Doesn't this mean that `as_symbol()` will sometimes return things that are not symbols?

lionel-: yes that isn't ideal. It was done so we could pattern-match calls with namespaced symbols:

```r
expr <- quote(base::list(x, y))
is_call(expr, quote(base::list))
```

We probably need a better way.

lionel-:
@hadley if you're ok with the proposal above (i.e. have `switch_type()` and `coerce_type()`, I'd like to introduce these changes in this PR so the next PRs can use these functions.

PR #42, opened 2017-02-23
Update lang API and introduce switchlang()

lionel-:
~~Includes #41~~

* Renames call_ functions to lang_ so we respect the typeof types.

* Rename `lang_standardise()` to `lang_homogenise()` to reflect the fact that it's usually not the function a user is looking for (slow and rarely needed features).

* New faster version `lang_standardise()` which uses `call.match()`.

* Argument matching is now optional in `call_modify()`. This makes it easier to deal with new calls that don't necessarily have a closure defined.

* All these functions have better handling of tidy quotes.

* Adds `lang_head()` and `lang_tail()` for extracting components of a call.

* Introduce `switchlang()` to dispatch on call subtype. The categories are now: `named`, `namespaced`, `inlined` and `recursive`. An error is issued when something other than a function is inlined. This little function already proved quite handy.


R/arg-homogenise.R
```diff
@@ -0,0 +1,219 @@
+#' Standardise a call against formal arguments.
+#'
+#' This is a slower but more thorough version of
```
hadley: Can you add a hint here as to when you use it over `lang_standardise()`?  Is it worth documenting them together?

R/lang-call.R
```diff
@@ -1,232 +1,121 @@
...
-    missing_nms <- setdiff(fmls_nms, c(args_nms, "..."))
-    missing_args <- rep(list(arg_missing()), length(missing_nms))
-    missing_args <- as.pairlist(set_names(missing_args, missing_nms))
-    call <- lsp_append(call, missing_args)
-  }
-
-  call
+#' @rdname is_lang
+#' @export
+is_language <- is_lang
```
hadley: Do we need this alias?

lionel-: Maybe not, I thought it'd be nice to have a predicate that corresponds to the sexp type.

R/arg-homogenise.R
lionel-: Sure. I think they should be separate because probably no one will be interested in `lang_homogenise()`. It is only useful for tracing dots across the call stack.

PR #43, opened 2017-02-23
Translate Unicode escapes when converting from name to character

krlmlr:
Working but incomplete. Missing parts:

- [x] Tests in UTF-8 locale
- [x] CRAN-safe checking of UTF-8 mode via `l10n_info()`
- [x] Cleanup of helpers
- [x] Use `switch_type()` or `switchpatch`

Closes #62.


R/types-chr.R
```diff
@@ -86,10 +86,20 @@ chr <- function(x, encoding = NULL) {
 #' str_encoding(latin1)
 #' as_bytes(latin1)
 #' }
-as_character <- function(x, encoding = NULL) {
+as_character <- function(x, encoding = NULL) UseMethod("as_character")
```
lionel-: We've decided not to make anything generic in rlang. We'll have a `switch_type()` or `switchpatch()` function for dispatch but it's still in the PRs under review.

krlmlr:
I'm having second thoughts about parsing UTF-8 escapes.

Both `symbol("幸")` and `symbol("<U+5E78>")` map to the same symbol in a latin-1 locale, we cannot distinguish them after the mapping. If we define `as_character(symbol("幸")) == "幸"`, we fix the Unicode problem, but give headaches to users that have column names containing `"<U+xxxx>"`. This may occur if the user intentionally converted the column names to native encoding beforehand; I'm not sure how common this is.

So, instead of always parsing Unicode escapes, we could just provide a function that does this on demand, and another function that checks for presence of Unicode escapes.

Regardless of what we do here, we might need additional measures in dplyr:

- [x] Keep column names in UTF-8
- [ ] Reencode to native for lookup
- [ ] Warn if the reencoding produces a clash

In addition, `dplyr` can provide its own `utf8_symbol()` constructor that just creates a character vector with S3 class. This doesn't need to live in `rlang`, because it's specific to the application.

lionel-:
> Both symbol("幸") and symbol("<U+5E78>") map to the same symbol in a latin-1 locale, we cannot distinguish them after the mapping.

hmm I don't understand what you mean. The UTF-8 string you mention is the 5E78th unicode point, so what is the issue?

krlmlr:
`"幸"` and `"\u5E78"` are identical, but "<U+5E78>" is a string of 8 ASCII characters.

lionel-:
I still don't get it. The UTF-8 representation of 幸 is `e5 b9 b8`.

krlmlr:
So is the UTF-8 representation of `"\u5E78"`, but `"<U+5E78>"` maps to `3c 55 2b 35 45 37 38 3e`.

lionel-:
isn't it expected?

krlmlr:
Yes. But we now have two ways to arrive at `"幸"`:

- `as_character(as_symbol("幸")) == "幸"`
- `as_character(as_symbol("<U+5E78>")) != "<U+5E78>"`

We knew that before, but I wonder if there might be data frames out there where the column names are actually `"<U+5E78>"`. I guess there might be; therefore `dplyr` shouldn't rely on the roundtrip, regardless of what we do in `rlang`.

lionel-:
> I wonder if there might be data frames out there where the column names are actually "<U+5E78>"

ah ok, yes I was wondering about those cases, where the user manipulates unicode points in that exact format. I would say it would not be extremely bad if we just required users to use a UTF-8 enabled system when they manipulate a dataset of unicode points with DSLs.

In any case, I don't think keeping column names as strings in the whole pipeline is an option.

lionel-:
Let's shelve this issue for now until we hear about Tomas's plans. We may have to just accept that Windows latin1 users cannot use the tidyverse with unrepresentable UTF-8 datasets. It will still work for Windows CKJ users working in their own language, which is probably the most important use case.

lionel-:
>  We may have to just accept that Windows latin1 users cannot use the tidyverse with unrepresentable UTF-8 datasets.

If we take this road, `symbol()` should issue a warning when the string cannot be converted faithfully. Maybe with a hashtable of those strings so we only warn once for each string.

krlmlr:
If at all, I would make the warning optional.

I think we should make the tidyverse an example of proper UTF-8 usage, and not restrict Windows users. I think I know how to achieve this in `dplyr` with limited effort: https://github.com/hadley/rlang/pull/43#issuecomment-282241299 (above on the same page).

lionel-:
it's just not possible to keep column names as strings in the tidy evaluation framework. And I don't think introducing a surrogate symbol class is a satisfying or clean solution, so it won't help showcasing proper UTF-8 usage.

krlmlr:
We also need to reencode `names2()` and provide an `env_names()` function (https://github.com/hadley/dplyr/issues/2469#issuecomment-282492741).

We have two options regarding the UTF-8 locale flag:

1. Avoid reencoding in UTF-8 locales. Faster, but will give inconsistent results and minor interoperability problems (e.g., an environment serialized on Windows and deserialized on Linux).
2. Always reencode. This will give consistent results for all platforms, but even processing a string that doesn't contain `<` requires walking it once.

I'll encapsulate this decision in a separate function so we don't need to decide now.

lionel-:
I don't think we should have `names2()`, we only want to translate on `symbol → string` conversion right?

krlmlr:
Ideally we'd also support the `as_list(as_env(x))` route. We can do the translation either in `as_list()` (which we don't have yet) or in `names2()`.

Here's another example, with a slightly different name for the list element.

### Windows

``` r
x <- list(1)
names(x) <- "\u5E78<U+798F>"
x
#> $`<U+5E78><U+798F>`
#> [1] 1
x$"\u5E78\u798F"
#> [1] 1
x$"<U+5E78>\u798F"
#> [1] 1
x$"<U+5E78><U+798F>"
#> [1] 1
x["\u5E78\u798F"]
#> $<NA>
#> NULL
x["<U+5E78>\u798F"]
#> $<NA>
#> NULL
x["<U+5E78><U+798F>"]
#> $<NA>
#> NULL
x[["\u5E78\u798F"]]
#> [1] 1
x[["<U+5E78>\u798F"]]
#> [1] 1
x[["<U+5E78><U+798F>"]]
#> [1] 1
```

### Linux

``` r
x <- list(1)
names(x) <- "\u5E78<U+798F>"
x
#> $`幸<U+798F>`
#> [1] 1
x$"\u5E78\u798F"
#> NULL
x$"<U+5E78>\u798F"
#> NULL
x$"<U+5E78><U+798F>"
#> NULL
x["\u5E78\u798F"]
#> $<NA>
#> NULL
x["<U+5E78>\u798F"]
#> $<NA>
#> NULL
x["<U+5E78><U+798F>"]
#> $<NA>
#> NULL
x[["\u5E78\u798F"]]
#> NULL
x[["<U+5E78>\u798F"]]
#> NULL
x[["<U+5E78><U+798F>"]]
#> NULL
```

lionel-:
> Ideally we'd also support the as_list(as_env(x)) route.

Oh you're right. We probably need a special branch for environments in `as_list()` then. I have a `as_list()` implementation in a local branch already, so please don't write one, I'll change it later to use `env_names()` once this PR is merged.

lionel-:
@krlmlr we shouldn't unescape when we are taking the names of a vector. Maybe we should add a `sxp_names()` function in `attr.R` that returns `env_names()` for environments and `names()` otherwise. We'd use that in `names2()`. But it still doesn't feel right that `names()` and `names2()` have different behaviours. I think we should just return an error when `names2()` is supplied an environment.

For `as_list()` you can isolate the switchpatch branch for environments from the vector branches and unescape the names there.

e.g.

```r
function(x) {
  coerce_type_vec(x, "list",
    logical = ,
    integer = ,
    double = ,
    character = ,
    complex = ,
    raw = as_base_type(x, as.list),
    environment = {
      out <- as_base_type(x, as.list)
      set_names(out, <...>)
    },
    list = zap_attributes(x)
  )
}
```

krlmlr:
`names()` and `names2()` already behave differently.

I have tried the code you suggested but had to revert in d0b5beb simply because `coerce_type_vec()` overrides the names I'm assigning in this branch.

lionel-:
> names() and names2() already behave differently.

I'd prefer if `names2()` just didn't work with environments. Vectors and dictionaries have different semantics for names (ordered vs unordered, etc) and maybe it was a mistake to make that operator handle environments a few R versions ago. Do you have examples where we use `names2()` on environments?


> I have tried the code you suggested but had to revert

hmm in that case you can nest the `coerce_type_vec()` in the default case of a `switch_type()`, that will produce correct error messages:

```r
switch_type(x,
  environment = { },
  coerce_type_vec(x, "list",
    ...
  )
)
```

krlmlr:
The main difference between `names()` and `names2()` is that it always returns a vector of length `n` and also substitutes `NA` values. As you point out, both `names()` and also `names2()` work for environments. Can we just specify that `names2()` should be used instead of `ls()` and `objects()`?

I'll add a `switch_type()`.

lionel-:
> Can we just specify that names2() should be used instead of ls() and objects()?

I think we should suggest to use `env_names()`. Environments always have names so it doesn't make sense to use `names2()` on them, it's for pre-processing vector names.

krlmlr:
@lionel-: I think I'm done here. This now has a new `env_names()` function and unescapes Unicode in calls to `as_character()`, `as_list()`, and `env_names()`.

src/symbol.c
```diff
@@ -0,0 +1,167 @@
...
+
+int attribute_hidden process_byte(char* tgt, char* const src, int* len_processed) {
+  if (!has_codepoint(src)) {
+    // Copy only the first character (angle bracket or not), advance
+    *tgt = *src;
+    *len_processed = 1;
+    return 1;
+  }
+
+  int codepoint = strtol(src + strlen("<U+"), NULL, 16);
```
lionel-: I'd have thought this should be a `wchar_t` instead of an `int`, but it looks like `Rf_ucstoutf8()` takes unsigned ints. This should at least be unsigned otherwise we lose one bit.

src/symbol.c
```diff
@@ -0,0 +1,167 @@
...
+  if (re_enc != src) {
+    // If the string has been copied, it's safe to use as buffer
+    tmp = (char*)re_enc;
+  }
+  else {
+    // If not, we're in a UTF-8 locale
+    // Need to check first if the string has any UTF-8 escapes
+    if (!has_unicode_escape(src)) return name;
+    int orig_len = strlen(re_enc);
+    tmp = alloca(orig_len + 1);
```
krlmlr: I don't understand this remark. Which other allocator are you suggesting to use?

R/vector-coercion.R
```diff
@@ -140,12 +140,18 @@ as_complex <- function(x) {
...
 }
 #' @rdname vector-coercion
+#' @useDynLib rlang rlang_symbol_to_character
 #' @export
 as_character <- function(x, encoding = NULL) {
-  x <- coerce_type_vec(x, "character",
-    character = zap_attributes(x)
+  coerce_type_vec(x, "character",
+    character = chr_set_encoding(zap_attributes(x), encoding),
+    symbol = {
```
krlmlr: Why not in both?

src/symbol.c
lionel-: I think with C99 you can allocate arrays with variable size (runtime `n`):

```c
char tmp[orig_len + 1];
```

src/symbol.c
```diff
@@ -0,0 +1,167 @@
...
+// Private functions -----------------------------------------------------------
+
+bool has_unicode_escape(const char* chr);
+int unescape_char(char* chr);
+int unescape_char_found(char* chr);
+int process_byte(char* tgt, char* const src, int* len_processed);
+bool has_codepoint(const char* src);
+bool is_hex(const char chr);
+
+void copy_character(SEXP tgt, SEXP src, R_xlen_t len) {
```
krlmlr: I can't include it from symbol.c because I then get "duplicate definition" compiler errors. Please advise.

lionel-: The functions in vector.h need an `inline` annotation. Could you add them please?

krlmlr: Can't add `inline` to `namespace_rlang_sym()`, because it has a `static` local symbol. Shouldn't these functions rather live in a separate module?

lionel-: hmm yeah but we might convert those to C++ (for enum templating) and we might ship them as header-only library in the future. Don't worry about it, we can keep `copy_character()` in symbol.c and figure this out later.

lionel-:
Thanks for this Kirill! I hope this finally sorts out all dplyr encoding issues.

krlmlr:
Yeah, let's hope for the best. This PR didn't include any documentation, would you like to add it?

lionel-:
yes this should be a small section of the advanced tidyeval PR yet to be written. I'll add some roxygen examples as well while I'm writing it, though maybe that's too advanced.

PR #47, opened 2017-03-01
Export and document all pairlist helpers

lionel-:
They are useful to manipulate expression trees. E.g. I use them in dplyr to replace some uses of `lazyeval::interp()`.


lionel-:
I was thinking about using `node` instead of `lsp` to refer to both pairlists and cons cells. So `dots_node()`, `node_walk()`, etc.

If you think the helper names are too short, we could have `node_car()` / `set_node_tag()` instead of `car()` and `tag()`.

hadley:
`node()` seems reasonable, and I like using it as a prefix to `car()` etc.

PR #48, opened 2017-03-01
Coercion functions and NA predicates

lionel-:
- Coercion functions for vectors. They consistently zap attributes except names, they bypass base S3 methods, and avoid copying atomic vectors when not needed.

- NA predicates. The `is_` variants are scalar, and `are_` variants are vectorised.

- Environment functions: turn generic `env()` to a switchpatch and rename `new_env()` to `child_env()`, since new_ is for S3 stuff.


R/attr.R
```diff
@@ -200,3 +200,15 @@ has_length <- function(x, n = NULL) {
 with_attributes <- function(.x, ..., .attrs = list()) {
   invoke("structure", c(list(.Data = .x, ...), .attrs))
 }
+
+zap_attributes <- function(x) {
+  switch_type(x,
+    char = ,
+    symbol = ,
+    environment = abort(paste0(
```
hadley: Needs `NULL` too

R/vector.R
```diff
@@ -133,11 +134,59 @@ modify <- function(.x, ..., .elts = list()) {
...
+#'
+#' # There are typed versions as well:
+#' is_lgl_na(NA)
+#' is_lgl_na(na_dbl)
 #' @name missing
 NULL
 
 #' @rdname missing
 #' @export
+are_na <- is.na
```
hadley: I'm not sure I love this convention. It seems count to `is_logical` and `is_scalar_logical`

R/env.R
```diff
@@ -54,13 +54,13 @@
 #' env(fn)
 #'
 #'
-#' # new_env() creates by default an environment whose parent is the
```
hadley: I'd prefer to keep this as `new_env()`. It's not perfect because it's not S3, but I think it's better than using a different word

R/vector.R
lionel-: I remember that one of the challenges of learning R was to figure out what's vectorised and what's not. I think `are_na()` would be beneficial to newcomers.

`is_logical()` and `is_scalar_logical()` are predicates about vectors that can themselves be scalar or vectors. But they both are scalar predicates, unlike `are_na()`. Then, `is_na()` is akin to `is_null()`. And it provides additional intent and type check when you're expecting `NA` as an input.

R/env.R
lionel-: I think you're right in general, but in this case I like that `child_` conveys that an environment cannot appear out of nowhere, it always derives from another environment. I'm also considering removing the default argument so as to be always explicit: `child_env(NULL)` (with the understanding that `NULL` refers to the empty environment here). What do you think?

R/vector.R
lionel-: More on the difference: `NA` is always a scalar object, there are no `NA` vectors, there can only be vectors filled with `NA`s.

R/env.R
hadley: Yeah, but sometimes you just want a basic environment, not caring that it's a child of the empty environment. Maybe we could split the difference and have `new_env()` with no arguments as well as `child_env()`?

R/vector.R
hadley: Ok, I think you've persuaded me

R/env.R
lionel-: ok. I think this should be handled in another PR with a new interface for base types ctors, and look like this: `env(a = obj1, b = obj2)`. Merging this for now.

PR #49, opened 2017-03-02
Changes for the tidyverse port

lionel-:
Includes #48 

Various changes needed for porting tibble, dplyr and tidyr to tidyeval.


lionel-:
Now uses the "quosure" and "overscope" vocabulary.

R/env.R
```diff
@@ -552,6 +595,26 @@ env_clone <- function(x, parent = env_parent(x)) {
   list2env(as.list(x, all.names = TRUE), parent = parent)
 }
 
+#' Does environment inherits from another environment?
```
lionel-: This is not consistent with how we have written titles so far, e.g. all those "Is object an xxx" titles for the type predicates. Should we adopt this style from now on?

R/quosure.R
```diff
@@ -0,0 +1,166 @@
...
+#'
+#'
+#' # Formulas and definitions are not quosures:
+#' is_quosure(a := b)
+#' is_quosure(a ~ b)
+#'
+#' # But they are quosureish objects:
+#' is_quosureish(a := b)
+#' is_quosureish(a ~ b)
+#' @md
```
lionel-: Yes I was planning to port the whole codebase to markdown once all branches are merged.

R/env.R
hadley: I was just correcting your English, not making a general statement about grammatical forms :wink:

lionel-: But I got this grammatical form from your early purrr documentation ;)

it feels a bit like contracted newspaper titles so it made sense to me.

hadley: Well you can keep it, but definitely needs "inherits" -> "inherit"

lionel-: oh yes sorry

PR #54, opened 2017-03-08
WIP: Splicing and a C++ish API

lionel-:
I thought I'd start working on the R wrapper API while implementing the backend for splicing. It uses a subset of C++:

- Templates and overloading for genericity and propagating type information
- Namespaces to organise code
- POD-only so no dtors issues during long jumps
- Export to R is accomplished by writing a switchpatch function with a pure C signature


lionel-:
TODO: names and implicit coercions

hadley:
I don't love the new C++ API here - it doesn't feel like it substantially simplifies the code, and makes it harder to read because I have to mentally translate.

lionel-:
I think the templating on sexptype does make the code simpler, and it should produce better compiled code. It's helpful for copying, coercing, etc.

As for translating that's unavoidable at first. What's important is whether the new API feels natural. After two days on it I think it does and I particularly like the namespacing that indicates the expected type of inputs, e.g. `vec::names()` should be used with vectors.

lionel-:
Also one goal would be to use as much of rlang conventions as possible so it's easier to switch to C/C++ after prototyping in R.

hadley:
Yeah, I agree in the long-run, it just feels like this is the wrong place to start.  I think you could substantially simplify the C++ side to only simplify to lists (leaving off the atomic vectors for now), and then we can discuss the bigger API choices later.

PR #55, opened 2017-03-10
Implement C backend for splicing 

lionel-:
- This adds a number of base-type constructors with splicing semantics. Coercion is currently handled with an R callback to rlang's atomic coercers.

- `dots_list()` and `dots_splice()` are aliases to `splice()` that provide a place to document the two different splicing semantics of `...`.


R/dots.R
```diff
@@ -1,11 +1,91 @@
...
+#' @seealso [dots_list()] for explicit splicing semantics, and
+#'   [dots_exprs()] for extracting dots without evaluation.
+#' @export
+#' @examples
+#' # dots_splice() splices lists marked with spliced() as well as bare
+#' # lists:
+#' x <- list(1, 2)
+#' dots_splice(spliced(x), 3)
+#' dots_splice(x, 3)
+dots_splice <- function(...) {
```
hadley: I think this should be documented with `dots_list()`.

R/dots.R
```diff
@@ -1,11 +1,91 @@
...
+  splice(..., .bare = TRUE)
+}
+
+#' Extract dots forwarded as arguments.
+#'
+#' These functions return the arguments forwarded through `...`.
+#' Contrarily to [dots_list()] and [dots_splice()], `dots_exprs()` and
+#' `dots_node()` do not evaluate the arguments. The former returns a
+#' list of expressions while the latter returns a [pairlist].
```
hadley: And do call-splicing?

src/splice.c
```diff
@@ -0,0 +1,221 @@
...
+
+
+typedef struct {
+  R_len_t size;
+  bool named;
+  bool warned;
+} splice_info_t;
+
+void splice_warn_names(void) {
+  Rf_warning("Outer names are only allowed for unnamed scalar atomic inputs");
```
hadley: I mildly prefer `Rf_warning_call(R_NilValue, ...)`

PR #58, opened 2017-03-14
WIP: Extract and export node_modify() from lang_modify()

lionel-:
For `purrr::partial()` rewrite.

I'm not sure if `node_modify()` should take a list instead of dots. That wouldn't be consistent with other `_modify` functions, but that would make sure it doesn't trip on this puzzling bug in R argument matching:

```r
fn <- function(node, ...) node
fn(pairlist(n = arg_missing()), n = "foo")
#> [1] "foo"
```

That seems to only happen for arguments named `n` and affected a purrr example with `runif()`.


hadley:
Isn't that just partial matching?

I think all `_modify` functions should take spliced dots.

lionel-:
> Isn't that just partial matching?

hmm that's part of it. Since it's no longer needed for `partial()` I'm closing this for now, we can figure it out when we next need it.

PR #59, opened 2017-03-15
Another iteration over API naming

lionel-:
- Rename `env()` to `get_env()`, and add new `env()` constructor that creates a new child of the current environment and takes its data as dots (with splicing semantics).

- Unabbreviate `new_` ctors, e.g. `new_fn()` is now `new_function()`. Also `lang()` is now `new_language()`

- `tidy_quote()` and `tidy_quote_expr()` have been renamed to `quosure()` and `expr()`. The previous definition of `quosure()` is now `new_quosure()`.

- `tidy_capture()` and `tidy_quotes()` have been renamed to `arg_quosure()` and `dots_quosures()`. I hope this makes it clearer where the quosures are coming from. E.g. `quosure()` is about the data (code) you supply as input, while `arg_quosure()` is about the data supplied to the argument. `tidy_defs()` is now `dots_definitions()`.

- `parse_f()` and `parse_fs()` are now `parse_quosure()` and `parse_quosures()`.

The plan for documenting the tidyeval functions is to have central documentation about quasiquotation in one place. Then `quosure()`, `expr()`, `arg_quosure()`, `dots_quosures()` and `dots_exprs()` would mention that they support unquoting and link to that page.


lionel-:
`have_names()` is now `have_name()` since a vector element can only have one name.

hadley:
How about `tidy_interp()`?

lionel-:
I didn't rename `tidy_eval()` and `tidy_interp()`. The latter could be just `interp()` or maybe `unquote()`?

lionel-:
should `expr_eval()` be `bare_eval()`? to mirror `tidy_eval()` and suggest that it does something simpler than `eval()`.

lionel-:
> Should tidy_interp() be quo_interp()? Or quo_unquote()?

hmm probably not because it's a switchpatch. So `interp()`?

hadley:
Oh yeah, in that case, `interp()` sounds good.

hadley:
BTW now that purrr and dplyr are using rlang, can you bump the version number when you change the API?

lionel-:
in the end I think we shouldn't reserve unprefixed `interp()` as it's only generic over expression wrappers. So I renamed it to `expr_interp()` which seems fitting

lionel-:
maybe the `eval_` functions should be `ctxt_` after all, to get better autocomplete

lionel-:
should `symbols()` be renamed to `sym_list()`? I much prefer the former but it clashes with `graphics::symbols()` :/

hadley:
Should `arg_quosure()` and `dots_quosures()` simply become `quosure()` and `quosures()`?

I like the `sym()` + `syms()` idea.

lionel-:
> Should arg_quosure() and dots_quosures() simply become quosure() and quosures()?

I think it's more natural with regard to our new convention for ctors that `quosure()` is the quoting operator: it creates a quosure from its input passed by expression, just like `env()`, `chr()` etc create a new data structure based on their inputs. Then `arg_quosure()` takes an argument name and returns a quosure, and `dots_quosures()` takes dots and returns a list of quosures. Does that feel consistent?

We could also have an alias `quosures()` for `dots_quosures()` since it kind of have double use: you can use it to capture dots as quosures and you can use it to capture a list of arguments supplied directly. The same way `splice()` and `dots_splice()` are aliases.

lionel-:
@hadley should we try the `mut_` idea? We'd use prefix `mut_` instead of `set_` everywhere there is side effects. So `mut_node_car()`, `mut_mbcs_locale()`, etc.

hadley:
I like the `mut()` idea

PR #62, opened 2017-03-19
Avoid test failures on systems that don't have a MBCS locale

krlmlr:
by aborting early on warnings in `set_mbcs_locale()`.


lionel-:
Closing since that is included in #43

PR #63, opened 2017-03-20
Fix typo

krlmlr:
in `env_get()`.


lionel-:
oops thanks, I'll fix it in #59 because `env()` has been renamed.

PR #64, opened 2017-03-20
Fix formula self eval

lionel-:
* This guards formulas on capture so we are compatible with purrr's lambdas cc @jennybc.

* This streamlines capture of literals so they are not rewrapped in a quosure. This should make it easier for instance to pass default `NULL` arguments between multiple NSE functions and check for a `NULL` quosure.


lionel-:
Merging now as dplyr's rlang update depends on the new formula behaviour.

PR #71, opened 2017-03-23
New names for quosure functions + S3 quosure class

lionel-:
No class for quosure lists yet.

Also:

- `with_attributes()` is now `struct()` and `zap_attributes()` is `unstruct()`
- `env_set()` renamed to `set_env()`. If called without argument, sets the current env.


lionel-:
@hadley just pushed three more commits to close your latest requests

R/quosure.R
```diff
@@ -211,6 +217,13 @@ quo_label <- function(quo) {
 quo_text <- function(quo, width = 60L, nlines = Inf) {
   expr_text(quo_expr(quo), width = width, nlines = nlines)
 }
+#' @export
+quo_name <- function(quo) {
+  if (!is_symbol(quo)) {
+    abort("`quo` must quote a symbol")
```
hadley: We need to do something here, not throw an error

lionel-: what's the specification for this function?

hadley: I'm not entirely sure. Something like:

* if name, convert to character
* if call, do something like `expr_label()` but wrap with `()`
* if atomic scalar, convert to character

Anything else can be an error

lionel-: do we need a name that reflects better the intent behind that function? otherwise users might be confused between all the options: `_text()`, `_label()`, `_name()`. I'm not sure I understand what is this function for.

The `_text()` versions might be multi-line so you want something that fits on one line is that right? Should `expr_label()` be `expr_format()`, and then `quo_name()` would be `quo_label()` (and we'd have `expr_label()` version etc)?

hadley: It's used when you want to turn an expression into a name that you can use for (e.g.) a dplyr column (or can easily be pasted together with something else). I think `expr_label()` is fine as is : it's a label for use in error messages.

lionel-: does this look right? https://github.com/hadley/rlang/pull/71/commits/2c10421c162ec509f63b5383f6a0892a13918797

R/expr.R
```diff
@@ -245,9 +248,36 @@ expr_label <- function(expr) {
...
-
+#' @rdname expr_label
 #' @export
+expr_name <- function(expr) {
+  name <- switch_type(expr,
+    symbol = as_name(expr),
+    quosure = ,
+    language = {
+      expr_text <- deparse(expr)
+      paste0("(", expr_text, ")")
```
hadley: This needs the same logic as `expr_label`, i.e. insert `...` if the call is too long

hadley: Otherwise looks good

PR #72, opened 2017-03-24
Fix dplyr issues

lionel-:
Capturing quosures now:

- has support for missing arguments by returning an empty quosure
- doesn't rewrap quosured literals on repeated captures

This helps simplifying some code in dplyr. It is also more efficient to avoid rewrapping literals (fewer quosure self-evaluation), and I wonder if we should also avoid rewrapping quosures?

In addition, there's now support in `quos()` for removing all empty arguments, none of them, or only trailing ones (the default). This fixes https://github.com/hadley/dplyr/issues/1039


R/tidy-quote.R
```diff
@@ -262,9 +262,13 @@ expr <- function(expr) {
 #'   name. If an integer, it is passed to the `width` argument of
 #'   `expr_text()`, if `TRUE`, the default width is used. See
 #'   [exprs_auto_name()].
-#' @return A list of quosures with class `quosures`.
-#' @name quosures
+#' @param .ignore_empty Whether to ignore empty arguments. If `TRUE`,
```
hadley: I think I'd prefer either `c("trailing", "all", "none")` or to use `NA` for the default trailing removal

tests/testthat/test-dots.R
```diff
@@ -43,5 +43,11 @@ test_that("empty dots return list()", {
 })
 
 test_that("dots_exprs() captures empty arguments", {
-  expect_identical(dots_exprs(, ), set_names(list(missing_arg(), missing_arg()), c("", "")))
+  expect_identical(dots_exprs(, , .ignore_empty = FALSE), set_names(list(missing_arg(), missing_arg()), c("", "")))
+})
+
+test_that("dots are always named", {
+  expect_identical(dots_list("foo"), set_names(list("foo"), ""))
```
hadley: Could also use `expect_named()` here

PR #79, opened 2017-03-29
Update dots API

lionel-:
~~Includes #77~~

- Use splicing semantics everywhere we take dots

- Add `ll()` list constructor. Or should we deprecate and rename `tibble::lst()` to `libble()`?

- Add `exprs()` alias and `enexpr()`

- Remove SE versions of `with_handlers()` and `with_restarts()`, now that we have unquoting.


hadley:
I'm not sure if `spliced()` feels quite right to me. Maybe `splice_in()`?

lionel-:
Maybe this should be `splice()` (and we'd have to find another name for the original `splice()` function)?

hadley:
Yeah, that sounds good.

lionel-:
it's going to be a bit painful for a while because purrr already exports a function called `splice()`. But since this operator is going to be common in tidyverse functions taking dots, it may be worth it.

hadley:
Let's kill `purrr::splice()` then

PR #82, opened 2017-04-01
Extend type_of() doc

krlmlr:
to cover `type_of(quote(a))`.


lionel-:
This bullet list is for differences with `typeof()`. Should we have a list just below it for the other types?

krlmlr:
Never mind, I was confused:

``` r
class(quote(a))
#> [1] "name"
```

lionel-:
yes S3 dispatches on the mode not the type :/

hadley:
We may want to have a `type_of()` variant with names tailored for used in error messages

PR #84, opened 2017-04-06
Update splice

lionel-:
This generalises the splicing code to handle flattening (1 level) and squashing (all levels) of lists. Compared to purrr's flatten functions, they have better handling of names:

```r
purrr::flatten_dbl(list(a = 1))
#> [1] 1

rlang::flatten_dbl(list(a = 1))
#> a 
#> 1 
```

And the typed flatten actually do flatten. In purrr they are merely `list_to_type()` coercers:

```r
purrr::flatten_dbl(list(a = 1, list(b = 2)))
#> Error: Can't coerce element 2 from a list to a double

rlang::flatten_dbl(list(a = 1, list(b = 2)))
#> a b 
#> 1 2 
```

Splicing, flattening and squashing have the same handling of names. They all provide an error message if they see a conflict between outer and inner names:

```r
purrr::flatten_dbl(list(a = c(A = 1)))
#> A 
#> 1 

rlang::flatten_dbl(list(a = c(A = 1)))
#> A 
#> 1 
#> Warning message:
#> Outer names are only allowed for unnamed scalar atomic inputs 
```

Also, this introduces `flatten_if()` and `squash_if()` which take a predicate. This predicate can be a closure, a pointer object, or a native info object.

Finally, the `!!!` operator now uses value-based splicing instead of call-based for dots passed by value.


R/dots.R
```diff
@@ -45,11 +41,34 @@ dots_list <- function(...) {
...
-  dots <- .Call(rlang_splice, dots_values(...), "list", bare = TRUE)
+  dots <- .Call(rlang_squash, dots_values(...), "list", is_spliced_bare, 1L)
   names(dots) <- names2(dots)
   dots
 }
 
+#' Evaluate dots with preliminary splicing.
+#'
+#' This is a tool for advanced users. It captures dots, processes
+#' unquoting and splicing operators, and evaluate them. Unlike
```
hadley: "evalutates"

PR #86, opened 2017-04-08
Rewrite predicate_pointer()

krlmlr:
- safe aborting in case of error
- now also accepts an external pointer wrapped in a list
- ~~doesn't care about the length of a `NativeSymbolInfo`~~
- still accepts a naked external pointer, not sure about the usefulness here

See remarks in https://github.com/hadley/rlang/commit/04a9bce7f474f6740f9c6e3661ccb81b270c1ed9.

### DIFF
--- a/src/splice.c
+++ b/src/splice.c
@@ -189,7 +189,7 @@ static
 is_spliceable_t predicate_pointer(SEXP x) {
   switch (TYPEOF(x)) {
   case VECSXP:
-    if (Rf_inherits(x, "NativeSymbolInfo")) {
+    if (Rf_inherits(x, "NativeSymbolInfo") && Rf_length(x) >= 3) {
       SEXP ptr = VECTOR_ELT(x, 1);
       if (TYPEOF(ptr) == EXTPTRSXP) {
         return (is_spliceable_t) R_ExternalPtrAddr(ptr);
--- a/src/splice.c
+++ b/src/splice.c
@@ -187,25 +187,29 @@ bool is_spliced(SEXP x) {
 
 static
 is_spliceable_t predicate_pointer(SEXP x) {
-  is_spliceable_t is_spliceable;
-
   switch (TYPEOF(x)) {
   case VECSXP:
-    if (Rf_inherits(x, "NativeSymbolInfo") && Rf_length(x) == 3) {
+    if (Rf_inherits(x, "NativeSymbolInfo")) {
       SEXP ptr = VECTOR_ELT(x, 1);
       if (TYPEOF(ptr) == EXTPTRSXP) {
-        is_spliceable = (is_spliceable_t) R_ExternalPtrAddr(ptr);
-        break;
+        return (is_spliceable_t) R_ExternalPtrAddr(ptr);
+      }
+    } else if (Rf_length(x) == 1) {
+      SEXP ptr = VECTOR_ELT(x, 0);
+      if (TYPEOF(ptr) == EXTPTRSXP) {
+        return (is_spliceable_t) R_ExternalPtrAddr(ptr);
       }
     }
-  case EXTPTRSXP:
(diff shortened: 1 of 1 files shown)


lionel-:
> now also accepts an external pointer wrapped in a list

hmm I don't like this UI. We can PROTECT external pointers if you're worried they get GC'd. I assume that's the reason you want to pass it in a list?

> doesn't care about the length of a NativeSymbolInfo

Could you still check for length with `>` please.

krlmlr:
I'm pretty sure we need to wrap in a list when we call `flatten_if()` from R in dplyr.

lionel-:
> I'm pretty sure we need to wrap in a list when we call flatten_if() from R in dplyr.

Could you explain why is that the case please?

krlmlr:
In https://github.com/tidyverse/dplyr/pull/2626/commits/b9b2bfaae4054e4a0af0190ee7b8ff801617761e I'm using a getter function to obtain the address of `bind_spliceable` in the R code, so that it can be used in the `flatten_if()` call. The getter function wraps the external pointer in a list, because otherwise it may be gc-ed at any point. But now I see that the call itself can work with a naked xptr.

lionel-:
if the pointer is in a list or in an environment (including a dynamic frame) it should never get GC'd, so we're probably fine. We can also PROTECT and UNPROTECT it to be extra safe.

krlmlr:
We should be fine with https://github.com/tidyverse/dplyr/pull/2626/commits/29f2f43c23f342e8602df155f018c85b7f792466, even without protection; but we just can't change `get_is_bind_spliceable()` to return a naked xptr.

lionel-:
> we just can't change get_is_bind_spliceable() to return a naked xptr.

IIUC, this is safe:

```r
ptr <- get_ptr()
flatten_if(x, ptr)
```

But this isn't:

```r
flatten_if(x, get_ptr())
```

So we should PROTECT to make the above safe.

krlmlr:
I doubt the first is safe, but it is if we wrap:

```r
wrapped_ptr <- get_wrapped_ptr()
flatten_if(x, ptr[[1L]])
```

lionel-:
> I doubt the first is safe

how can an object get GC'd if it's bound in an environment?

krlmlr:
From R-exts:

> [External pointers] are unusual in their copying semantics in that when an R object is copied, the external pointer object is not duplicated. (For this reason external pointers should only be used as part of an object with normal semantics, for example an attribute or an element of a list.)

The second sentence reads like "in R, always wrap external pointers" to me.

In a previous version the bindrcpp package collected naked xptrs in an environment, this didn't work well.

Other than that, I don't understand the mechanisms behind xptrs very well.

lionel-:
> In a previous version the bindrcpp package collected naked xptrs in an environment

was the environment unprotected at any point? I don't understand well that second sentence in the quote. There are other uncopyable types in R and they have predictable GC semantics. Interesting though, I'd like to get to the bottom of this :)

krlmlr:
It's mostly R code, but C++ code returned an XPtr. From what I remember, the problem was:

```r
a <- get_ptr()
b <- a
rm(a)
gc() # a is freed, b becomes invalid or a NULL pointer
```

lionel-:
I cannot reproduce your example. Also after a cursory reading of GC code, I can't see why external pointers would be freed unexpectedly, I think they should be refcounted or named just like any other SEXP. Are you sure the environment in which the pointers were assigned wasn't freed at some point?

Also I interpret the text you quoted from the manual as a warning that modifying external pointers has side effects:

```r
ptr <- get_ptr()

# This changes ptr by reference
structure(ptr, foo = "bar")
```

hadley:
My understanding is inline with @lionel- — the warning in R-exts is just echoing that external pointers have reference semantics, so if you want to add attributes, you need wrap.

PR #91, opened 2017-04-10
New approach for C plugins

lionel-:


lionel-:
cc @krlmlr

hadley:
@jimhester should probably take a look too

lionel-:
@jimhester the context is to have an easy way of providing C-level methods to R-level functions to avoid R callbacks in loops.

I previously used native symbol objects for this but this wasn't a good idea because setting `registration = TRUE` changes internal details in those objects. Tthey become pointers to struct containing a function pointer, instead of being direct pointers to the function pointer.

So this PR introduces a simple S3 object containing an external pointer to a function pointer. These pointers are created using the same strategy as R 3.4.0's new tool `Rf_MakeExternalPtrFn()`, e.g. an union that ensures that the function pointer has the same size as data pointers (which is otherwise not true on all platforms).

lionel-:
Maybe we should also call `R_RegisterCCallable()` with the pointer, so it's available both from R and from C.

src/splice.c
```diff
@@ -187,25 +188,23 @@ bool is_spliced(SEXP x) {
...
+  case EXTPTRSXP:
+    return (is_spliceable_t) rlang_ExternalPtrAddrFn(x);
+
   default:
-    Rf_errorcall(R_NilValue, "`predicate` must be a closure or external pointer");
+    break;
   }
 
-  return is_spliceable;
+  Rf_errorcall(R_NilValue, "`predicate` must be a closure or function pointer");
```
jimhester: I believe this will throw a warning if `-Wreturn-type` is on because there is no explicit return at the end of the function, even though control will never reach there.

tests/testthat/test-vector.R
```diff
@@ -163,8 +163,10 @@ test_that("flatten_if() handles external pointers", {
   obj <- set_attrs(list(1:2), class = "foo")
   x <- list(obj, splice(obj), unclass(obj))
 
-  pred <- is_clevel_spliceable$address
-  expect_identical(flatten_if(x, pred), list(obj[[1]], splice(obj), unclass(obj)))
+  expect_identical(flatten_if(x, test_is_spliceable), list(obj[[1]], splice(obj), unclass(obj)))
+
+  ptr <- test_is_spliceable[[1]]
+  expect_identical(flatten_if(x, ptr), list(obj[[1]], splice(obj), unclass(obj)))
```
jimhester: Maybe worth it to verify test_is_splicable inherits from 'fn_pointer', just to prevent future breakage.

src/export.c
```diff
@@ -0,0 +1,36 @@
...
+
+// Compat for R 3.4.0
+SEXP rlang_MakeExternalPtrFn(DL_FUNC p, SEXP tag, SEXP prot) {
+  fn_ptr ptr;
+  ptr.fn = p;
+  return R_MakeExternalPtr(ptr.p, tag, prot);
+}
+DL_FUNC rlang_ExternalPtrAddrFn(SEXP s) {
+  fn_ptr ptr;
+  ptr.p =  EXTPTR_PTR(s);
```
jimhester: There is an extra space here ;)

src/export.c
```diff
@@ -0,0 +1,36 @@
...
+}
+
+void rlang_register_pointer(const char* ns, const char* ptr_name, DL_FUNC fn) {
+  SEXP ptr = PROTECT(rlang_MakeExternalPtrFn(fn, R_NilValue, R_NilValue));
+
+  SEXP ptr_obj = PROTECT(Rf_allocVector(VECSXP, 1));
+  SET_VECTOR_ELT(ptr_obj, 0, ptr);
+
+  SEXP class = PROTECT(Rf_mkString("fn_pointer"));
+  Rf_setAttrib(ptr_obj, R_ClassSymbol, class);
```
jimhester: If you assign the string object directly you don't need the `PROTECT()` here or name the temp variable.

`Rf_setAttrib(ptr_obj, R_ClassSymbol, Rf_mkString("fn_pointer"))`

src/export.c
```diff
@@ -0,0 +1,36 @@
+#define R_NO_REMAP
+#include <Rinternals.h>
+
+#include "export.h"
+
+// Compat for R 3.4.0
+SEXP rlang_MakeExternalPtrFn(DL_FUNC p, SEXP tag, SEXP prot) {
```
jimhester: I realize this is functionally equivalent to the R 3.4.0 definition, but it does have an extra function call, why are we not using the [exact definition from R](https://github.com/wch/r-source/blob/37388191f5493825923b98246ec02e9b13ab1770/src/main/memory.c#L3347-L3356)?

src/export.c
```diff
@@ -0,0 +1,36 @@
+#define R_NO_REMAP
+#include <Rinternals.h>
+
+#include "export.h"
+
+// Compat for R 3.4.0
```
jimhester: If these functions are only for compatibility with R versions prior to R 3.4.0 we should conditionally define them `if defined(R_VERSION) && R_VERSION >= R_Version(3, 3, 0)` and use the R definitions if they are available.

src/splice.c
```diff
@@ -187,25 +188,23 @@ bool is_spliced(SEXP x) {
...
-    if (Rf_inherits(x, "NativeSymbolInfo") && Rf_length(x) == 3) {
-      SEXP ptr = VECTOR_ELT(x, 1);
-      if (TYPEOF(ptr) == EXTPTRSXP) {
-        is_spliceable = (is_spliceable_t) R_ExternalPtrAddr(ptr);
-        break;
-      }
+    if (Rf_inherits(x, "fn_pointer") && Rf_length(x) == 1) {
+      SEXP ptr = VECTOR_ELT(x, 0);
+      if (TYPEOF(ptr) == EXTPTRSXP)
+        return (is_spliceable_t) rlang_ExternalPtrAddrFn(ptr);
```
jimhester: I always put conditional bodies in braces, it is too easy for someone to insert an additional statement between the condition and the expression with matched indentation and mess up the control flow.

src/export.c
lionel-: because this assignment is forbidden:

```c
EXTPTR_PTR(s) = tmp.p;
```

src/export.c
lionel-: jim the lintr :)

src/splice.c
lionel-: hmm, I've always used braceless bodies in C if both branches have only statement (e.g. Linux kernel style). If we switch we should do it for the whole package, though I think I prefer the braceless style.

src/splice.c
lionel-: is it ok to return this?

```c
  return (is_spliceable_t) 0;
```

lionel-: oh I can just return `NULL` actually.

src/splice.c
jimhester: linux kernel style also uses 8 character tabs ;), It is very easy for someone else to write

```c
if (TYPEOF(ptr) == EXTPTRSXP)
  Rf_warning("External ptr");
  return (is_spliceable_t) rlang_ExternalPtrAddrFn(ptr);
```

krlmlr: An omission like this would be caught e.g. by astyle(), or by a C code linter.

src/splice.c
krlmlr: `Rf_errorcall()` is annotated with `__attribute__((noreturn))`, perhaps should be `NORET`... Then modern compilers would understand.

src/splice.c
jimhester: > An omission like this would be caught e.g. by astyle(), or by a C code linter.

Not if they are not run ;)

Why do you prefer astyle to clang-format?

krlmlr: Just plain inertia. dplyr and RSQLite run astyle in their tests, works for me.

PR #92, opened 2017-04-11
Fix remaining issues with dictionary class

lionel-:
- Rename `data_source()` to `as_dictionary()`. `is_dictionarish()` is now the predicate for testing uniquely named vectors or environments, and `is_dictionary()` tests for dictionary S3 class.

- Provide replacement methods. The dictionary ctors and coercers have a `read_only` argument to allow replacement. Tidyeval `.data` pronoun is set to read only. `.env` is left as a pure environment because its purpose is now technical, end users should use `!!` to evaluate in the current env.

- Provide a callable entry point for dictionary ctor.

- Export and document the `.data` pronoun.

- Improve the pronoun print method. I didn't put too much effort into it because these pronouns are not typically used interactively.

cc @krlmlr


R/attr.R
```diff
@@ -30,34 +30,34 @@ zap_attrs <- function(x) {
 #' Is object named?
 #'
 #' `is_named()` checks that `x` has names attributes, and that none of
-#' the names are missing or empty (`NA` or `""`). `is_dictionary()`
+#' the names are missing or empty (`NA` or `""`). `is_dictionarish()`
```
hadley: I think it should be`is_dictionaryish()`

R/dictionary.R
```diff
@@ -0,0 +1,169 @@
...
+  new_dictionary(list(), lookup_msg, read_only)
+}
+#' @export
+as_dictionary.environment <- function(x, lookup_msg = NULL, read_only = FALSE) {
+  lookup_msg <- lookup_msg %||% "Object `%s` not found in environment"
+  new_dictionary(x, lookup_msg, read_only)
+}
+#' @export
+as_dictionary.data.frame <- function(x, lookup_msg = NULL, read_only = FALSE) {
+  lookup_msg <- lookup_msg %||% "Variable `%s` not found in data"
```
hadley: This should probably be `` "Column `%s` not found in data" `` to be consistent with dplyr

R/eval.R
```diff
@@ -82,6 +82,26 @@ eval_tidy <- function(f, data = NULL) {
...
+#'
+#' You can import this object in your package namespace to avoid `R
+#' CMD check` errors when referring to overscoped objects.
+#'
+#' @name tidyeval-data
+#' @export
+#' @examples
+#' quo <- quo(.data$foo)
+#' eval_tidy(quo, list(foo = "bar"))
+.data <- structure(class = "dictionary", list(
```
hadley: It would be safer to generate this with `as_dictionary()` to ensure that data structure stays the same

lionel-: there is a timing problem because the `.Call`able ctor is not available at build time :/

hadley: You could use delayed assign 😉

PR #93, opened 2017-04-11
Use friendly type in error messages

lionel-:
@hadley this currently treats vectors and calls specially:

- vectors are suffixed with "vector", e.g. "an integer vector"
- language are suffixed with "call", e.g. "a language call"

Is this missing anything else?

Also, do we need our own `arg_match()` to get consistent error messages for matched arguments?


R/types.R
```diff
@@ -432,11 +431,57 @@ coerce_class <- function(.x, .to, ...) {
...
+
+  type <- switch(type,
+    logical = ,
+    integer = ,
+    double = ,
+    complex = ,
+    character = ,
+    raw =
+      paste(type, "vector"),
+    language = "language call",
```
hadley: I think as a general principle it's better to right code like:

```R
switch(type, 
  logical = "a logical vector",
  integer = "an integer vector", 
  ...
)
```

Otherwise it's a basically impossible to translate to other languages.

lionel-: I didn't know we supported translation. So we can't ever generate error messages automatically? That's going to be rigid for extension, e.g. here it automatically supports `expr_type_of()` and any future ones.

lionel-:
does this now look good to be exported? So we get proper error messages in dplyr:

```
Error: `.id`: must be a scalar string, not integer of length 1
```

→

```
Error: `.id`: must be a scalar string, not an integer vector of length 1
```

R/types.R
```diff
@@ -431,10 +431,89 @@ switch_class <- function(.x, ...) {
...
+    integer = "an integer vector",
+    numeric = ,
+    double = "a double vector",
+    complex = "a complex vector",
+    character = "a character vector",
+    raw = "a raw vector",
+    string = "a string",
+    list = "a list",
+
+    NULL = "the NULL object",
```
hadley: I think this can just be `NULL`

R/types.R
```diff
@@ -431,10 +431,89 @@ switch_class <- function(.x, ...) {
...
+    primitive = ,
+    builtin = ,
+    special = "a primitive function",
+    closure = "a function"
+  )
+}
+
+friendly_lang_type_of <- function(type) {
+  switch(type,
+    named = "a named language call",
```
hadley: "language call" seems a bit ugly to me. A quoted call? An unevaluated call? Just a call?

R/types.R
```diff
@@ -507,7 +586,7 @@ switch_lang <- function(.x, ...) {
 #' @rdname switch_lang
 #' @export
 coerce_lang <- function(.x, .to, ...) {
-  msg <- paste0("Cannot convert objects of type `", type_of(.x), "` to `", .to, "`")
+  msg <- paste0("Cannot convert ", type_of(.x), " to ", .to, "")
```
hadley: I think we standardised on contractions (i.e. "can't") in dplyr

R/types.R
lionel-: so it's not important to make the connection to the "language"  type?

hadley: Oh yeah. How about "A (language) call" ?

OTOH:

> While it is usually better to avoid using S terminology, it would probably be even more confusing to refer to "calls" as "language".

lionel-: I think I'd prefer to make the connection to language when we can so that people can understand the prefix `lang_`. Parentheses looks like a good option.

hadley: In that case, maybe "A call (lang)" would be better.

PR #99, opened 2017-04-13
Add `.ignore_empty` arg to dots_list() and dots_splice()

lionel-:
So all functions taking dots by value now trim trailing empty arguments by default. E.g.

```r
chr("a", )
#> [1] "a"
```


lionel-:
Do we want to support and document the fact that supplying `.ignore_empty` to user-facing functions is eventually passed to `dots_values()`?

```r
ll( , 1, , .ignore_empty = "all")
#> [[1]]
#> [1] 1
```

PR #100, opened 2017-04-13
Treat formulas literally and remove UQF()

### LINKED ISSUE #74: Get rid of UQF() ?
Now that quosures have an S3 class, we'd guard all unquoted bare formulas.

lionel-:
Closes #74 

With this change, bare formulas are always treated literally.

Now that quosures and bare formulas are treated quite differently, I've changed the quosure print method to make it explicit that an object is a quosure.

```r
quo(foo)
#> # A quosure from the global environment
#> ~foo

new_quosure(quote(foo), base_env())
#> # A quosure from the base environment
#> ~foo

locally(quo(foo))
#> # A quosure from a local environment (0x7fa2e0faf400)
#> ~foo

fn <- function() print(quo(foo))
fn()
#> # A quosure from a frame environment (0x7fa2ea323ed8)
#> ~foo
```


lionel-:
`type_of()` still yields "quosure" for any one-sided formulas. Should this now be "quosureish" or should we only treat proper quosures as their own type? The latter option seems to make more sense.

lionel-:
> I use `<my_type>`

What's the preferred format for auxiliary information?

```
<quosure>
~foo(bar)
# from the global environment
```

This would be consistent with stacks and frames print methods:

```
<quosure> [global]
~foo(bar)
```

hadley:
Or maybe

```R
<quosure [global]> 
~foo(bar)
```

R/quo.R
```diff
@@ -411,10 +412,23 @@ as_quosure <- function(x, env = caller_env()) {
...
 }
+
 #' @export
 print.quosure <- function(x, ...) {
-  x <- set_attrs(x, class = "formula")
-  NextMethod()
+  env_type <- env_type(get_env(x))
+
+  if (env_type %in% c("frame", "local")) {
+    addr <- sxp_address(get_env(x))
```
hadley: I have mixed feelings about printing the environment address — it doesn't add much information and adds spurious diffs when re-rendering markdown.  OTOH it does help indicate that evaluating too things that otherwise look identical will yield different results.

lionel-:
```
<quosure [local: 0x7fa2e99ec038]>
~foo(bar)
```

and then `ctxt_frame()` would print this?

```
<frame 0 [global]>
  expr: NULL
  env: <environment: R_GlobalEnv>
```

R/quo.R
lionel-: I'm not sure either. It does makes the printout look more technical. And the address can still be printed with `get_env()` when needed.

hadley:
```R
<frame 0 [global]>
  expr: NULL
  env: [global]
```

?

R/quo.R
lionel-: One option is to print the address in an `str()` method.

hadley: I like that idea

lionel-:
does that look good?

```
## locally(quo(foo))
<quosure [local]>
~foo

## str(locally(quo(foo)))
<quosure [local 0x7fa2eba13868]>
~foo
```

```
## identity(identity(ctxt_stack()))
[[1]]
<frame 2> (0)
  expr: identity(ctxt_stack())
   env: [local 0x7fa2e87b85f8]

[[2]]
<frame 1> (0)
  expr: identity(identity(ctxt_stack()))
   env: [local 0x7fa2e87b8710]

[[3]]
<frame 0> [global]
  expr: NULL
   env: [global]
```

lionel-:
I like that the frame printout now makes it clear which frames are still on the stack.

lionel-:
e.g.:

```
## fn <- function() print(identity(ctxt_stack()))
## fn()
[[1]]
<frame 3> (1)
  expr: identity(ctxt_stack())
   env: [local 0x7fa2ebad7238]

[[2]]
<frame 2> (1)
  expr: print(identity(ctxt_stack()))
   env: [frame 0x7fa2ebad73c0]

[[3]]
<frame 1> (0)
  expr: fn()
   env: [frame 0x7fa2ebad7510]

[[4]]
<frame 0> [global]
  expr: NULL
   env: [global]
```

hadley:
Looks good. But I think you should left align the labels:

```
  expr: fn()
  env:  [frame 0x7fa2ebad7510]
```

PR #102, opened 2017-04-14
Fix tidyeval formulas

lionel-:
- Formulas and quosures are now treated separately in tidyeval. Only the latter self-evaluate. Formulas are evaluated (as formulas) in the overscope (more precisely, the parent of the bottom layer since the bottom contains tidyeval definitions for `~` etc).

- Improve genericity of exprs, envs, and quosures a bit.

- Better print method for quosures: `<quosure: local>`. I think the local vs global qualifiers are a bit confusing. How about printing "local" if it's a quosure from the current environment, "frame" if it's a quosure from a frame on the stack, and something else when it's a quosure from outside the current stack? Maybe "external".


NAMESPACE
```diff
@@ -25,6 +25,7 @@ S3method(rst_muffle,simpleMessage)
 S3method(rst_muffle,simpleWarning)
 S3method(str,dictionary)
 S3method(str,quosure)
+export("!!")
```
hadley: Do we need to export `!!!` to?

R/env.R
```diff
@@ -132,16 +135,24 @@ child_env <- function(parent = NULL, data = list()) {
...
+  out <- switch_type(env,
     environment = env,
     quosure = attr(env, ".Environment"),
     primitive = base_env(),
     closure = environment(env),
-    string = pkg_env(env),
-    list = coerce_class(env, target, frame = env$env)
+    list = switch_class(env, frame = env$env)
   )
+  if (!is_null(out)) {
```
hadley: I think this would be a little cleaner as:

```R
out <- out %||% default

if (is_null(out)) {
   ...
}

out
```

R/formula.R
```diff
@@ -127,8 +127,7 @@ f_env <- function(f) {
 #' @rdname f_rhs
 `f_env<-` <- function(x, value) {
   stopifnot(is_formula(x))
-  f <- new_formula(f_lhs(x), f_rhs(x), value)
-  copy_lang_name(f, x)
+  set_attrs(x, .Environment = value)
```
hadley: Is this a special feature of `set_attrs()` or can you do it with `structure()` too?

R/quo-enquo.R
```diff
@@ -89,16 +89,14 @@ enquo <- function(x) {
   forward_quosure(expr, arg$env)
 }
 forward_quosure <- function(expr, env) {
-  if (quo_is_missing(expr)) {
+  if (inherits(expr, "quosure")) {
```
hadley: `is_quosure()` ?

R/formula.R
lionel-: you can do it with structure as well:

```r
x <- ~foo
y <- structure(x, .Environment = empty_env())

x
#> ~foo
y
#> ~foo
#> <environment: R_EmptyEnv>
```

PR #106, opened 2017-04-19
Formulas and quosures

lionel-:
##### Handling of quosures

- expr predicates no longer take the RHS of formulas and quosures.

- `type_of()` returns formula for both quosures and formulas. It seems to make more sense to treat all formulas as a base type rather than just quosures.

##### Formula predicates

- the predicates for formulas, quosures and definitions are now more consistent. `is_quosureish()` returns TRUE for quosures and one-sided formula. `is_formulaish()` returns TRUE for definitions and formulas (it is the most general predicate of the family).

- `is_formula()` and `is_formulaish()` gain an `lhs` argument to pattern-match on the presence or absence of LHS.

- `is_formula()` returns TRUE for quosures, but the new `is_bare_formula()` predicate doesn't.

##### Removed functions

- `as_generic_expr()`, `eval_tidy_lhs()`, `eval_tidy_rhs()`

- `as_symbol()`. Its main purpose was to take the CAR of a call if it is a symbol, but it doesn't seem to be an intuitive way of doing this.

- `as_lang()`. It was used to transform symbols to a call, but it seems better to use `lang()` for that purpose. It also used to parse strings into calls, which we now advise against.

##### Pattern-match on namespace in is_lang()

`as_symbol()` was used as part of a suboptimal scheme for handling namespaces in `is_lang()`. This has been replaced by a `ns` argument that allows pattern-matching on namespaces. If `NULL` (the default), the namespace doesn't participate in the match and so `is_lang(quote(foo::bar()), "bar")` and `is_lang(quote(bar()), "bar")` are now equivalent.


R/expr-sym.R
```diff
@@ -24,51 +24,28 @@ syms <- function(x) {
...
-#' @export
-#' @rdname as_symbol
+#' quo <- quo(sym)
+#' as_name(get_expr(quo))
 as_name <- function(x) {
-  switch_type(x,
+  coerce_type(x, "a name",
     string = x,
-    as_string(as_symbol(x))
+    symbol = as_string(x)
```
hadley: This seems to return a string now?

lionel-: in rlang names are about strings (unlike the old S mode). It seems better to keep a clear separation between symbols and names that way, e.g. the names attributes are a character vector, and `as_name()` coerces symbols to strings.

lionel-: oops, just realised that now we've removed `as_symbol()`, there is no sense in keeping `as_name()` since we have `as_string()`. I'll remove it as well.

hadley: Ok great, I'd variable consistently use symbols and strings, rather than allow using names, which is confusing.

PR #114, opened 2017-05-01
Polish API and documentation for lang and expr functions

lionel-:


R/expr-lang.R
```diff
@@ -1,7 +1,7 @@
 #' Create a language call by "hand"
 #'
-#' @param .fn Function to call. Must be a callable object: a string,
-#'   symbol, call, or a function.
+#' @param .fn Function to call. Must be a callable object: symbol,
```
hadley: A string still is ok though? (And I think it should be for convenience)

R/expr-lang.R
```diff
@@ -40,8 +40,35 @@ lang <- function(.fn, ..., .ns = NULL) {
...
+#' is not legal to manually set language heads to strings.
+#'
+#' @param x An object to test.
+#' @export
+#' @examples
+#' # Symbolic objects and functions are callable:
+#' is_callable(quote(foo))
+#' is_callable(base::identity)
+#'
+#' # You can safely set the head of a language node to a callable
```
hadley: I don't get what you're trying to say here

R/expr-lang.R
```diff
@@ -203,17 +231,17 @@ is_binary_lang <- function(x, name = NULL, ns = NULL) {
 #' # You can also modify quosures inplace:
 #' f <- ~matrix(bar)
 #' lang_modify(f, quote(foo))
-lang_modify <- function(.call = caller_frame(), ..., .standardise = FALSE) {
+lang_modify <- function(.lang = caller_frame(), ..., .standardise = FALSE) {
```
hadley: I wonder if these have diverged sufficiently from the original function that the first argument should no longer have a default.

R/expr-lang.R
```diff
@@ -326,41 +356,43 @@ lang_fn <- function(call = caller_frame()) {
...
-#' Return the head or tail of a call object.
+#' Return the head or tail of a call object
 #'
 #' @description
 #'
-#' Internally, calls are structured as a tree of expressions (see
-#' [switch_lang()] and [pairlist] documentation pages). A `language`
-#' object is the top level node of the tree. `lang_head()` and
-#' `lang_tail()` allow you to retrieve the node components.
+#' Internally, calls (`language` objects) are structured as a tree of
```
hadley: This seems inconsistent with how we use "language" elsewhere. I think you need to somehow qualify this is what base R calls these things?

R/expr-lang.R
```diff
@@ -326,41 +356,43 @@ lang_fn <- function(call = caller_frame()) {
...
+#'   checks that the head is a symbol and converts it to a string).
 #'
 #' * The second component of the tree node contains the arguments. The
-#'   type of arguments is _pairlist_. Pairlists are actually
+#'   type of arguments is _pairlist_. Pairlists are structurally
 #'   equivalent to `language` objects (calls), they just have a
-#'   different name. `lang_tail()` returns the pairlist of arguments
-#'   ([lang_args()] returns the same object converted to a regular
-#'   list).
+#'   different type name. `lang_tail()` returns the pairlist of
```
hadley: Remove name?

R/expr-lang.R
lionel-: oops yes it is still ok there

R/expr-lang.R
lionel-: `mut_node_car()` lets you put anything in a call and that can lead to crashes in some circumstances. `is_callable()` is a predicate that checks the node head is safely callable.

lionel-:
I will fix the lang vs call thing tomorrow

PR #115, opened 2017-05-02
Update of env API and doc

lionel-:
- merge `env_assign()`, `env_bind()` and `env_define()` in one function `env_bind()` that takes dots with splicing syntax

- rename `env_assign_promise()` and `env_assign_active()` to `env_bind_exprs()` and `env_bind_fns()`. They now take dots. The former takes dots with `exprs()` and supports unquoting.

- `env_bury()` now takes dots as well.

- Refactoring of documentation.


R/env.R
```diff
@@ -1,191 +1,125 @@
...
+#' Create a new environment
+#'
+#' `env()` and `child_env()` create new environments. `env()` always
+#' creates a child of the current environment while `child_env()` lets
+#' you specify a parent (see section on inheritance).
+#'
+#' @section Environments as objects:
+#'
+#' Environments are containers of uniquely named objects. Their most
+#' common use is to provide a scope for the evaluation of R
```
hadley: I think you should say "first-class environments" somewhere in here. I think it's a good name for this property of R.

R/env.R
```diff
@@ -1,191 +1,125 @@
...
-#' @param parent A parent environment. Can be an object with a S3
-#'   method for `as_env()`.
-#' @param default The default environment in case `env` does not wrap
-#'   an environment. If `NULL` and no environment could be extracted,
-#'   an error is issued.
-#' @param ...,data Uniquely named bindings. See [is_dictionaryish()].
-#'   Dots have [explicit-splicing semantics][dots_list].
-#' @param n The number of generations to go through.
-#' @seealso `scoped_env`, [env_has()], [env_assign()].
+#' Create a new environment
```
hadley: This documentation is great!

R/env.R
```diff
@@ -257,31 +191,153 @@ as_env_ <- function(x, parent = NULL) {
   list2env(x, parent = parent %||% empty_env())
 }
 
+#' Get parent environments
+#'
+#' `env_parent()` returns the parent environment of `env` if called
```
hadley: Do we want `env_parents()` too? (i.e. return all parent envs)

R/env.R
```diff
@@ -311,176 +367,220 @@ mut_parent_env <- function(env, new_env) {
...
-#' @param nm The name of the binding.
-#' @param x The value of the binding.
-#' @param ... Pairs of unique names and R objects used to define new
-#'   bindings.
-#' @return The input object `env`, with its associated environment
+#' Bind symbols to objects in an environment
+#'
+#' @description
+#'
+#' These functions create bindings in an environment. These bindings
```
hadley: These bindings -> The bindings

R/env.R
```diff
@@ -311,176 +367,220 @@ mut_parent_env <- function(env, new_env) {
...
+#' are supplied through `...` as pairs of names and values or
```
hadley: "values" -> "values, "

R/env.R
```diff
@@ -311,176 +367,220 @@ mut_parent_env <- function(env, new_env) {
...
+#' expressions. `env_bind()` is equivalent to evaluating a `<-`
+#' expression within the given environment, e.g. `with_env(env, foo <-
+#' "bar")`. This function should take care of the majority of use
```
hadley: Maybe: "Note this is not the same as `assign()`, which evaluates the value in the current environment, not the target environment"

R/env.R
```diff
@@ -311,176 +367,220 @@ mut_parent_env <- function(env, new_env) {
...
+#' cases but the other variants can be useful for specific problems.
+#'
+#' - `env_bind()` takes named _values_. The arguments are evaluated
+#'   once (with [explicit splicing][dots_list]) and bound in `.env`.
+#'
+#' - `env_bind_fns()` takes named _functions_ and creates active
```
hadley: Should note that this is equivalent to `makeActiveBinding`

R/env.R
```diff
@@ -311,176 +367,220 @@ mut_parent_env <- function(env, new_env) {
...
+#'   (see the implementations of `dplyr::do()` methods for an
+#'   example).
+#'
+#' - `env_bind_exprs()` takes named _expressions_. The arguments are
+#'   captured with [exprs()] (and thus support call-splicing and
+#'   unquoting) and assigned to symbols in `.env`. These expressions
+#'   are not evaluated immediately but lazily. Once a symbol is
+#'   evaluated, the corresponding expression is evaluated in turn and
+#'   its value is bound to the symbol (the expressions are thus
+#'   evaluated only once, if at all).
```
hadley: Should note this is equivalent to `delayedAssign()`

R/env.R
```diff
@@ -311,176 +367,220 @@ mut_parent_env <- function(env, new_env) {
...
+#'
+#' @details
+#'
+#' Since environments have reference semantics (see relevant section
+#' in [env()] documentation), modifying the bindings of an environment
+#' produces effects in all other references to that environment. In
+#' other words, `env_bind()` and its variants have side effects.
```
hadley: Maybe this should be `@section Side effects` and since it's called primarily for its side-effects, it should return the first arg invisibly.

R/env.R
hadley: Unless you actually mean something different?

R/env.R
```diff
@@ -311,176 +367,220 @@ mut_parent_env <- function(env, new_env) {
...
-  args <- list(
-    x = nm,
-    value = get_expr(expr),
-    eval.env = get_env(expr, eval_env),
-    assign.env = get_env(env)
-  )
-  do.call("delayedAssign", args)
-}
+#' # You can unquote expressions. Note that quosures are not
+#' # supported, only raw expressions:
```
hadley: Why aren't quosures supported?

R/env.R
```diff
@@ -311,176 +367,220 @@ mut_parent_env <- function(env, new_env) {
...
+
+  for (i in seq_along(fns)) {
+    makeActiveBinding(nms[[i]], fns[[i]], env_)
+  }
+
+  .env
 }
 
-#' Bury bindings and define objects in new scope.
+#' Overscope bindings by defining symbols deeper in a scope
```
hadley: I think this needs to start with a verb

R/env.R
```diff
@@ -311,176 +367,220 @@ mut_parent_env <- function(env, new_env) {
...
 #'
-#' `env_bury()` is like `env_bind()` but it creates the bindings in a
-#' new child environment. Note that this function does not modify its
-#' inputs.
+#' `env_bury()` is like [env_bind()] but it creates the bindings in a
+#' new child environment. This makes sure the new bindings have
+#' precedence over old ones, without altering existing environments.
+#' Unlike `env_bind()`, this function does not have side effects.
```
hadley: ; it returns a new environment.

R/env.R
```diff
@@ -574,29 +676,30 @@ env_get <- function(env = caller_env(), nm, inherit = FALSE) {
...
 #' @examples
-#' env_names()
+#' env <- env(a = 1, b = 2)
+#' env_names(env)
 #' @useDynLib rlang rlang_unescape_character
-env_names <- function(env = caller_env()) {
-  env_ <- get_env(env)
-  x <- names(env_)
-  .Call(rlang_unescape_character, x)
+env_names <- function(env) {
```
hadley: Above you talk about binding symbols to objects. I think being consistent would help?

lionel-: This function returns the symbols as strings. The name of a symbol is a string in the terminology I've used in rlang. I don't know how to call those otherwise. I'll add some explanation about strings versus symbols, and why environments bind values to symbols rather than strings. Might also be a good place to talk about encoding issues.

R/env.R
lionel-: Thanks!

R/env.R
lionel-:
```r
#' @section Names of symbols and objects:
#'
#' Technically, objects are bound to symbols rather than strings,
#' since the R interpreter evaluates symbols (see [is_expr()] for a
#' discussion of symbolic objects versus literal objects). However it
#' is often more convenient to work with strings. In rlang
#' terminology, the string corresponding to a symbol is called the
#' _name_ of the symbol (or by extension the name of an object bound
#' to a symbol).
#'
#' @section Encoding:
#'
#' Note that there are deep encoding issues when you convert a string
#' to symbol and vice versa. Symbols are _always_ in the native
#' encoding (see [set_chr_encoding()]). If that encoding (let's say
#' latin1) cannot support some characters, these characters are
#' serialised to ASCII. That's why you sometimes see strings looking
#' like `<U+1234>`, especially if you're running Windows (as R doesn't
#' support UTF-8 as native encoding on that platform).
#'
#' To alleviate some of the encoding pain, `env_names()` always
#' returns a UTF-8 character vector (which is fine even on Windows)
#' with unicode points unserialised.
```

cc @krlmlr

krlmlr: > as R doesn't support UTF-8 as native encoding on that platform

Nitpick: Windows doesn't support UTF-8 as a native encoding for any application, not just for R.

lionel-: but specific applications can support UTF-8 on Windows, e.g. text editors using UTF-8 as working encoding, like RStudio. R could support it as well if someone was willing to put the work in.

lionel-:
@hadley I think that instead of an `inherit` argument, we should have `env_sees()` and `env_fetch()` as variants to `env_has()` and `env_get()`. This distinction comes up often and the semantics are sufficiently different that it probably deserves separate functions.

R/types.R
```diff
@@ -685,7 +685,7 @@ is_copyable <- function(x) {
 is_equal <- function(x, y) {
   identical(x, y)
 }
-#' @useDynLib rlang rlang_is_identical
-is_identical <- function(x, y) {
-  .Call(rlang_is_identical, x, y)
+#' @useDynLib rlang rlang_is_reference
+is_reference <- function(x, y) {
```
lionel-: @hadley Just renamed `is_identical()` to `is_reference()`. I think that function can be generally useful and that name feels right. Would it be ok to export it?

lionel-: an alternative would be `is_alias()`

hadley: Let's still leave for now. We'd definitely want `are_alias()` too

hadley: Maybe `is_aliased()` and `are_aliased()`?

hadley: `is_aliased(x)` could return `NAMED(x) > 2`?  (although that would require some gymnastics to make sure we don't take a reference inside the function)

R/env.R
lionel-: We're capturing the expression. If we say we support quosures, that means we allow users to unquote quosures into the captured expression. Unless we evaluate the promise in an overscope, that will only work at top-level and therefore will probably be confusing.

R/types.R
lionel-: that's a good idea. I think one goal of rlang should be to make it easy to prototype C-side code. This should also help new developers get acquainted with R internals (e.g. playing with the `node_` functions to explore the structure of expressions).

PR #120, opened 2017-05-05
Import new set_names() features from purrr

lionel-:
Also:

- fix `set_names(x, NULL)`
- pass names to `chr()` to gain splicing semantics
- pass `...` to `nm` when it's a function

This exports `dots_n()` though I ended up removing it from `set_names()` implementation.


R/attr.R
```diff
@@ -163,25 +163,56 @@ has_name <- function(x, name) {
...
+set_names <- function(x, nm = x, ...) {
   if (!is_vector(x)) {
-    stop("`x` must be a vector", call. = FALSE)
+    abort("`x` must be a vector")
+  }
+
+  if (is_function(nm) || is_formula(nm)) {
+    nm <- as_function(nm)
+    nm <- nm(names2(x), ...)
+  } else if (!is_null(nm)) {
```
hadley: I think you can drop the `is_null()` stuff.  I'm now not sure why I thought it was necessary

lionel-: it is necessary so this zaps names: `set_names(mtcars, NULL)`

hadley: Oh I forgot about that case

R/attr.R
```diff
@@ -163,25 +163,56 @@ has_name <- function(x, name) {
...
+#'
+#'   * If you do nothing, `x` will be named with itself.
+#'
+#'   * If `x` already has names, you can provide a function or formula
+#'     to transform the existing names. In that case, `...` is passed
+#'     to the function.
+#'
+#'   * In all other cases, `nm` and `...` are passed to [chr()]. This
+#'     gives implicit splicing semantics: you can pass character
+#'     vectors or list of character vectors indistinctly.
```
hadley: This needs:

> Alternatively, you can supply `NULL` to remove all existing names.

PR #122, opened 2017-05-05
Add pkgdown website

lionel-:
- Categorise functions in the reference file
- Fix documentation issues revealed by the website
- Remove `dots_exprs()` and `dots_quos()` aliases to avoid people using different function names


hadley:
I'd say just publish it and then I can review the site.

PR #125, opened 2017-05-06
Fix a wrong vignette index

yutannihilation:
Congratulations for the release!

I found a weird index name for the vignette "Tidy evaluation". (Maybe is this a mistake when you copied from another existing vignette?)

> Vignettes: 	Specifying fonts
https://cran.r-project.org/web/packages/rlang/index.html


lionel-:
looks like it, thanks!

PR #127, opened 2017-05-08
Evaluate tilde calls in overscope

lionel-:
Fixes https://github.com/tidyverse/dplyr/issues/2745


R/eval-tidy.R
```diff
@@ -342,3 +339,12 @@ f_self_eval <- function(overscope, overscope_top) {
     .Call(rlang_eval, f_rhs(f), overscope)
   }
 }
+tilde_eval <- function(f, env) {
+  # Inline the base primitive because overscopes override `~` to make
+  # quosures self-evaluate
+  mut_node_car(f, base::`~`)
+
+  # Change both old and new formulas to original head avoid side-effects
+  on.exit(mut_node_car(f, sym_tilde))
```
hadley: I worry about this implementation as it seem unlikely to interact well with the byte code compiler

lionel-: what do you mean? We're manipulating calls in other places.

hadley: We're mutating calls in place and then re-mutating them back??

lionel-: hmm that, probably not, but we do manipulate calls in place. Most (all?) of the time after duplication though.

The compiler doesn't do anything to tilde calls currently, and I don't think it ever will because it would need runtime information to do so. If you prefer I can just duplicate the tilde call so we're sure.

lionel-: > If you prefer I can just duplicate the tilde call so we're sure.

Done in last commit.

lionel-:
is this now good @hadley?

R/eval-tidy.R
```diff
@@ -342,3 +339,16 @@ f_self_eval <- function(overscope, overscope_top) {
...
 }
+tilde_eval <- function(f, env) {
+  if (is_env(attr(f, ".Environment"))) {
+    return(f)
+  }
+  f <- duplicate(f, shallow = TRUE)
+
+  # Inline the base primitive because overscopes override `~` to make
+  # quosures self-evaluate
+  mut_node_car(f, base::`~`)
```
hadley: This still feels risky to me. Couldn't we construct a new call and evaluate that? 

Something like this?

```R
eval_bare(call(`~`, !!! f[-1]), env)
```

lionel-: that's what we are now doing, constructing a new call (by shallow duplication). I don't think we need full duplication. On the other hand that's only for bare formulas, not for quosures, so it doesn't matter as much. Also we need to inline the primitive function because `~` has a different definition in the overscope.

hadley: Oh I see what you mean now. Why don't you like my version? It seems simpler to me.

lionel-: `lang()` uses value-splicing, so it converts the pairlist to a list, then back to a pairlist, which incurs many allocations.

lionel-: At least that's will happen once value-splicing supports pairlists.

hadley: Ah ok, got it. This is good to go then.

lionel-: Maybe we need `new_language()`, which would have a bit more type checking than `node()`.

lionel-:
@hadley I've tweaked `is_pairlist()` and added `is_node()` in order to implement `new_language()`. The latter is now used in `tilde_eval()`.

PR #129, opened 2017-05-09
Add new_env()

lionel-:


lionel-:
@hadley just noticed that this function breaks the rule that functions taking dots are not prefixed with `new_`.

hadley:
Since `new_env()` is a strict constructor, not a helpful constructor, I don't think it needs to take `...`. That would match how I typically use it.

PR #131, opened 2017-05-09
Registration tweaks

jimhester:
The real fix is the second commit (29c542f), using `Rf_install()` in the init block was causing the problem, in fact this crashes with a simple `devtools::load_all()` in R compiled with `-O0`. The first commit simply passes the address of the functions rather than the functions directly. The latter wasn't producing an error, but every other use I could find passes the address so that is clearly the intended use.


lionel-:
Could you set the symbols in static variables then, or just use `Rf_install()` without intermediate values? The latter is probably good enough. I was just trying to avoid some interning in functions that might be called a lot because part of quasiquotation.

Too bad we can't set up such constants in packages. Did it cause issues outside of devtools?

jimhester:
I think it is going to be dangerous anytime you are unloading a namespace. The overhead is mainly just a hash lookup if the symbol is already in the symbol table, so will be pretty quick, I don't think performance is a major concern here.

lionel-:
Agreed, but I think it still makes sense to use the init function for other things than symbol registration, so it would be nice if devtools supported it, if possible at all.

Thanks for the PR!

lionel-:
>  The overhead is mainly just a hash lookup if the symbol is already in the symbol table

After the hash lookup there's also a linear search through a linked list (whose contents are thus scattered in memory) by the way. Tomas has been patching R to avoid this kind of repeated lookups. I did not do any benchmarking about these issues though, I agree this normally shouldn't matter much.

jimhester:
Well yes, you have to deal with hash collisions of course. This is compounded by the fact that the symbol hash size `HSIZE` is extremely low (`4119`) and hasn't been [increased in 19 years](https://github.com/wch/r-source/blob/trunk/src/include/Defn.h#L306).

Using a simple function 
```c
#define HSIZE 4119
extern SEXP* R_SymbolTable;

SEXP rlang_symbol_lengths() {
  SEXP lengths = PROTECT(Rf_allocVector(INTSXP, HSIZE));
  for (int i = 0; i < HSIZE; ++i) {
    INTEGER(lengths)[i] = Rf_length(R_SymbolTable[i]);
  }
  UNPROTECT(1);
  return lengths;
}
```

you can return the sizes for each bucket and see the average load per bucket is quite high (I simply ran `library(tidyverse)` to populate the symbol table).

``` r
setwd("~/p/rlang")
pkgload::load_all()
#> Loading rlang
#> unloadNamespace("rlang") not successful, probably because another loaded package depends on it. Forcing unload. If you encounter problems, please restart R.
suppressPackageStartupMessages(library(tidyverse))
lens <- .Call(rlang_symbol_lengths)
sum(lens)
#> [1] 18880
summary(lens)
#>    Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
#>   0.000   3.000   4.000   4.584   6.000  14.000
hist(lens)
```



It seems to me an easy way to speed up symbol lookup would be to increase `HSIZE` a decent amount (maybe 10x), which would remove the linked list traversal for a higher percentage of the cases with only relatively minimal increase in memory cost.

lionel-:
> Well yes, you have to deal with hash collisions of course.

Is it common to deal with collisions with a linear search? My first intuition would be that binary search is more appropriate.

hadley:
I think the intuition for a linear search is that you expect the collisions to be few, so you don't expect the added complexity of a binary search to pay off.

lionel-:
there's also that line that might be problematic:

```c
  rlang_register_pointer("rlang", "test_is_spliceable", (DL_FUNC) &is_clevel_spliceable);
```

@jimhester could we safely call our own init function from `.onLoad()`?

jimhester:
@lionel- cool, I made a small package with some additional benchmarks of the performance as well https://github.com/jimhester/hashsize#readme, was planning on writing a post on r-devel to see if we can get the size increased so the load factor is more like .75 instead of 4.75 for a typical use case.

PR #132, opened 2017-05-09
Fix coverage comments

jimhester:
`nocov` ignores just the line it is on, you need to use `nocov start` to
define a block to ignore


lionel-:
could you update the revision in those files please?

lionel-:
thanks!

PR #139, opened 2017-05-14
Fix typos, insert omitted links

egnha:
Looks like roxygen's markdown syntax was used to link to rlang docs (e.g., `[caller_env()]`), but that doesn't seem to be recognized in (the current version of) knitr. So such links have been manually inserted (e.g., `` [`caller_env()`](http://rlang.tidyverse.org/reference/caller_env.html) ``).

### DIFF
--- a/vignettes/tidy-evaluation.Rmd
+++ b/vignettes/tidy-evaluation.Rmd
@@ -53,7 +53,9 @@ were not necessarily envisioned or planned by their designers.
 ## Parsing versus evaluation
 
 There are two ways of dealing with unevaluated expressions to create a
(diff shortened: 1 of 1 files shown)


hadley:
These will be linked by pkgdown on the rlang website, not roxygen2.

egnha:
I see, I didn't realize that. Then I just want to point out that the typos and link omissions remain in the article on the rlang website (as well as in the vignette).

lionel-:
Could you send another PR that fixes the typos please?

egnha:
Would be glad to.

PR #144, opened 2017-05-17
Fix capture of optimised promises

lionel-:
This fixes the "Argument was already evaluated" error of the lisy and epidata packages.

This one is kind of serious. Should we make another rlang release before dplyr's?

cc @krlmlr

### DIFF
--- a/src/capture.c
+++ b/src/capture.c
@@ -26,6 +26,10 @@ SEXP attribute_hidden capture_arg(SEXP x, SEXP env) {
 }
 
 SEXP attribute_hidden capture_promise(SEXP x, int strict) {
+    // If promise was optimised away, return the literal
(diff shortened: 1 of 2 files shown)


krlmlr:
Can you give a dplyr example that fails?

lionel-:
@krlmlr any functions that takes named arguments (not dots) by expression. This includes:

* all scoped verbs with a `.funs` argument because it's taken both by value and by expression
* functions like `sample_n()`

If these functions are part of another compiled function and are supplied constants, this triggers the bug. E.g.:

```r
# `weight` passed a constant
compiler::cmpfun(function() sample_n(mtcars, 2, weight = NULL))

# `.funs` passed a constant
compiler::cmpfun(function() mutate_all(mtcars, "identity"))
```

PR #149, opened 2017-05-18
Support for R 3.1

lionel-:
Looks like devtools isn't compatible with 3.1 because of `trimws()`. I have added a compat that could be used to fix this.


lionel-:
I have added a `port_back()` function to conditionally import backports. This way devtools can just call that to get a `trimws()` backport (see https://github.com/hadley/devtools/pull/1508)

I think it's safer if we rely maintain our own backports, it shouldn't be a lot of work. cc @krlmlr @gaborcsardi

gaborcsardi:
I still think that it would be nice to have a `backports` package that we can use everywhere.

krlmlr:
@gaborcsardi: What license would that have use?

lionel-:
> I still think that it would be nice to have a backports package that we can use everywhere, not just in devtools.

rlang is meant to be used everywhere, it will always be a light dependency. No deps and quick compilation are parts of rlang specifications.

gaborcsardi:
Sure, it makes sense.

jimhester:
Any backports package basically has to be GPL-2 or 3 as the code will be derived directly from R sources.

gaborcsardi:
@jimhester unless you don't look at the sources. But that's a pain, and I don't mind if it is GPL, as long as it is in another package and it does not make my package GPL.

hadley:
Can we get this merged and plan for a CRAN release in the near future? We need it in order to get devtools working on 3.1 (a vision circle)

Issue #32, opened 2017-02-15
C API for messages


jimhester:
Similar to https://github.com/hadley/rlang/issues/31, there is not a C API for messages, would be nice to have one.

PR #152, opened 2017-05-23
C API for conditions

### LINKED ISSUE #30: In abort() etc, should call be an integer?
Sometimes when signalling an error, it's useful to use the call one level up the stack (i.e. in `check()` style functions). Maybe as well as being an logical, `call()` could also be an integer with the same interpretation as in `call_stack()` etc?

### LINKED ISSUE #31: C API for signaling conditions
There isn't currently a way to signal a custom condition from C, it would be nice to define functions in rlang to do so. If they are header only and included in `inst/include` they can be used from other packages by `LinkingTo: rlang`.

lionel-:
- Rename the condition functions to follow our naming scheme. `new_cnd()` is now `cnd()`, `cnd_error()` is now `error_cnd()`, etc. I don't think anyone uses this yet (not even us), but I left the original functions with a warning.

- Allow call depth as `call` argument in signalling function to pluck the call from further up the call stack. Closes #30.

- Move several implementations of condition function to C side, so we get a C API that is consistent with the R functions. I reorganised the code in the process and added some wrappers around R internal functions. Closes #31 and #32.


R/cnd.R
```diff
@@ -34,41 +34,66 @@
...
+    call. = FALSE)
+  error_cnd(.type = .type, ..., .msg = .msg)
 }
-#' @rdname new_cnd
+#' @rdname deprecated-cnd
 #' @export
 cnd_warning <- function(.type = NULL, ..., .msg = NULL) {
-  new_cnd(c(.type, "warning"), ..., .msg = .msg)
+  # Deprecated in 0.1.2
+  warning("`cnd_warning()` has been renamed to `cnd_warning()` for consistency",
```
hadley: to `warning_cnd()` (and similar for message)

_pkgdown.yml
```diff
@@ -398,10 +398,10 @@ reference:
     desc: >
       These helpers make it easy to create typed conditions.
     contents:
-      - new_cnd
-      - cnd_error
-      - cnd_warning
-      - cnd_message
+      - cnd
+      - error_cnd
```
hadley: `ends_with("cnd")` ?

src/cnd.c
```diff
@@ -0,0 +1,139 @@
...
+                                                \
+    BUF[BUFSIZE - 1] = '\0';                    \
+  }
+
+
+void r_inform(const char* fmt, ...) {
+  char buf[BUFSIZE];
+  INTERP(buf, fmt, ...);
+
+  SEXP lang = PROTECT(Rf_lcons(sym("message"), string(buf)));
```
hadley: Shouldn't these set `call. = FALSE`?

src/cnd.h
```diff
@@ -0,0 +1,19 @@
...
+#define RLANG_CND_H
+
+#include <stdbool.h>
+
+
+void r_inform(const char* fmt, ...);
+void r_warn(const char* fmt, ...);
+void r_abort(const char* fmt, ...);
+
+SEXP interp_str(const char* fmt, ...);
```
hadley: Can we have an consistent prefix for exported functions?

lionel-: What do you mean? Functions exported to R side are prefixed with `rlang_` but we do it in the init file.

hadley: Sorry, I was confused - this is still an internal C API, not an exported C API.

lionel-: yes but I'm writing it as if it was an exported API. I guess you're right that public C functions need a consistent prefix. That will make the API much less friendly to use.

hadley: Much less friendly?

lionel-: In the sense that it will clutter the code: `rlang_interp_str()`, `rlang_string()`, `rlang_vec_alloc()` etc, that hinders readability.

lionel-: And takes time to type.

hadley: Why not just make it `r_`? But even `rlang_` is not that bad as you'll get autocomplete.  

Anyway, that's the way C APIs have to be, and there's no point in fighting convention.

lionel-: I wasn't fighting this convention, we can't clutter the global namespace without a consistent organisation of the symbols. I'll review all to-be-exported functions in another PR.

lionel-:
Is this good to merge @hadley?

src/cnd.c
hadley: Apart from this comment, looks ok to me.

lionel-: `message()` doesn't take a `call.` argument but I changed the warning function.

PR #153, opened 2017-05-23
Hook environment for lookups

lionel-:
Includes #152 

- Implements a hook environment as an object table. Requires some trickery to instantiate an environment and set the hashtable to the object table. This environment signals a typed condition when it gets requests for an object.

- Instruments tidyeval overscopes with the hook environment so we get conditions for lookup failures. The hook environment is installed at the end of the lexical scope.

Not as successful as I'd hoped:

- As we discussed grabbing a function is not a distinct operation on object tables, so we can't signal different conditions for symbol lookup and function lookup.

- R often does not call the `exists()` method but uses `get()` and checks for `R_UnboundValue`. In particular it does this when looking up for S3 methods. This means we signal a lot of false positive conditions :(


hadley:
Maybe the hook env will be more effective if it knows its parent so it can proxy the calls and we can attach it higher up - that should avoid the spurious S3 stuff.

PR #159, opened 2017-05-25
Explain the default value of top

egnha:
Otherwise, the significance of `top = NULL` is unclear from the docs alone.


R/eval-tidy.R
```diff
@@ -257,7 +257,8 @@ as_overscope <- function(quo, data = NULL) {
 #'   mechanism that ensures hygienic scoping: the bindings in the
 #'   overscope have precedence, but the bindings in the dynamic
 #'   environment where the tidy quotes were created in the first place
-#'   are in scope as well.
+#'   are in scope as well. The default value of `NULL` sets `top` to the
```
lionel-: Maybe: 

If `NULL` (the default), `bottom` is also the top of the overscope.

egnha: Yes, that is more succinct.

lionel-:
Thanks!

Issue #151, opened 2017-05-23
Native implementations for the `is_()` functions


krlmlr:
See e.g. https://github.com/tidyverse/tibble/issues/262, the lion's share of the run time is spent in `is_vector()` et al..

I don't see a much better solution than translating the boilerplate code that currently exists in R.

lionel-:
yes we should start on that C API at some point

PR #165, opened 2017-06-05
Implement env_set()

### LINKED ISSUE #162: env_set ?
```R
env_set <- function(env, nm, value) {
 
}
```

Obviously you can achieve the same thing with `env_bind()` but it has a nice symmetry with `env_get()` and avoids tidyeval in some common scenarios (i.e. using an env as a cache)

lionel-:
Closes #162 

This probably all should be rewritten in C to avoid performance issues when rlang functions are used in loops (e.g. #151).

* Should we go with a separate `scope_set()` function that would have a different default value for `create`?

* Should `env_bind()` and variants gain a `.create` argument as well?


R/env.R
```diff
@@ -723,14 +723,26 @@ env_has <- function(env = caller_env(), nms, inherit = FALSE) {
...
+#' Get or set an object in an environment
 #'
-#' `env_get()` extracts an object from an enviroment `env`. By
-#' default, it does not look in the parent environments.
+#' @description
+#'
+#' * `env_get()` extracts an object from an enviroment `env`. By
+#'   default, it does not look in the parent environments.
+#'
+#' * `env_set()` will assign or reassign a binding in `env` if
```
hadley: Need to note that this is different to `<<-` which will create in the global env if the binding doesn't already exist.

hadley:
I don't think we need to export `scope_set()` yet. And simiarly, `env_bind()` seems different enough that it doesn't need `.create`

R/env.R
lionel-: `<<-` creates the binding in the current environment rather than in the global environment, so `env_set()` behaves the same when `create` is `TRUE`. Do you mean that by default it doesn't behave like `<<-`?

lionel-: e.g.

```r
```

hadley: > The operators <<- and ->> are normally only used in functions, and cause a search to be made through parent environments for an existing definition of the variable being assigned. If such a variable is found (and its binding is not locked) then its value is redefined, otherwise assignment takes place in the global environment.

lionel-: ooops good to know thanks

lionel-: The default value for `create` is now a function of `inherit`, cf last commit. It is not clear what's the best behaviour is (`env_set()` vs `<<-`), but otoh it is clear that the main purpose of the inheriting case is to override an existing binding so an error is the safest option by default.

hadley: I think binding in `env` is the right choice, but we should note that it's different to `<<-`

PR #166, opened 2017-06-05
Informative error message when coercing a ctrl flow primitive

### LINKED ISSUE #158: fn_fmls gives uninformative error message when called with primitives
```R
fn_fmls(`while`)
```

lionel-:
Closes #158 

Also improve doc about the formals names returned for primitives.


lionel-:
Last commit adds replacement functions for formals

R/fn.R
```diff
@@ -62,6 +62,26 @@ prim_name <- function(prim) {
...
+#' for many primitives the base R argument names are purely
+#' placeholders since they don't perform regular argument matching.
+#' E.g. this returns `5` instead of `-5`:
+#'
+#' ```
+#' `-`(e2 = 10, 5)
+#' ```
+#'
+#' To regularise the semantics of primitive functions, it is usually a
+#' good idea to coerce them to a closure first:
```
hadley: Should we note that `fn_fmls<-` does this automatically?

lionel-: hmm I can't find a meaningful example where this is useful though. That would probably obscure the documentation. Maybe we should issue an error when not a closure.

PR #183, opened 2017-06-19
Typos and the like

jennybc:


vignettes/tidy-evaluation.Rmd
```diff
@@ -115,7 +115,7 @@ positions. Therefore, `select()` understands column symbols like
 dplyr::select(mtcars, cyl)
 
 # Taking an unquoted symbol:
-var <- quote(cyl)
+var <- quo(cyl)
```
jennybc: This is the only non-typo change. I assume this should be `rlang: quo()` but I could be wrong?

lionel-: This doesn't have to be a quosure since I'm referring to overscoped data (in this case a data frame column). I think @hadley only used `quo()` in the dplyr vignette to reduce the number of concepts. I think it's fine to leave it as a bare symbol here though.

hadley: Yeah.  We probably need a whole section that goes more into the details.

jennybc: OK `quo()` changed back to `quote()`.

lionel-:
Thanks Jenny!

PR #186, opened 2017-06-21
Expand C API

lionel-:
This moves quosure self-evaluation and several functions to C.

I'm experimenting with new principles for the UI:

- All allocators like `r_pairlist()` or `r_new_language()` protect their arguments. Underscored versions do not protect. This is intended to make it easier to create data structures, e.g. no need to protect a transient newly allocated string before assigning it in a container. This convention is only for allocators, other functions expect protected arguments.

- PROTECT and UNPROTECT are aliased to the shorter KEEP and FREE to declutter code.


src/env.c
```diff
@@ -1,6 +1,26 @@
 #include "rlang.h"
 
+
+bool r_is_unbound_value(SEXP x) {
+  return x == R_UnboundValue;
+}
+
 SEXP r_mut_env_parent(SEXP env, SEXP new_parent) {
```
hadley: Should mutating functions return void? Should we make them take a `SEXP*` just to be explicit?

src/env.c
```diff
@@ -1,6 +1,26 @@
...
+
+bool r_is_env(SEXP x) {
+  return r_typeof(x) == ENVSXP;
+}
+
+SEXP r_env_get(SEXP env, SEXP sym) {
+  return Rf_findVarInFrame3(env, sym, TRUE);
+}
+
+SEXP r_ns_env(const char* pkg) {
```
hadley: `r_ns_find()`?

src/sexp.c
```diff
@@ -50,3 +62,27 @@ bool is_named(SEXP x) {
...
+}
+bool r_is_null(SEXP x) {
+  return x == R_NilValue;
+}
+
+
+SEXP r_duplicate(SEXP x) {
+  return Rf_duplicate(x);
+}
+SEXP r_shallow_duplicate(SEXP x) {
```
hadley: Should this be an argument to `r_duplicate()`? That would force you to consider whether it should be shallow or deep whenever you duplicate.

hadley:
Maybe we should use `r_new_` as the common prefix for anything that creates a new SEXP.

src/env.c
lionel-: I like the `void` return value, but taking a `SEXP*` would imply that the pointer holding the SEXPREC would change, which won't be the case.

I'd like to typedef the `SEXPREC` to `sexp`, so we'd use `sexp*` instead of `SEXP` to make semantics more explicit.

src/env.c
lionel-: Shouldn't this be `_env` since it returns an environment?

lionel-:
> Maybe we should use r_new_ as the common prefix for anything that creates a new SEXP.

Good idea, that will be more consistent with our convention that unprefixed ctors are variadic. Also it makes it easier to document the protecting semantics: functions prefixed with `new_` accept unprotected arguments.

lionel-:
> Should we have a convention for functions whose return values should be protected?

that would be nice, the `new_` prefix is probably not enough

src/env.c
hadley: `r_ns_find_env()`?

src/env.c
hadley: Could we use a typedef of `SEXP*` to indicate when an argument is modified?  It wouldn't improve the type safety, but it might improve documentation slightly

src/env.c
lionel-: could it just be `r_ns_env()` like the R functions `rlang::ns_env()`, `rlang::pkg_env()`, etc?

hadley: I guess, but the lack of a verb in those function names leaves me guessing as to what they do.

src/env.c
lionel-: Maybe `#define mut`?

```c
r_mut_env_parent(mut sexp* env, sexp* new_parent);
```

That doesn't look bad but I'm not sure I like it because it seems to imply that `sexp` are otherwise const.

lionel-: > because it seems to imply that sexp are otherwise const

on hindsight maybe not: we can just see it as a "restating the obvious" keyword. It doesn't change the status of the argument since it's mutable by default, but stresses that this is a case where we do want it mutable, which is what we want for documentation purposes.

hadley: Alternatively, maybe we should just using a naming convention for the argument.

Calling them input/output parameters might be more clear than mutable? https://en.wikipedia.org/wiki/Parameter_(computer_programming)#Output_parameters

lionel-: e.g.

```c
r_mut_env_parent(mut sexp* env, sexp* new_parent);
```

versus

```c
r_mut_env_parent(sexp* output_env, sexp* new_parent);
```

The latter does seem clearer upon reading, but maybe easier to overlook.

PR #188, opened 2017-06-22
Parse unicode points when call-splicing a list

lionel-:
Part of tidyverse/dplyr#2886


R/vector-chr.R
```diff
@@ -82,7 +99,7 @@ string <- function(x, encoding = NULL) {
 #' as_bytes(latin1)
 #' }
 as_utf8_character <- function(x) {
-  enc2utf8(as_character(x))
+  .Call(rlang_unescape_character, as_character(x))
```
krlmlr: This looks like a very far-reaching design choice. Maybe we should provide a separate operation for unescaping?

R/dots.R
```diff
@@ -189,6 +189,15 @@ dot_interp <- function(dot, quosured = TRUE) {
     dots <- call("alist", expr)
     dots <- .Call(rlang_interp, dots, env, quosured)
     dots <- eval_bare(dots)
+
+    # Handle serialised unicode for characters that are not
+    # representable in the current locale. They have been serialised
+    # when converted to symbols (during pairlist-splicing).
+    nms <- names(dots)
```
krlmlr: Maybe even add an entry point `unescaped_names()`?

R/vector-chr.R
lionel-: Why is it far-reaching? This is our take on converting to UTF-8.

krlmlr: Sure it's a way of dealing with the problem. My point is that it may have implications beyond fixing this particular problem.

lionel-: The base function `enc2utf8()` is there if you don't want to parse unicode points. What implications have you in mind?

krlmlr: Calling `as_utf8_character()` without having read its documentation and understood what it does. This can be us, a package that uses rlang, or user code, perhaps already implemented against rlang 0.1.1.

lionel-: I think it's a good thing to be consistent about translation to UTF-8 throughout rlang.

krlmlr: I don't mind being consistent, I'd just love to see a new entry point for new functionality.

lionel-: I have now added the new exported function `chr_translate_unicode()`.

PR #192, opened 2017-06-25
Eliminate spurious enumeration in stack doc

egnha:
Markdown gotcha: a ” 1.” ending of a sentence happened to start a `#'`-line.


egnha:
BTW, I now notice there are other inaccuracies in the stack docs (refactoring casualties?), particularly in “Details” (the call depths off by 1, components of return value of `call_frame()`, `ctxt_frame()` not exactly as described) and “Examples” (same problems).

lionel-:
Thanks!

> BTW, I now notice there are other inaccuracies in the stack docs (refactoring casualties?), particularly in “Details”

oh right, this should be solved by the first commits in #189

PR #197, opened 2017-06-28
Make stack UI more consistent

lionel-:
Extracted from #189 

- `n = 0` now refers to the current frame rather than `n = 1`.
- `call_depth()` and `ctxt_depth()` called at top-level now return 0.


hadley:
This is a major API change so we need to think about it a bit.

lionel-:
Let's close this as the stack functions will be marked as experimental and probably belong to another package.

PR #210, opened 2017-07-13
Utils for serialisation

lionel-:


R/expr-lang.R
```diff
@@ -251,6 +251,17 @@ is_binary_lang <- function(x, name = NULL, ns = NULL) {
   is_lang(x, name, n = 2L, ns = ns)
 }
 
+#' @rdname is_lang
```
hadley: Let's leave these as is, and consider a fix in a separate PR.

PR #230, opened 2017-08-08
Signal depreciated functions

lionel-:
Also adds `cnd_inform()` and `cnd_warn()` as complements to `cnd_signal()` and `cnd_abort()`.


NEWS.md
```diff
@@ -42,6 +42,14 @@
   (#174). Previously it would always refer to the environment of the
   outermost quosure.
 
+* The new functions `cnd_warn()` and `cnd_inform()` transform
+  conditions to warnings or messages before signalling them.
+
+* `signal_deprecated()` and `warn_obsolete()` depreciate functions. A
```
hadley: I'm slight concerned that we're creating a parallel set of terminology to deprecated and defunct in base R.

R/cnd-utils.R
```diff
@@ -0,0 +1,71 @@
+#' Signal a deprecated or obsolete function
+#'
+#' @description
+#'
+#' These functions signal a function as deprecated or obsolete.
```
hadley: I'd rather use the same verb here, even though the actions are slightly different.

I think we should also use a suffix `signal_dep_fun()` so that we can naturally extend to `signal_dep_arg()`.

I'd also like to consider capturing the date or package version somehow, so that we can have a standard soft dep -> dep -> defunct -> deleted process.

lionel-:
This is now a set of changes needed for the `oldie` package.

R/state.R
```diff
@@ -0,0 +1,99 @@
...
+  eval_bare(expr, frame)
+
+  invisible(expr)
+}
+
+#' Change global options
+#'
+#' @description
+#'
+#' * `scoped_options()` changes options for the duration of a stack
```
hadley: Seeing these all together makes me think that the "option(s)" should be a prefix, not a suffix?

lionel-: I think the prefix should be `scoped_` because we've already started using it, e.g. `scoped_vars()`. Also this makes these families more consistent: `with_`, `scoped_`, `poke_`, `peek_`.

PR #238, opened 2017-08-20
Ensure that is_integerish() heeds length requirement

egnha:
Previously, `is_integerish(x, n)` would disregard `n` for `x` an integer vector.


egnha:
The tests are rather pedantic. Feel free to cut the fat. ;)

lionel-:
Thanks!

egnha:
My pleasure!

PR #251, opened 2017-09-04
typo: "withdplyr"

infotroph:


lionel-:
Could you modify the `.R` file as well please? And redocument afterwards so that the `.Rd` file is updated too.

infotroph:
Sorry for editing the wrong file -- I didn't see the string "withdplyr" anywhere else in the repository, and missed that this was a generated page. 

But I haven't used pkgdown and don't want to learn for a one-character patch, so I'll close this and leave my report for another contributor to fix.

PR #254, opened 2017-09-06
Add options API

lionel-:
I think `options` shouldn't be a prefix here because we already have other `scoped_`, `with_`, `peek_` and `poke_` functions.


hadley:
Although the three current `scoped_` functions are all to do with the scoped environments, not limited scope effects.

I kind of like the idea of `scoped_options()` and `with_options()`, but `options_peek()` and `options_poke()` — for scoped and with, the emphasise is on changing some global option in a contained way, but peeking and poking is more about the options.

lionel-:
I don't know. `options_peek()` would also imply that the input is an option, but that's not the case here. It might work for `options_poke()`.

All considered I think these function names are more predictable if we consistently keep `peek_`, `scoped_` etc as prefix.

Do you think `peek_vars()` should be `vars_peek()` as well?

hadley:
So far we've used the prefix to refer to the type, but I think it's generally makes more sense for the "topic" (e.g. options here) to be the prefix, because it allows you to think about the topic and then use autocomplete to figure out what you can do with it.

But maybe this is suggesting that these option handling functions don't belong in rlang? Maybe they should live in withr instead? (Since there's a whole bunch of other side-effects in that package)

lionel-:
> But maybe this is suggesting that these option handling functions don't belong in rlang? Maybe they should live in withr instead?

Maybe though I'm not sure the `peek_` and `poke_` would fit in the package.

But it doesn't solve the naming issue which will come up again for exits and events. I've thought about it for tidyselect and found at the time that using `vars` as grammatical objects was best. E.g. this would look weird:

```r
vars_scoped()
vars_with()`
```

Or do you think we should keep `scoped_vars()` etc the way they are but that `poke_vars()` and `peek_vars()` should be renamed like this?

```r
vars_peek()
vars_poke()
```

I agree that topical prefixes make sense, though I was thinking of poking and peeking as being the topics. Just like we have the topic of predicates which all start with `is_`.

lionel-:
Other examples: `set_` and `mut_` functions (maybe the latter should be renamed to `poke_`)

PR #260, opened 2017-09-08
Add is_pairlist_node()

lionel-:
Change predicates so that:

- "node" refers to both language and pairlist types. These types support the same kind of operation.
- "pairlist node" refers to pairlist type, i.e. a cons cell.

This definition hasn't changed:

- "pairlist" refers to both pairlist and NULL types. This is consistent with lisp where `nil` is the empty list.


R/expr-node.R
```diff
@@ -203,10 +203,14 @@ as_pairlist <- function(x) {
 #'   node of pairlists and an empty pairlist is thus the `NULL`
 #'   object itself.
 #'
-#' * `is_node()` checks that `x` has type `pairlist`.
+#' * `is_pairlist_node()` checks that `x` has type `pairlist`.
```
hadley: Is that better than just `is_pairlist()`?

lionel-: `is_pairlist()` is taken since rlang 0.1 for the the pairlist data structure: either a pairlist node or `NULL`. Maybe it should be the other way around? `is_pairlist()` to check for pairlist base type? But then I don't know how to call the predicate that checks for either pairlist or `NULL`.

I don't remember having used `is_pairlist()` for checking NULLs as well, so we could just remove that meaning for now until the need for it comes up.

hadley: I'm confused. What's the difference between `is_pairlist()` and `is_pairlist_node()`, just whether or not `NULL` is included?

lionel-: yes `is_pairlist()` also returns true for NULL while `is_pairlist_node()` only returns true for cons cells.

lionel-: How about:

- `is_pairlist()` for pairlist type only
- `is_node_list()` for pairlist + NULL types
- `is_node()` for pairlist + language types

hadley: Sounds good

PR #261, opened 2017-09-08
Scoped bindings

lionel-:
Useful to overscope symbols in a given environment instead of a child. I.e. when evaluating a stack-sensitive expression.

Also adds `env_get_list()` to retrieve multiple objects in a list.


NAMESPACE
```diff
@@ -383,6 +384,7 @@ export(syms)
 export(type_of)
 export(warn)
 export(warning_cnd)
+export(with_bindings)
```
hadley: I think you forgot to export `scoped_bindings`

R/env.R
```diff
@@ -637,6 +637,61 @@ env_bind_fns <- function(.env, ...) {
...
+#' @return `scoped_bindings()` returns the values of old bindings
+#'   invisibly; `with_bindings()` returns the value of `expr`.
+#' @examples
+#' foo <- "foo"
+#' bar <- "bar"
+#'
+#' # `foo` will be temporarily rebinded while executing `expr`
+#' with_bindings(paste(foo, bar), foo = "rebinded")
+#' paste(foo, bar)
+scoped_bindings <- function(..., .env = .frame, .frame = caller_env()) {
```
hadley: Where does `.frame` come from?

R/env.R
```diff
@@ -778,6 +835,12 @@ env_has <- function(env = caller_env(), nms, inherit = FALSE) {
 env_get <- function(env = caller_env(), nm, inherit = FALSE) {
   get(nm, envir = get_env(env), inherits = inherit)
 }
+#' @rdname env_get
+#' @export
+env_get_list <- function(env = caller_env(), nms, inherit = FALSE) {
```
hadley: `env_get_many()`?

R/env.R
lionel-: It's the third argument with default `caller_env()`.

R/env.R
lionel-: I thought `_list` would be consistent with the return type rule. We already have other `_list` like `dots_list()`.

R/env.R
hadley: Doh

R/env.R
hadley: Ok

PR #263, opened 2017-09-11
Is there still a version prefixed or sufixed with `f_` or `_f`?

maurolepore:
It seems that the valid suffix is `_quosure` or `_quosures`.

### DIFF
--- a/R/parse.R
+++ b/R/parse.R
@@ -11,7 +11,7 @@
 #' [base::parse()] which returns an base::expression vector). All
 #' functions also support R connections.
 #'
-#' The versions prefixed with `f_`?? return expressions quoted in
+#' The versions suffixed with `_quosure` return quosures quoted in
 #' formulas rather than raw expressions.
 #'
 #' @param x Text containing expressions to parse_expr for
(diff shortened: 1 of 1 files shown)


R/parse.R
```diff
@@ -11,7 +11,7 @@
 #' [base::parse()] which returns an base::expression vector). All
 #' functions also support R connections.
 #'
-#' The versions prefixed with `f_` return expressions quoted in
+#' The versions prefixed with `f_`?? return expressions quoted in
```
lionel-: Could you change this to ``suffixed with `_quosure` `` return quosures.

R/parse.R
```diff
@@ -32,7 +32,7 @@
 #' # A string can contain several expressions separated by ; or \n
 #' parse_exprs("NULL; list()\n foo(bar)")
 #'
-#' # The versions suffixed with _f return formulas:
+#' # The versions suffixed with _quosure/s return formulas:
```
lionel-: return quosures

lionel-:
Thanks!

PR #265, opened 2017-09-14
Improve dots_definitions() API

lionel-:
Currently `dots_definitions()` supports both regular arguments and `:=` definitions. The return value is thus a list of regular inputs and `:=` inputs. This complicates things and doesn't seem very useful since an UI mixing both types is unlikely. With this patch `dots_definitions()` only supports `:=` inputs and has a simpler return value.

In addition I added:

- `dots_formulas()` to capture two-sided formulas in quosures. This is just like `dots_definitions()` but for `~`.

- `dots_patterns()` to capture either two-sided formulas or definitions. This lets the user choose her preferred style. Is this a good idea? Personally I prefer supplying `:=` because I find it much easier to read in functions like `case_when()`.


egnha:
@lionel- , if you were to keep `dots_definitions()`, would it cause you a lot of trouble?

It plays a crucial role in my package [valaddin](https://github.com/egnha/valaddin). Without it, I would have to implement a substantially more complicated API or overhaul the API completely (again).

However, if `dots_definitions()` goes away, do you think it is possible to recreate its functionality without accessing rlang's private API?

lionel-:
Maybe we could have something like `dots_extract(..., two_sided = c(":=", "~"), one_sided = TRUE)`

egnha:
@lionel- Thanks for taking this into consideration. (And sorry for the slow response—been rather occupied lately.)

Minor comment: The arguments `one_sided`, `two_sided` have similar names but rather different roles and types. To me it seems that `one_sided = TRUE` and `one_sided = FALSE` might be better off as separate functions (the former being basically the original `dots_definitions()` but with an option to use `:=` or `~`).

Your thoughts?

lionel-:
> The arguments one_sided, two_sided have similar names but rather different roles and types. To me it seems that one_sided = TRUE and one_sided = FALSE might be better off as separate functions 

Hmm they have different input types but it seems like they are doing the same thing: selectively allowing extraction of dots components.

egnha:
OK, then do you mean that `dots_extract(..., one_sided = TRUE)` would function as the original `dots_definitions()` did, and that `dots_extract(..., one_sided = FALSE)` would be a stricter version of that that only accepts definitions (according to the syntax specified by the `two_sided` option)? (I.e., `one_sided = TRUE` means “_additionally_, allow one-sided inputs.”)

lionel-:
Yup that's what I meant. The point is for `dots_extract()` to return a list of lists (with configurable components) and for the other variants to return a simple list.

egnha:
A thought about the names: since `dots_extract(..., one_sided = FALSE)` is doing what I suppose `dots_definitions()` should have been doing, how about calling that `dots_definitions()` (your original proposal), and restricting `dots_extract()` to the case `dots_extract(..., one_sided = TRUE)`?

In other words, two functions, with signatures like
- `dots_definitions(..., definer = c(":=", "~"))` — strict version of the original `dots_definitions()`
- `dots_extract(...)` — original `dots_definitions()`

Might this be more intuitive?

lionel-:
@egnha I have now restored the original definition of `dots_definitions()` but I have marked it as experimental. We are trying out a new convention to make it clear in the documentation what API stability to expect for each function.

@hadley This PR now only adds a `dots_patterns()` function. By default it takes formulas by expression and extracts each components into quosures. It optionally take an operator name so you can use it with e.g. `:=`. Like `dots_definitions()`, this function is marked as experimental.

`dots_patterns()` is needed for redpen so it would be helpful to merge this branch.

lionel-:
I'll just embed `dots_patterns()` in redpen for now.

PR #270, opened 2017-09-22
Changes for redpen

lionel-:
Includes #265.


lionel-:
@hadley I have embedded all the new functions in the redpen package. This PR is now just about adding an `env` argument to `call_modify()` and `call_standardise()` in case the input is not a quosure, so we can find the function definition in the right place.

PR #273, opened 2017-09-30
Better handling of non-finite values in numeric predicates

lionel-:
- `is_integerish(c(Inf, -Inf, NaN))` now returns `TRUE`

- There is a new `finite` argument for `is_integerish()` and `is_double()` to check whether all values are finite.

### DIFF
--- a/R/types.R
+++ b/R/types.R
@@ -18,6 +18,8 @@
 #'
 #' @param x Object to be tested.
 #' @param n Expected length of a vector.
+#' @param finite Whether values must be finite. Examples of non-finite
+#'   values are `Inf`, `-Inf` and `NaN`.
 #' @param encoding Expected encoding of a string or character
 #'   vector. One of `UTF-8`, `latin1`, or `unknown`.
 #' @seealso [bare-type-predicates] [scalar-type-predicates]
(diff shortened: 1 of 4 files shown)


hadley:
Isn't the definition of integerish something like `abs(x - as.integer(x)) < tol)`? So I don't think Inf and -Inf qualify.

lionel-:
It's an equality check with `==` at C level. We currently don't check for finite values which causes `is_integerish()` to return `NA` with a coercion warning in corner cases. That prevents it from being used as is in `if ()` tests. Relatedly `is_integerish(na_dbl)` and `is_integerish(na_int)` return NA which also causes issues.

We could return `FALSE` when we detect a non-finite value. However I think it makes sense to accept them by default because `is_integerish()` is meant to be used instead of `is.numeric()` while avoiding S3 issues. The `finite` parameter allows more fine-grained tests.

hadley:
Hmmm, in that case I think the default should be to return `FALSE` for Inf and -Inf.

lionel-:
Would it make sense to have `is_finite()` and `are_finite()` predicates? The former would ignore missing values (so that `is_finite(c(1, NA))` is `TRUE`) while the vectorised predicate would return `NA` for missing values. By comparison `is.finite()` returns `FALSE` for missing values.

PR #285, opened 2017-10-27
Vignette links

ijlyttle:
I'm not entirely sure on this PR - some parts you might like, others you might not.

Each of the changes here concerns a link to the package's documentation. It seems that some of the links were using the newer roxygen-markdown syntax that I don't think works in vignettes (or at least is not working in the latest publication of the rlang pkgdown-site).

In some cases the link could be redone without changing the appearance. In a couple of cases, I proposes explicit links to a package functions, which makes things a little "clunkier". In one case, I propose an internal reference, rather than an external reference (rlang.tidyverse.org, which _could_ change).

I am happy to revert any of these changes, as you may like.


ijlyttle:
Understood - no worries.

PR #289, opened 2017-11-13
Fix bang bang operator precedence

lionel-:
This fixes the parse tree so that `!!` binds tightly:

```r
foo <- "foo"

expr(!! foo * bar == baz)
#> "foo" * bar == baz

expr(!! foo %>% bar())
#> "foo" %>% bar()
```


src/rlang/expr-interp.c
```diff
@@ -16,6 +16,17 @@ static const char* uq_names[UQ_N] = { "UQ", "UQE", "!!" };
 static const char* uqs_names[UQS_N] = { "UQS", "!!!"};
 
 
+#define FIXUP_OPS_N 9
+#define FIXUP_UNARY_OPS_N 2
+
+static const char* fixup_ops_names[FIXUP_OPS_N] = {
+  "<", ">", "<=", ">=", "==", "!=", "*", "/", ":"
```
hadley: What about `^`?

`$`, `@`, `::` and `:::` are ok, I think, because you can't put `!!` on the LHS.

lionel-: oh yes, `^` needs to be handled separately because it's right-associative, I forgot about it.

lionel-: Actually right-associative ops are a special case of the logic used for left-associative operators.

Issue #295, opened 2017-11-21
Option to accept evaluated arguments


lionel-:
When switched on, this:

```r
f <- function(x) {
  force(x)
  enexpr(x)
}
f(foo)
```

Would be equivalent to this:

```r
f <- function(x) {
  enexpr(x)
}
f(!! foo)
```

lionel-:
Useful for this sort of situation:

```r
lapply(1:2, tibble::tibble)
```

Issue #290, opened 2017-11-13
Fix performance of dots capture


lionel-:
cf https://github.com/tidyverse/purrr/issues/388

lionel-:
Now fixed on master.

Issue #280, opened 2017-10-17
Unquote-splicing a pairlist may mutate it


egnha:
Under 0.1.2.9000 and R 3.4.2, unquote-splicing a pairlist in `quo()` can mutate the pairlist itself.

Example where there should be no side-effects:
```R
x <- pairlist(0)

quo(foo(UQS(x), y))
#> <quosure: global>
#> ~foo(0, y)

x
#> [[1]]
#> [1] 0
#> 
#> [[2]]
#> y
```
The same occurs for `expr()`. `x` is not mutated if it is a list.

(Is the culprit the C function `rlang_interp()`?)

lionel-:
Hmm... This can probably be fixed by adding a `duplicate()` within the R function `UQS()`.

egnha:
Thanks for the workaround.

Nevertheless, do you think this ought to be regarded a bug?

lionel-:
Yes this is a bug, there should be no side effects.

Issue #267, opened 2017-09-19
Feature request: add .named arg to exprs()


Hong-Revo:
So that

```r
exprs(..., .named=TRUE)
```
would be equivalent to

```r
lapply(quos(..., .named=TRUE), get_expr)
```

PR #297, opened 2017-11-24
Implement dots capture in C

### LINKED ISSUE #201: enquo() is inconsistent with default value for an argument
enquo() shows an inconsistent behaviour when the default value for an argument is the same as the name of an argument.

``` r
library(rlang)
f <- function(x = x) {
  enquo(x)
}

f2 <- function(x = x1) {
  enquo(x)
}

# both work fine
f(x1)
#> <quosure: frame>
#> ~x1
f2(x1)
#> <quosure: frame>
#> ~x1

# unexpected behaviour for f
# I'd rather expect this to be the same as quo(x)
f()
#> <quosure: empty>
#> ~
f2()
#> <quosure: local>
#> ~x1
```

<details><summary>Session info</summary>

``` r
devtools::session_info()
#> Session info -------------------------------------------------------------
#>  setting  value                       
#>  version  R version 3.4.0 (2017-04-21)
#>  system   x86_64, darwin15.6.0        
#>  ui       X11                         
#>  language (EN)                        
#>  collate  en_US.UTF-8                 
#>  tz       Europe/Berlin               
#>  date     2017-06-29
#> Packages -----------------------------------------------------------------
#>  package   * version    date       source                          
#>  backports   1.0.5      2017-01-18 CRAN (R 3.4.0)                  
#>  base      * 3.4.0      2017-04-21 local                           
#>  compiler    3.4.0      2017-04-21 local                           
#>  datasets  * 3.4.0      2017-04-21 local                           
#>  devtools    1.13.1     2017-05-13 CRAN (R 3.4.0)                  
#>  digest      0.6.12     2017-01-27 CRAN (R 3.4.0)                  
#>  evaluate    0.10       2016-10-11 CRAN (R 3.4.0)                  
#>  graphics  * 3.4.0      2017-04-21 local                           
#>  grDevices * 3.4.0      2017-04-21 local                           
#>  htmltools   0.3.6      2017-04-28 CRAN (R 3.4.0)                  
#>  knitr       1.16       2017-05-18 CRAN (R 3.4.0)                  
#>  magrittr    1.5        2014-11-22 CRAN (R 3.4.0)                  
#>  memoise     1.1.0      2017-04-21 CRAN (R 3.4.0)                  
#>  methods   * 3.4.0      2017-04-21 local                           
#>  Rcpp        0.12.11    2017-05-22 CRAN (R 3.4.0)                  
#>  rlang     * 0.1.1.9000 2017-06-29 Github (tidyverse/rlang@c4887ed)
#>  rmarkdown   1.5        2017-04-26 CRAN (R 3.4.0)                  
#>  rprojroot   1.2        2017-01-16 CRAN (R 3.4.0)                  
#>  stats     * 3.4.0      2017-04-21 local                           
#>  stringi     1.1.5      2017-04-07 CRAN (R 3.4.0)                  
#>  stringr     1.2.0      2017-02-18 CRAN (R 3.4.0)                  
#>  tools       3.4.0      2017-04-21 local                           
#>  utils     * 3.4.0      2017-04-21 local                           
#>  withr       1.0.2      2016-06-20 CRAN (R 3.4.0)                  
#>  yaml        2.1.14     2016-11-12 CRAN (R 3.4.0)
```

</details>

lionel-:
Closes #201, #217, #242, #267, #280, #290, #295

The bulk of the refactor is in https://github.com/tidyverse/rlang/pull/297/files#diff-3b4f8126fe47843caf246e738300bee2:

`dots_unquote()` walks the captured dots, evaluates the unquoted expressions and records the size of spliced values. It ignores empty argument depending on the value of `ignore_empty`. A switchpatch ensures the dots values are correctly wrapped depending on the target type (raw expressions, quosures, values). A second function `dots_interp()` is in charge of building the final list.

The new implementation puts dots capture with few arguments on the order of 3 to 10 µs, a bit more if splicing is involved (mostly because a larger final vector is allocated and filled). With the previous implementation it was often on the order of milliseconds.

On my machine the following expression takes 50 ms with CRAN rlang but 45 µs with this PR:

```r
quos(
  foo, bar(), foo,
  !!! letters,
  !! letters[[4]] := foo,
  !!! letters,
  !! letters[[4]] := foo,
  !!! letters,
  !! letters[[4]] := foo
)
```

The dots capture functions have a more consistent interface (e.g. `exprs()` accepts `.named` argument). Also they now all accept forced arguments, which are treated as if they were unquoted.

Unquoting has been improved which also benefits `enquo()` and `enexpr()`. Unquoting is now faster with `expr()` than with `bquote()`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -75,6 +75,11 @@
   off interpretation of `:=` as a name operator. This should be useful
   for programming on the language targetting APIs such as `data.table`.
 
+* The backend for `quos()`, `exprs()`, `dots_list()`, etc is now
+  written in C. This greatly improve the performance of dots capture.
+  The unquoting algorithm has been improved and `enexpr()` and
+  `enquo()` are also more efficient as a result.
+
 
 ## Breaking changes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -65,20 +65,22 @@
 * The tidy eval `!!` operator now binds tightly. You no longer have to
   wrap it in parentheses, i.e. `!! x > y` will only unquote `x`.
 
-* `!!!` now accepts any kind of objects. Scalar types are treated as
-  vectors of length 1. Previously only symbolic objects like symbols
-  and calls were treated as such in order to allow splicing of
-  function bodies (which are not necessarily wrapped in a `{` block).
-  Allowing any kind of objects is more consistent.
+* `!!!` now accepts any kind of objects for consistency. Scalar types
+  are treated as vectors of length 1. Previously only symbolic objects
+  like symbols and calls were treated as such in order to allow
+  splicing of function bodies (which are not necessarily wrapped in a
+  `{` block).
 
 * `exprs()` and `quos()` gain a `.unquote_names` arguments to switch
   off interpretation of `:=` as a name operator. This should be useful
-  for programming on the language targetting APIs such as `data.table`.
+  for programming on the language targetting APIs such as data.table.
 
 * The backend for `quos()`, `exprs()`, `dots_list()`, etc is now
-  written in C. This greatly improve the performance of dots capture.
-  The unquoting algorithm has been improved and `enexpr()` and
-  `enquo()` are also more efficient as a result.
+  written in C. This greatly improve the performance of dots capture,
+  especially with the splicing operator `!!!` which now scales much
+  better (you'll see a 1000x performance gain in some cases). The
+  unquoting algorithm has also been improved which makes `enexpr()`
+  and `enquo()` more efficient as well.
 
 * `enquo()` and `enexpr()` now deal with default values correctly (#201).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -80,6 +80,29 @@
   The unquoting algorithm has been improved and `enexpr()` and
   `enquo()` are also more efficient as a result.
 
+* `enquo()` and `enexpr()` now deal with default values correctly (#201).
+
+* Functions taking dots by value rather than by expression
+  (e.g. regular functions, not quoting functions) have a more
+  restricted set of unquoting operations. They only support `:=` and
+  `!!!`, and only at top-level. I.e. `dots_list(!!! x)` is valid but
+  not `dots_list(deep(!!! x))` (#217).
+
+* Functions taking dots by value now support splicing a `NULL`
+  value. `dots_list(!!! NULL)` is equivalent to `dots_list()` (#242).
+
+* `exprs()` gains a `.named` option to auto-label its arguments (#267).
+
+* Splicing a list no longer mutates it (#280).
+
+* Capture operators now support evaluated arguments. Capturing a
+  forced or evaluated argument is exactly the same as unquoting that
+  argument: the actual object (even if a vector) is inlined in the
+  expression. Capturing a forced argument occurs when you use
+  `enquo()`, `enexpr()`, etc too late. It also happens when your
+  quoting function is supplied to `lapply()` or when you try to quote
+  the first argument of an S3 method (which is necessarily evaluated
+  in order to detect which class to dispatch to). (#295, #300).
 
 ## Breaking changes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -65,6 +65,9 @@
 * The tidy eval `!!` operator now binds tightly. You no longer have to
   wrap it in parentheses, i.e. `!! x > y` will only unquote `x`.
 
+* `!!!` now fails with symbols and calls. Only vectors, node lists,
+  `NULL` and `{` blocks can be spliced.
+
 
 ## Breaking changes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -68,6 +68,10 @@
 * `!!!` now fails with symbols and calls. Only vectors, node lists,
   `NULL` and `{` blocks can be spliced.
 
+* `exprs()` and `quos()` gain a `.unquote_names` arguments to switch
+  off interpretation of `:=` as a name operator. This should be useful
+  for programming on the language targetting APIs such as `data.table`.
+
 
 ## Breaking changes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -65,8 +65,11 @@
 * The tidy eval `!!` operator now binds tightly. You no longer have to
   wrap it in parentheses, i.e. `!! x > y` will only unquote `x`.
 
-* `!!!` now fails with symbols and calls. Only vectors, node lists,
-  `NULL` and `{` blocks can be spliced.
+* `!!!` now accepts any kind of objects. Scalar types are treated as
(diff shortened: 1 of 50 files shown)


src/dots.c
```diff
@@ -0,0 +1,502 @@
...
+  return dots;
+}
+
+
+static int match_ignore_empty_arg(sexp* ignore_empty) {
+  if (!r_is_character(ignore_empty) || r_length(ignore_empty) == 0) {
+    r_abort("`.ignore_empty` must be a character vector");
+  }
+  const char* arg = r_c_string(ignore_empty);
+  switch(arg[0]) {
```
hadley: This seems excessively efficient 😉 

(this is ok if `.ignore_empty = ""` because you can switch on null?)

src/dots.c
```diff
@@ -0,0 +1,502 @@
...
+ error:
+  r_abort("`.named` must be a scalar logical or number");
+}
+
+static sexp* maybe_auto_name(sexp* x, sexp* named) {
+  int names_width = find_auto_names_width(named);
+  sexp* names = r_names(x);
+
+  if (names_width && (!names || r_chr_has(names, ""))) {
+    sexp* auto_fn = rlang_ns_get("quos_auto_name");
```
hadley: I wonder if we can/should wrap this up into a function?

```c
call_r_fun(r_empty_env, "rlang", "quos_auto_name", x, width)
```

src/dots.c
```diff
@@ -0,0 +1,502 @@
...
+
+static sexp* init_names(sexp* x) {
+  sexp* nms = KEEP(r_new_vector(r_type_character, r_length(x)));
+  r_push_names(x, nms);
+  FREE(1);
+  return nms;
+}
+
+
+sexp* capturedots(sexp* frame);
```
hadley: Should a comment say why you need this forward declaration?

src/dots.c
```diff
@@ -0,0 +1,502 @@
...
+
+  sexp* out = KEEP(r_new_vector(r_type_list, capture_info->count));
+
+  // Add default empty names unless dots are captured by values
+  sexp* out_names = r_null;
+  if (capture_info->type != DOTS_VALUE || dots_names != r_null) {
+    out_names = init_names(out);
+  }
+
+  for (size_t i = 0, count = 0; i < r_length(dots); ++i) {
```
hadley: Is dots a pairlist? If so, calling `r_length()` repeatedly might be expensive.

NEWS.md
```diff
@@ -65,6 +65,44 @@
 * The tidy eval `!!` operator now binds tightly. You no longer have to
   wrap it in parentheses, i.e. `!! x > y` will only unquote `x`.
 
+* `!!!` now accepts any kind of objects. Scalar types are treated as
+  vectors of length 1. Previously only symbolic objects like symbols
+  and calls were treated as such in order to allow splicing of
+  function bodies (which are not necessarily wrapped in a `{` block).
+  Allowing any kind of objects is more consistent.
```
hadley: "kind of objects" -> "kind of object" x2

NEWS.md
```diff
@@ -65,6 +65,44 @@
...
+
+* `exprs()` and `quos()` gain a `.unquote_names` arguments to switch
+  off interpretation of `:=` as a name operator. This should be useful
+  for programming on the language targetting APIs such as `data.table`.
```
hadley: Convention is to not quote package names

NEWS.md
```diff
@@ -65,6 +65,44 @@
...
+
+* The backend for `quos()`, `exprs()`, `dots_list()`, etc is now
+  written in C. This greatly improve the performance of dots capture.
+  The unquoting algorithm has been improved and `enexpr()` and
+  `enquo()` are also more efficient as a result.
```
hadley: I think you should briefly mention how much of a speed up this is - e.g. ~1000x

R/expr-interp.R
```diff
@@ -95,34 +95,20 @@ NULL
 #' @export
 #' @rdname quasiquotation
 UQ <- function(x) {
-  x
+  abort("`UQ()` can't be called directly")
```
hadley: Add something like "Check that you are in a quasiquoted argument" ?

R/expr-sym.R
```diff
@@ -37,8 +29,8 @@ sym <- function(x) {
 }
 #' @rdname sym
 #' @export
-syms <- function(...) {
-  map(chr(...), sym)
+syms <- function(x) {
```
hadley: This is another API breaking change that you should explore the consequences of

R/quos.R
```diff
@@ -26,6 +26,8 @@
 #'   name. If an integer, it is passed to the `width` argument of
 #'   `expr_text()`, if `TRUE`, the default width is used. See
 #'   [exprs_auto_name()].
+#' @param .unquote_names Whether to treat `:=` as `=`. Unlike `=`, the
+#'   `:=` syntax supports unquoting with `!!`.
```
hadley: "on the LHS"

src/capture.c
```diff
@@ -4,11 +4,12 @@
 #define _(string) (string)
 
 
-SEXP attribute_hidden capture_arg(SEXP x, SEXP env) {
+SEXP attribute_hidden new_captured_arg(SEXP x, SEXP env) {
     static SEXP nms = NULL;
     if (!nms) {
         nms = allocVector(STRSXP, 2);
         R_PreserveObject(nms);
```
hadley: `r_mark_precious` ?

tests/testthat/test-dots.R
```diff
@@ -1,11 +1,12 @@
...
-  fn <- function(f, ...) f(...)
-  expect_identical(fn(exprs), named_list())
+test_that("exprs() without arguments creates an empty named list", {
+  expect_identical(exprs(), named_list())
+})
 
-  g <- function(f, ...) fn(f, ...)
-  expect_identical(g(exprs, a = 1, foo = bar), list(a = 1, foo = quote(bar)))
+test_that("exprs() captures arguments forwarded with `...`", {
+  wrapper <- function(...) wrapper(...)
```
hadley: Why doesn't this cause infinite recursion?

lionel-: This looks like a test refactor gone wrong. It infloops if called directly but not when executed by testthat (the old version).

R/expr-interp.R
lionel-: @hadley How about:

```
"`UQ()` can only be called as part of a quasiquoted argument"
```

Or:

```
"`UQ()` can only be called within a quasiquoted argument"
```

hadley: The second I think. Maybe used instead of called?

src/dots.c
hadley: Yeah, that's useful.

hadley:
LGTM: feel free to merge, then rebase the other PRs and I'll review them.

PR #299, opened 2017-11-26
Update NAMESPACE

dan87134:
is_condition is listed in the documentation as an rlang function but it is not exported from rlang. It is available as rlang:::is_condition though.


lionel-:
Good catch. The NAMESPACE file is automatically generated. You need to add an `@export` statement above the function and redocument.

dan87134:
Thanks... by "you" do you mean me?... I'm just using the public install from https://www.r-project.org/

lionel-:
Yes if you would like this PR merged. But I can do it that's no worries at all.

dan87134:
Ah, I see what to do now... next time :-)

PR #301, opened 2017-11-27
Deep unquoting of names

lionel-:
Includes #297, closes #279.

To treat `expr(list(a := foo))` as synonym of `expr(list(a = foo))` and allow name-unquoting everywhere.

I'm not too fond of the `unquote_names` argument name, but I can't think of anything better.

PR #302, opened 2017-11-27
Remove one layer of parentheses around !!

lionel-:
Includes #297 and #301.

With this change `(!! sym("foo"))(arg)` expands to `foo(arg)` instead of `(foo)(arg)`. It is possible to double wrap to force the parentheses: `((!! 1)) + 1` expands to `(1) + 1`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -110,10 +110,12 @@
   the first argument of an S3 method (which is necessarily evaluated
   in order to detect which class to dispatch to). (#295, #300).
 
-* Parentheses around `!!` are automatically removed. This is useful
-  for unquoting function names: `(!! sym("name"))(arg)`. If you need
-  the parenthesis wrapping, you can add a second layer since only one
-  layer is removed: `((!! x)) + 1` (#296).
+* Parentheses around `!!` are automatically removed. This makes the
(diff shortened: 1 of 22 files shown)


NEWS.md
```diff
@@ -108,6 +117,12 @@
   the first argument of an S3 method (which is necessarily evaluated
   in order to detect which class to dispatch to). (#295, #300).
 
+* Parentheses around `!!` are automatically removed. This is useful
```
hadley: I think you need to somehow mention that removing the automatic parens removal will not change precedence, and parens will be inserted when printing if needed

``` r
library(rlang)
expr(-!!expr(x+1))
#> -(x + 1)
```

hadley: "is useful" -> "makes the generated call cleaner"

NEWS.md
```diff
@@ -108,6 +117,12 @@
...
+  for unquoting function names: `(!! sym("name"))(arg)`. If you need
+  the parenthesis wrapping, you can add a second layer since only one
```
hadley: Why would you need it?

lionel-: I was thinking about generating code for some DSLs such as lme4. Just tried it with lme4 and it doesn't require explicit parentheses though. I'll remove the note.

Issue #309, opened 2017-12-04
Safe quosures and non-formula implementation


lionel-:
These would never be overscoped to ensure deterministic lookup of symbols and to make it possible to use frame-sensitive functions like return() or parent.frame(). The new quosures already support the latter when there is no overscope though.

lionel-:
This will require https://github.com/tidyverse/rlang/tree/perf-eval-tidy

lionel-:
Will revisit in the future. Safe quosures would have the big downside of being very noisy in recorded calls and backtraces.

PR #310, opened 2017-12-04
Use non-formula implementation for quosures

lionel-:
* Greatly improves the speed of evaluation when there is no overscope. Closes https://github.com/tidyverse/purrr/issues/388.

* Quosures no longer require evaluation within an overscope, which means the original environment can be preserved. Implementation of "safe" quosures that would never be overscoped even when `data` is supplied should now be a few lines patch (see #309).

* `as_overscope()` now takes `data` instead of a quosure which makes more sense. This is a breaking change but I don't see any CRAN package using `as_overscope()`.

* The main downside is that raw quosures don't print nicely now that they are calls to a thunk rather than to the symbol `~`. The quosure print method hides this but users will encounter strange print output if they embed quosures in raw expressions, e.g. `expr(list(!! quo(foo))`.

* Another downside is the printing bug that affects all language objects with a class and causes them to be evaluated during printing if they are wrapped in a list, e.g. `list(quo(foo))`. This doesn't cause issues with formulas but will trigger strange errors with the new quosures. We'll propose a fix to the R printer soon. The `quosures` print method works around it so lists returned by `quos()` are not affected.

* The print method for quosures is now a bit nicer. It appears like a simple expression with the environment displayed beneath.


hadley:
Could you include a brief summary of how this effects revdeps? (i.e. by running revdepcheck)

lionel-:
This will require a transition strategy for dplyr as the C++ code assumes formula-quosures. I moved this PR to a branch: https://github.com/tidyverse/rlang/tree/perf-eval-tidy

PR #315, opened 2017-12-05
Deprecate UQE() and soft-deprecate UQ() and UQS()

lionel-:
* `UQE()` is deprecated and now issues a warning.

* The prefix form `` `!!`() `` is now an alias of `!!` rather than `UQE()`. The prefix form `` `!!!`() `` is now officially part of the tidy eval API.

* `UQ()` and `UQS()` are soft-deprecated. They issue a warning if `lifecycle_verbose_retirement` is TRUE which is enabled by the last version of testthat (but not on CRAN to avoid check failures).

* Added lifecycle documentation topics. Each life cycle stage has its own page listing the functions belonging to the category. Maturing functions do not have a topic as rlang is currently maturing.


NEWS.md
```diff
@@ -126,6 +126,13 @@
   are printed as a flat expression with the environment printed below.
   Previously they would appear as formulas.
 
+* `UQ()` and `UQS()` are soft-deprecated in order to make the syntax
+  of quasiquotation more consistent. The prefix forms are now ``
+  `!!`() `` and `` `!!!`() `` which is consistent with other R
+  operators (e.g. `` `+`(a, b) `` is the prefix form of `a + b`).
+  These operators are scheduled for deprecation with a warning in
+  rlang 0.3.0.
```
hadley: That's possibly too soon.

R/expr-interp.R
```diff
@@ -13,30 +13,42 @@
...
-#'   expression. Each component of the vector is embedded as its own
-#'   argument in the surrounding call. If the vector is named, the
-#'   names are used as argument names.
 #'
 #' @section Theory:
 #'
-#' Formally, `quo()` and `expr()` are quasiquote functions, `UQ()` is
-#' the unquote operator, and `UQS()` is the unquote splice operator.
+#' Formally, `quo()` and `expr()` are quasiquote functions, `!!` is
+#' the unquote operator, and `!!!` is the unquote splice operator.
```
hadley: unquote-splice is a bit clearer I think

R/expr-interp.R
```diff
@@ -13,30 +13,42 @@
...
 #' These terms have a rich history in Lisp languages, and live on in
 #' modern languages like
 #' [Julia](https://docs.julialang.org/en/stable/manual/metaprogramming/)
 #' and
 #' [Racket](https://docs.racket-lang.org/reference/quasiquote.html).
 #'
+#'
+#' @section Life cycle:
+#'
+#' * `UQ()` and `UQS()` were soft-deprecated in rlang 0.2.0 in order
```
hadley: I think you also need to note here that the prefix forms aren't as important now because of the other changes to the syntax (i.e. operator precedence and parens stripping).

R/lifecycle.R
```diff
@@ -0,0 +1,43 @@
+#' Life cycle of the rlang package
+#'
+#' The rlang package is currently maturing. Unless otherwise stated,
+#' this applies to all its exported functions. See [lifecycle-stable]
+#' for a list of functions marked as stable.
+#'
+#' @name lifecycle
+NULL
+
+#' Stable functions
```
hadley: Might be better to documents these all one file? (With sections for different phases)

R/quo.R
```diff
@@ -122,31 +122,26 @@
 #' eval_tidy(quo)
 #'
 #'
+#' # The `!!` operator can be used in prefix form as well. This is
+#' # useful for using it with `$` for instance:
+#' var <- sym("cyl")
+#' quo(mtcars$`!!`(var))
```
hadley: cf. `` quo(`$`(mtcars, !!var) `` ?

NEWS.md
lionel-: Should we give a timeframe instead? E.g. 6 months or 12?

PR #327, opened 2017-12-22
parse_expr returns language object not formula

dan87134:
There is no formula in the object returned by parse_expr or parse_exprs. The typeof(parse_expr("asdf")) is language and its class is call.

I don't know how to add the @return for parse_quosure but parse_quosure returns a language object too but its class is quosure which derives from formula.

### DIFF
--- a/R/parse.R
+++ b/R/parse.R
@@ -21,8 +21,8 @@
 #' @param env The environment for the formulas. Defaults to the
 #'   context in which the parse_expr function was called. Can be any
 #'   object with a `as_env()` method.
-#' @return `parse_expr()` returns a formula, `parse_exprs()` returns a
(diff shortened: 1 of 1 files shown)


lionel-:
This should be "expression" not language object. The parser does not necessarily return calls, it also returns symbols (e.g. `parse_expr("foo")`) and vectors (e.g. `parse_expr(1)`).

dan87134:
I missed that it could parse_expr could also produce a symbol or an atomic.

parse_expr(1) produces and error but

parse_expr("1") produces a double

and parse_expr("1 + a")  produces a language object

But parse_expr doesn't produce a formula.

lionel-:
Could you change it to "expression" please?

PR #328, opened 2017-12-26
Fix bang bang precedence

lionel-:
This adds a rotating version of interpolation / expansion for operators with problematic precedence. The general expansion + rotation strategy is described in [expr-interp-rotate.c](https://github.com/tidyverse/rlang/compare/master...lionel-:fix-bang-bang-precedence?expand=1#diff-52691165fd88c9bb6ace5dc84ed116f0).

I am using [Linux kernel documentation style](https://www.kernel.org/doc/html/v4.9/kernel-documentation.html#writing-kernel-doc-comments) for embedded documentation. After a quick review of available options it seems like the cleanest syntax. A [perl script](https://github.com/torvalds/linux/blob/master/scripts/kernel-doc) can be used to generate Sphinx output from the C files. This might be a good option to generate the documentation of the C wrapper library later on.

* [dots.c](https://github.com/tidyverse/rlang/compare/master...lionel-:fix-bang-bang-precedence?expand=1#diff-3b4f8126fe47843caf246e738300bee2) / [dots.h](https://github.com/tidyverse/rlang/compare/master...lionel-:fix-bang-bang-precedence?expand=1#diff-304617922aa93ebf3b6880ddb65ff217) / [expr-interp.c](https://github.com/lionel-/rlang/blob/cbb6dcba43ebc0b939e9a89dd1f6960041d5bd30/src/expr-interp.c): Add a FIXUP operation type. The expansion + rotation algorithm is only used for problematic operators (binary ops whose precedence is between prec(`!`) and prec(`!!`) to avoid the overhead for the normal unquoting case.

* [parse.c](https://github.com/tidyverse/rlang/compare/master...lionel-:fix-bang-bang-precedence?expand=1#diff-984682cffd12ab2ca3a98d41cd829a96) / [parse.h](https://github.com/tidyverse/rlang/compare/master...lionel-:fix-bang-bang-precedence?expand=1#diff-fd510818d73257282d4b4b827e37d05b): `r_which_operator()` tokenises a symbol to an `enum r_operator`. The operation precedence table `r_ops_precedence` contains information for programming on the language with operators (binding power, unary, delimited).

* [rlang.h](https://github.com/tidyverse/rlang/compare/master...lionel-:fix-bang-bang-precedence?expand=1#diff-50ed1f2855cf86cb0d682acdfecaee1d) / [init.c](https://github.com/tidyverse/rlang/compare/master...lionel-:fix-bang-bang-precedence?expand=1#diff-885908120835b66f90958a39baeb2018): Add a static assertion macro borrowed from Linux source code and use it to check that tables (such as operation precedence table) are correctly initialised at compile-time. To work around the limitations of static assertions the elements of the table are checked at load time.

* [expr-interp-rotate.c](https://github.com/tidyverse/rlang/compare/master...lionel-:fix-bang-bang-precedence?expand=1#diff-52691165fd88c9bb6ace5dc84ed116f0): Implementation of the expand + rotate algorithm for binary operators with problematic precedence.

* [test-tidy-unquote.R](https://github.com/tidyverse/rlang/compare/master...lionel-:fix-bang-bang-precedence?expand=1#diff-20d2c7b15a736c94480c1004c35eab17): Unit tests for the new operator precedence.
  [test-c-api.R](https://github.com/tidyverse/rlang/pull/328/files#diff-810e014010775ef58cadc8b13416a131): Unit tests for the `enum r_operator` tokeniser.


lionel-:
Hi @jimhester. Would you like to be my first reviewer for the "teammate under review" exercise? This PR is a bit complicated but I've put a lot of effort into documenting it. Unfortunately the links to the PR files don't seem to work :/

Besides the algorithm, I'm looking for comments on C style and idioms, use of kernel-doc style for embedded documentation, and anything you can help me with :)

src/rlang/parse.c
```diff
@@ -0,0 +1,332 @@
...
+    return R_OP_NONE;
+  }
+
+  const char* name = r_sym_c_str(head);
+  size_t len = strlen(name);
+  bool is_unary = r_node_cddr(call) == r_null;
+
+  switch (name[0]) {
+  case 'w':
+    if (strcmp(name, "while") == 0) {
```
jimhester: Why are we converting the symbol to a character and using strcmp, can we not lookup the symbol with `install` and compare it directly?

jimhester:
Is it possible to use roxygen for the documentation? (http://r-pkgs.had.co.nz/src.html#cpp-man) If you mark them internal they should not show up in the help index and that gives us a uniform source of documentation.

src/rlang/parse.c
lionel-: Hmm I don't think we can switch on sexp pointers because they are not compile-time constants.

jimhester: Yes that is true, you would have to use a different method of control flow, it is probably fine as is.

lionel-:
These are not generated by Roxygen but by `Rcpp::compileAttributes()`. The main problem of roxygen is that it is targetted at R rather than C, so has no specific support for enums or structs. We'll also want to generate a website to document the exported API later on and I expect this task to be easier with sphinx than with roxygen.

src/rlang/parse.c
lionel-: Yup if-else would be slow with this many cases.

lionel-:
> If not I would think doxygen would have more similarity to roxygen than Sphinx.

The documentation is expressed as kernel-doc syntax and I think it looks more similar to roxygen than doxygen.

jimhester:
roxygen syntax was based on doxygen though, that is why it is called roxygen after all. Recent versions of doxygen also support markdown syntax https://www.stack.nl/~dimitri/doxygen/manual/markdown.html much like that which was recently added to roxygen.

Maybe down the line it would be worth extending roxygen support to compiled code. Assuming you don't try to do some of the more sophisticated lookup roxygen it does for R it should really only be a matter of changing the characters used to delimit comments.

lionel-:
You're right, when I looked in doxygen I stumbled onto the backslash syntax which wasn't very roxygen-like but now I see that's only for non-inline documentation. So Doxygen is indeed a good candidate for documenting the library, especially since its syntax is more featureful than kernel-doc's.

> Maybe down the line it would be worth extending roxygen support to compiled code.

That would be nice.

PR #331, opened 2017-12-28
Soft-deprecate accessing of quosures with formula accessors

lionel-:
- And deprecate `is_quosureish()` and `as_quosureish()` (they are not used by CRAN packages).
- And export quosure-specific getters and setters. Compared to generic versions they only accept quosures and are a bit more efficient. `quo_get_expr()` and `quo_get_env()` thus replace `f_env()` and `f_rhs()`.


lionel-:
Yes and also more generally when you are writing a function where the object you are operating on is supposed to be a quosure. It feels like specific accessors help with readability as you don't have to worry about kind of objects are getting through there.

I think these accessors are for package writers and we'd still use the generic ones for user documentation and scripts.

In a subsequent PR I'll move their implementation to C side and export C callables for use in other packages like dplyr.

PR #332, opened 2017-12-29
Pretty printer for quosures

lionel-:
After using the flat printing style for a while I think that hiding quosures is actually more confusing. Otoh displaying quosures as formulas is confusing as well because R users don't think of formulas as something that evaluates, e.g. when they see `list(~letters)` they expect to get a list containing `~letters` rather than the letters vector.

So how about printing quosures with a leading `^`? This PR uses a naive approach to printing but it should work fine in most cases until we have a custom expression printer in rlang:

```r
quo(foo)
#> ^foo

local(quo(foo))
#> ^foo
#> <environment: 0x7fc9d305d008>

quo(foo(!!quo(bar())))
#> ^foo(^bar())
```


hadley:
Could you use crayon to add some colour? I think that would be v. helpful.

What if we always printed the environment line and added something to make it clear this object is a quosure?

```R
quo(foo)
#> ^foo
#> <quo_env: 0x7fc9d305d008>
```

lionel-:
> What if we always printed the environment line and added something to make it clear this object is a quosure?

I think it is better to use the same display as formulas and functions in order to make quosures as familiar-looking as possible. With quosures being printed with a leading `^` users should quickly learn to recognise quosures, e.g. we don't mistake a raw expression with a formula thanks to the leading `~`. What do you think?

hadley:
The problem is that if you don't know you have a quosure, there's nothing in the output that will tell you, and no way to search the docs or google to find out.

hadley:
Let me know when this is ready for review

lionel-:
Here you go @hadley 

- [utils.R](https://github.com/lionel-/rlang/blob/2307373184c211e52cd0802657fed06a61176167/R/utils.R): Add r6lite class with basic method lookup and inheritance.

- [expr-deparse.R](https://github.com/lionel-/rlang/blob/2307373184c211e52cd0802657fed06a61176167/R/deparse.R): Implements the deparser as an r6lite class. Unit tests are in [tests/testthat/test-deparse.R](https://github.com/lionel-/rlang/blob/2307373184c211e52cd0802657fed06a61176167/tests/testthat/test-deparse.R). This deparser always wraps inlined objects with angular `<` `>` brackets, only adds parentheses when needed, and respects `peek_option("width")` correctly.  S3 objects are printed using the `pillar::type_sum()` generic when available. The deparser is extensible by deriving from the class.

- [quo.R](https://github.com/lionel-/rlang/blob/2307373184c211e52cd0802657fed06a61176167/R/quo.R) Derives a deparser that colourises quosures according to their environment. The global env is always printed normally while other environments are always printed in colour. When there are more than 4 different quosure environments nested in the expression the remaining quosures from different environments are all printed in blurred-italic.

hadley:
I skimmed the code and it looks fine. I had two questions:

* Can we have a function that tells you if an expression is deparsable? (i.e. it only contains symbols + calls + constants)

* Can you render a few examples that shows the pretty printed in action?

lionel-:
> Can we have a function that tells you if an expression is deparsable?

All expressions are deparsable but I suppose you meant "parsable"? I think the current name for the predicate you describe is `is_expr()` (should be renamed to `is_expression()`). IIRC our last conversation neither of us were completely happy with that name but we didn't come up with something better.

hadley:
I don't think `is_expr()` is quite right because `is_expr(expr(list(!!list())))` returns TRUE - I want something that only returns TRUE if it can be deparsed into something that can be reparsed (i.e. you can roundtrip the expression to text and back)

Should we also export a function that you can use with expressions?

lionel-:
This is how quosures are colourised:

quo-printer

lionel-:
And this is how objects are printed:

sexp-printer

hadley:
Looks good!

PR #333, opened 2017-12-29
Export C accessors for quosures

lionel-:
- Add `r_register_c_callable()` and `r_register_r_callables()` to the wrapper library. Also `r_fn_ptr_t` as alias to `DL_FUNC`.

- Move implementation of quosure accessors to C side and export them as C callables.

- Add static versions in the headers library so users of the library don't have to deal with the callables. These static versions call pointers which are initialised within `r_init_library()`. This function should be called during initialisation of all packages that use the wrapper library, it initialises the global constants such as symbols.


lionel-:
Maybe `r_fnptr_t` would be more idiomatic than `r_fn_ptr_t`, cf https://en.wikibooks.org/wiki/C_Programming/stdint.h

lionel-:
Rearranged the PR to make the relevant changes clearer.

In the private C files:

- `src/internal/quo.c`: Private definitions for the quosure accessors (not part of this PR).
- `src/export/init.c`: Registers the quosure accessors defined privately as C callables.


In the (to-be) public headers-only C library:

- `src/lib/quo.h`: Declares pointers to the quosure accessors.
- `src/lib/quo.c`: Defines the pointers.
- `src/lib/rlang.h`: Defines an init function that should be called when the client package (the one compiling against the exported C library) is loaded. It fills up the pointer functions of the quosure accessors by requesting the pointers to the rlang package with `r_peek_c_callable()` (the name of `R_GetCCallable()` in the library). It is this init function that will deal with multiple versions of the signatures if they are ever updated.


In addition:

- `src/export/exported-tests.c`: Exports C functions wrapping the function pointers for testing the C accessors wrappers from testthat. I had to wrap the function pointers because the table of registered routines requires constant function pointers.

src/lib/rlang.h
```diff
@@ -85,8 +87,15 @@ static inline sexp* KEEP_N(sexp* x, int* n) {
...
-static inline void rlang_init_library() {
+// This *must* be called before making any calls to the functions
+// provided in the library
+static inline void r_init_library() {
   r_dot_environment_sym = r_sym(".Environment");
+
+  r_quo_get_expr = (sexp* (*)(sexp*)) r_peek_c_callable("rlang", "rlang_quo_get_expr");
+  r_quo_set_expr = (sexp* (*)(sexp*, sexp*)) r_peek_c_callable("rlang", "rlang_quo_set_expr");
+  r_quo_get_env = (sexp* (*)(sexp*)) r_peek_c_callable("rlang", "rlang_quo_get_env");
+  r_quo_set_env = (sexp* (*)(sexp*, sexp*)) r_peek_c_callable("rlang", "rlang_quo_set_env");
```
gaborcsardi: I think it would be more readable to typedef some function types.

tests/testthat/test-c-api.R
```diff
@@ -174,3 +174,24 @@ test_that("r_which_operator() returns correct tokens", {
...
+  r_quo_set_env <- function(quo, env) {
+    .Call(rlang_test_r_quo_set_env, quo, env)
+  }
+  quo <- quo(foo)
+
+  expect_identical(r_quo_get_expr(quo), quo_get_expr(quo))
+  expect_identical(r_quo_get_env(quo), quo_get_env(quo))
+  expect_identical(r_quo_set_expr(quo, NULL), quo_set_expr(quo, NULL))
+  expect_identical(r_quo_set_env(quo, empty_env()), quo_set_env(quo, empty_env()))
+})
```
gaborcsardi: I think it would be good to test the C api explicitly by creating a toy package that uses it. Either by including a toy package in `rlang`, in `tests/testthat/fixtures`, or a helper function that creates such a package on the fly.

lionel-: I think it's a good idea but it is a bit involved to achieve. IIUC I need to build rlang and install it in a fake library, then build and check the embedded package in the context of that library. Since the library is not public yet as it's still in an early stage it's probably not worth the effort. I'll try to make it work at a later stage.

src/lib/rlang.h
lionel-: my general approach with the library is to give a chance to C syntax. I.e. trying not to hide structs, enums, and pointers with typedefs. So far I am enjoying the explicitness it brings. In this  particular case I also think anonymous function pointers are readable as you can immediately see the number and types of arguments and return value.

tests/testthat/test-c-api.R
```diff
@@ -174,3 +174,55 @@ test_that("r_which_operator() returns correct tokens", {
...
+  sink(temp)
+  on.exit({
+    sink()
+    close(temp)
+  })
+
+  ## tools::testInstalledPackage() can't find the test package when
+  ## working with a temp lib:
+
+  ## old_libpaths <- .libPaths()
```
lionel-: @gaborcsardi The test globally works but if I uncomment this paragraph to install the embedded package in a temporary library, `tools::testInstalledPackage()` can't find the package.

src/lib/rlang.h
gaborcsardi: Right. I agree that this is clean now. But I suppose the api will grow. Will you ever have non-sexp arguments for example? Making a SEXP in C code is not always needed. 

Anyway, I agree that this is not an issue right now.

tests/testthat/test-c-api.R
jimhester: What happens if you pass the temp library directly to `lib.loc` rather than using `.libPaths()`?

lionel-: Same thing, it can't find the package.

jimhester: I would also not use `devtools` to do the install, just use `install.packages("pkg_path", repos = NULL, lib = temp_lib)`.

jimhester: probably also need `INSTALL_ops = "--install-tests"`

lionel-: Thanks! I didn't know `install.packages()` would do the building. Got rid of the devtools `Suggest`. It still can't find the package in the temporary lib though :/

PR #334, opened 2018-01-04
Implement enquos()

### LINKED ISSUE #323: Add enquos()
Would be like `quos()` but would only work when supplying the `...` symbol (whereas `quos()` also supports direct arguments, e.g. `quos(..., foo, bar)`. The parallel with `enquo()` should be less confusing to newcomers.

lionel-:
It is awkward to transition from `enquo()` to `quos()` while teaching tidy eval, and users have reported this naming to be confusing. `enquos()` should make things easier. The difference between the two are:

- `enquos()` only takes symbols while `quos()` accepts any expressions.
- supplying symbols to `enquos()` is like using `enquo()` on each of them. I.e. it captures arguments one level up.
- both capture arguments passed to `...` the same way.

```r
capture_quos <- function(foo, ...) quos(foo, ...)
capture_quos(bar, quux)
#> [[1]]
#> <quosure>
#>   expr: ^foo
#>   env:  0x7fdab7ca0238
#>
#> [[2]]
#> <quosure>
#>   expr: ^quux
#>   env:  global

capture_enquos <- function(foo, ...) enquos(foo, ...)
capture_enquos(bar, quux)
#> [[1]]
#> <quosure>
#>   expr: ^bar
#>   env:  global
#>
#> [[2]]
#> <quosure>
#>   expr: ^quux
#>   env:  global
``` 

Closes #323


hadley:
Do we need `enexprs()` and `ensyms()` for symmetry?

lionel-:
I thought about that as well, I'll add an issue for those.

PR #339, opened 2018-01-05
Add documentation of `.env` vs. unquoting

karldw:
I added a little documentation to capture @lionel-'s comments from https://github.com/tidyverse/dplyr/issues/3286. Feel free to merge if this is helpful, not worries if it's not.

I added a few examples that use magrittr pipes (`%>%`), but `\dontrun{}`-ed run them to avoid making magrittr a dependency.

### DIFF
--- a/R/eval-tidy.R
+++ b/R/eval-tidy.R
@@ -85,16 +85,20 @@
 #' eval_tidy(quo(.data$cyl), mtcars)
 #' eval_tidy(quo(.env$cyl), mtcars)
 #'
-#' # Note that instead of using `.env` it is often equivalent and often
+#' # Note that instead of using `.env` it is often equivalent and may be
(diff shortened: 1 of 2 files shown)


lionel-:
The formulation is not quite right: it's not `.env` that doesn't look through parents, but the `$` operator. Also it looks like the two magrittr examples are the same.

karldw:
Oops, thanks.  Is it correct to say "`.env$` will not look in a parent environment, while unquoting will."?

lionel-:
Secondly, subsetting `.env` with the `$` operator may be brittle because `$` does not look through the parents of the environment. Using `.env$name` in a magrittr pipeline is an instance where this poses problem because the magrittr pipe currently evaluates its operands in a child of the current environment (this is where it defines the pronoun `.`).

lionel-:
Thanks!

PR #342, opened 2018-01-07
Fix capture of lexically scoped arguments and dots

lionel-:


NEWS.md
```diff
@@ -191,6 +191,14 @@
   In addition, `enexprs()` is like `enquos()` but like `exprs()` it
   returns bare expressions. And `ensyms()` expects strings or symbols.
 
+* It is now possible to use `enquo()` within a magrittr pipe. Stated
```
hadley: Can you include the example from slack?

PR #347, opened 2018-01-08
Update eval_bare() documentation

lionel-:
And mark `with_env()`, `locally()` and `invoke()` as experimental.

### DIFF
--- a/R/eval.R
+++ b/R/eval.R
@@ -80,6 +80,12 @@ eval_bare <- function(expr, env = parent.frame()) {
 #' evaluation stack, and has stack-consistent semantics. It is thus
 #' more suited for experimenting with the R language.
 #'
+#'
+#' @section Life cycle:
+#'
+#' These functions are experimental. Expect API changes.
+#'
+#'
 #' @inheritParams eval_bare
 #' @param env An environment within which to evaluate `expr`. Can be
 #'   an object with an [get_env()] method.
--- a/R/eval.R
+++ b/R/eval.R
@@ -31,6 +31,11 @@
 #' frame on the stack that confuses stack-sensitive functions. We call
 #' these evaluation semantics "stack inconsistent".
 #'
+#'
+#' @section Life cycle:
+#'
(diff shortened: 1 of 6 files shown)


R/eval.R
```diff
@@ -1,36 +1,67 @@
...
+#' `eval_bare()` is a lower-level version of function [base::eval()].
+#' Technically, it is a simple wrapper around the C function
+#' `Rf_eval()`. It has different evaluation semantics than `eval()`.
+#'
+#' * It does not accept auxiliary data and has no `data` argument.
+#'
+#' * It can't evaluate `break` or `next` expressions even if called
+#'   within a loop.
+#'
+#' * It evaluates `return()` calls in frame environments on the call
```
hadley: It would be useful to note what `eval()` does here.

R/eval.R
```diff
@@ -136,6 +173,17 @@ locally <- function(expr) {
 #' uses the same strategy as [eval_bare()] to evaluate with minimal
 #' noise.
 #'
+#'
+#' @section Life cycle:
+#'
+#' `invoke()` is experimental. Now that we understand better the
```
hadley: Shouldn't this be questioning?

R/eval.R
```diff
@@ -1,36 +1,67 @@
...
-#' stack when `expr` is evaluated. Thus, any command that looks up
-#' frames on the stack (stack sensitive functions) may find the
-#' parasite frame set up by `eval()` rather than the original frame
-#' targetted by `env`. As a result, code evaluated with `base::eval()`
-#' does not have the property of stack consistency, and stack
-#' sensitive functions like [base::return()], [base::parent.frame()]
-#' may return misleading results.
+#' @description
+#'
+#' `eval_bare()` is a lower-level version of function [base::eval()].
```
hadley: I think this needs a sentence or two about why you might use it instead of `eval()`. 

Currently the description is probably a bit long, so you could then move the bulleted list to details.

PR #348, opened 2018-01-09
Rename from `lang` to `call`

lionel-:
Except for `lang_type_of()`, `switch_lang()` and `coerce_lang()` which are marked as experimental instead.

Do you think any of the accessors should be marked experimental as well @hadley? I'm not sure we need the `call_head()` and `call_tail()` accessors. They do provide safer versions of `node_car()` and `node_tail()` for people that want to experiment with R internal representations of calls.


NEWS.md
```diff
@@ -210,6 +210,21 @@
   reflect that it is a lossy operation that flattens all nested
   quosures.
 
+* The particle "lang" has been renamed to "call":
+
+    - `lang()` has been renamed to `call2()`.
+    - `new_language()` has ben renamed to `new_call()`.
+    - `is_lang()`, `is_unary_lang()`, and `is_binary_lang()` have been
```
hadley: Should we take this opportunity to eliminate `is_unary_call` and `is_binary_call` in favour of the `n` argument to `is_call`?

R/call.R
```diff
@@ -83,30 +97,30 @@ lang <- function(.fn, ..., .ns = NULL) {
...
+  new_call(.fn, as.pairlist(dots_list(...)))
 }
-#' @rdname lang
+#' @rdname call2
 #' @param head A [callable][is_callable] object: a symbol, call, or
 #'   literal function.
 #' @param tail A [node list][pairlist] of arguments.
 #' @export
-new_language <- function(head, tail = NULL) {
+new_call <- function(head, tail = NULL) {
```
hadley: This should probably be `@keyword internal` even if it means documenting it somewhere else. Very very few people will need to use this.

NEWS.md
lionel-: Sounds good.

PR #350, opened 2018-01-10
Update tidy eval low level API

lionel-:
- Dictionaries are replaced with a simpler `rlang_data_pronoun` class. You can create one with `as_data_pronoun()`. This function is not generic but users can create their own pronoun class if needed (I don't think any package needs a custom pronoun at the moment). `as_data_pronoun()` is also available as C callable which is needed for dplyr.

- Overscope stuff is renamed to data mask. `overscope_eval_next()` is retired because `eval_tidy()` now supports data masks as `data` argument.

Another PR will update the documentation.


R/eval-tidy.R
```diff
@@ -172,14 +147,14 @@ eval_tidy_ <- function(expr, bottom, top = NULL, env = caller_env()) {
...
-#' * `as_overscope()` is the function that powers [eval_tidy()]. It
+#' * `as_data_mask()` is the function that powers [eval_tidy()]. It
 #'   could be useful if you cannot use `eval_tidy()` for some reason,
 #'   but serves mostly as an example of how to build a dynamic scope
 #'   for tidy evaluation. In this case, it creates pronouns `.data`
 #'   and `.env` and buries all dynamic bindings from the supplied
 #'   `data` in new environments.
 #'
-#' * `new_overscope()` is called by `as_overscope()` and
+#' * `new_overscope()` is called by `as_data_mask()` and
```
hadley: Should be `new_data_mask()`?

R/lifecycle-retired.R
```diff
@@ -137,6 +137,141 @@ lang_tail <- function(lang) {
   node_cdr(call)
 }
 
+#' Create a data mask
```
hadley: Create an overscope?

PR #362, opened 2018-01-16
fix typo in lang_head() and lang_tail()

kevinykuo:
Will be refactoring to `node_cdr()` on our end but figured this is worth fixing in case it breaks stuff in the wild.

cc @lionel- this is the issue I mentioned

Closes https://github.com/rstudio/sparklyr/issues/1221


lionel-:
Thanks! It's a good idea to test the retired functions. Are you sure you need `node_cdr()`? It's unlikely you need it.

lionel-:
I'm not sure why sparklyr didn't show up in the revdep checks...  I'll launch them again.

PR #375, opened 2018-01-21
Switch to "mask" in remaining rlang API

lionel-:
Following the switch from overscope to data mask, it seems we should also switch from "symbol burrying" to "symbol masking". I've never been happy with the name of `env_bury()` but `env_mask()` feels right.


R/env.R
```diff
@@ -704,17 +704,20 @@ with_bindings <- function(.expr, ..., .env = caller_env()) {
   .expr
 }
 
-#' Mask bindings by defining symbols deeper in a scope
+#' Mask symbols with a new child environment
+#'
+#' @description
+#'
+#' `env_mask()` is like [env_bind()] but it creates the bindings in a
```
hadley: Should the prefix be `env_`? It doesn't actually work on environments, but objects that have environments?

lionel-: A lot of our `env_` functions are generic in this way. Do you want to reconsider this UI?

hadley: I don't want to reconsider anything major; maybe it's just the framing here that threw me off? `env_mask()` now seems very similar to `child_env()`

lionel-: That's true. Probably best not change anything until we figure out `env` vs `child_env` then.

hadley: I think the confusing thing about  `env_bury(x)` is that it returns the same type as `x`, not an environment. Do we have other functions that behave like that?

lionel-: I see `env_bind()` and variants, and `env_unbind()`.

PR #389, opened 2018-01-26
Document tidy dots

lionel-:


R/dots.R
```diff
@@ -1,23 +1,70 @@
-#' Extract dots with splicing semantics
+#' Extract tidy dots
```
hadley: "Capture dots, tidily" ?

(Although we've used capture to imply unevaluated in other places)

R/dots.R
```diff
@@ -1,23 +1,70 @@
...
+#' - You can unquote names by using the [unquote][quasiquotation]
+#'   operator `!!` on the left-hand side of `:=`.
+#'
+#' We call quasiquotation support in dots **tidy dots** semantics and
+#' functions taking dots with `dots_list()` tidy dots functions.
+#' Quasiquotation is an alternative to `do.call()` idioms and gives
+#' the users of your functions an uniform syntax to supply a variable
+#' number of arguments or a variable name.
+#'
+#' Note that while all tidy eval [quoting functions][quotation] have
```
hadley: I think this para could move to `@details`. I think in most cases we want to keep the description pretty short since it's at the very top of the docs and you have to scroll past to get to everything else.

R/dots.R
```diff
@@ -1,23 +1,70 @@
...
+#' tidy dots semantics, not all tidy dots functions are quoting
+#' functions. `dots_list()` is for standard functions, not quoting
+#' functions.
+#'
+#'
+#' @section Comparison to tidy eval splicing:
+#'
+#' The `!!!` operator works differently in _standard_ functions taking
```
hadley: This section needs to more concretely address _why_ a user should care about these differences

R/dots.R
```diff
@@ -1,23 +1,70 @@
...
 #'
 #'
 #' @section Life cycle:
 #'
-#' * `dots_splice()` is experimental. It is part of our experiments
-#'   with dots semantics. We now lean towards adopting a single type
-#'   of dots semantics (those of `dots_list()`) where splicing is
-#'   explicit.
+#' * `dots_list()` returns named dots, even if no names were supplied.
+#'   This behaviour is for consistency with dots returned by [exprs()]
```
hadley: Isn't it inconsistent? `exprs()` has `.named = FALSE`

R/dots.R
```diff
@@ -57,37 +118,57 @@ dots_list <- function(...,
...
-#' dots_splice(!!! x, 3)
-#' dots_splice(x, 3)
 dots_splice <- function(...,
                         .ignore_empty = c("trailing", "none", "all")) {
   dots <- .Call(rlang_dots_flat_list, environment(), FALSE, .ignore_empty, TRUE)
   names(dots) <- names2(dots)
   dots
 }
 
+#' Splice a list within a vector
```
hadley: Should we document `dots_splice()` here? And then mark this whole API as questioning

R/dots.R
lionel-: Yes I think "extract" is better than "capture" to avoid confusion.

R/dots.R
lionel-: `.named = FALSE` still names the output with an empty vector of names, we should probably tweak the UI to make it more explicit.

R/dots.R
lionel-: I provided some pointers (e.g. perf implications) but moved the section to the topic in questioning stage that now includes `splice()` and `dots_splice()`. I moved it there because that's a rather technical discussion that shouldn't matter for most users and we can find a better place for it later.

R/dots.R
hadley: But what are we extracting them from? How about collect?

lionel-: From the `...` pairlist of arguments :)

"collect" sounds good to me.

PR #402, opened 2018-02-06
Rename ll() to list2()

lionel-:


R/lifecycle-retired.R
```diff
@@ -431,3 +431,16 @@ cnd_message <- function(.type = NULL, ..., .msg = NULL) {
     call. = FALSE)
   message_cnd(.type = .type, ..., .msg = .msg)
 }
+
+#' Deprecated list constructor
+#'
+#' `ll()` has been soft-deprecated and renamed to [list2()] in rlang
```
vspinu: What does soft-deprecation mean? Would you mind keeping it as an alias instead? `ll` is handy and visually appealing shorthand.

lionel-: It means it's going to go through a deprecation process. I don't mind keeping `ll()` around as a shorthand though, perhaps undocumented.

PR #447, opened 2018-03-08
testthat >= 2.0.0 is required for tests

QuLogic:
This is due to the [`expect_reference` call in `test-quotation.R`](https://github.com/r-lib/rlang/blob/master/tests/testthat/test-quotation.R#L434).


lionel-:
Already fixed, thanks!

PR #449, opened 2018-03-08
API for environment bindings

lionel-:
Introduces a new `env_binding_` prefix.

* New verbs `env_binding_lock()` and `env_binding_unlock()`. They change the locks and return the old value of `env_binding_are_locked()` invisibly.

* New predicates `env_binding_are_locked()`, `env_binding_are_active()` and `env_binding_are_promise()`. They return logical vectors named after their input.

All these functions accept multiple bindings. If no bindings are supplied they operate on the entire environment.

Also adds an internal `env_binding_type_sum()` that will be useful in `env_show()`.


lionel-:
Then we'll have `env_lock()` and `env_is_locked()`. And the internal function `env_unlock()`.

lionel-:
> Just needs lifecycle labelling

We've used the convention that the default lifecycle for functions is the package's one. Would you like to mark these new functions as experimental? Or do you want to explicitly label all maturing functions in the package as such?

hadley:
I think we should mark as experimental, just to be on the safe side.

PR #469, opened 2018-03-17
Use .Call instead of .Primitive(".Call")

krlmlr:
for the tilde thunk. With this patch:

``` r
library(rlang); eval_tidy(quo(`~`))
#> function (...) 
#> .Call(list(name = "rlang_tilde_eval", address = <pointer: 0x5613a31c3030>, 
#>     dll = list(name = "rlang", path = "/home/muelleki/R/x86_64-pc-linux-gnu-library/3.4/rlang/libs/rlang.so", 
#>         dynamicLookup = FALSE, handle = <pointer: 0x5613a4e5a120>, 
#>         info = <pointer: 0x5613a20f6650>), numParameters = 4L), 
#>     sys.call(), <environment>, <environment>, environment())
#> <environment: base>
```

> Created on 2018-03-17 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

### DIFF
--- a/R/c-api.R
+++ b/R/c-api.R
@@ -8,7 +8,7 @@ init_c_constants <- function() {
       formals(function(...) NULL),
 
     tilde_thunk_body =
-      expr((!!.Call)(!!(rlang_tilde_eval),
+      expr((!!expr(.Call))(!!(rlang_tilde_eval),
         sys.call(),
         "data_mask_arg",
         "data_mask_top_arg",


lionel-:
When we switch to thunk quosures they won't be pretty etiher. I like that quosures can be called from the empty environment even if that's purely out of principle.

krlmlr:
I don't see a difference between using `.Primitive` and `.Call` w.r.t. empty environments:

``` r
eval(.Primitive, emptyenv(), emptyenv())
#> function (name)  .Primitive(".Primitive")
eval(.Call, emptyenv(), emptyenv())
#> function (.NAME, ..., PACKAGE)  .Primitive(".Call")
```

> Created on 2018-03-17 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

But using `.Primitive(".Call")` inhibits joint profiling of R and native code.

lionel-:
Try with expressions rather than literals.

krlmlr:
Can you please rephrase?

lionel-:
In both `eval()` calls you're evaluating literals so they will return the same thing no matter the environment. That's the reason I'm embeddig `.call()`. Otoh `eval(quote(.Call), emptyenv())` will fail. Now we could embed a closure in CAR position if that helps with profiling. That will be a little less efficient though.

krlmlr:
Still indistinguishable, I can't access `.Primitive` from an empty environment either:

``` r
eval(quote(.Primitive), emptyenv(), emptyenv())
#> Error in eval(quote(.Primitive), emptyenv(), emptyenv()): object '.Primitive' not found
eval(quote(.Call), emptyenv(), emptyenv())
#> Error in eval(quote(.Call), emptyenv(), emptyenv()): object '.Call' not found
```

> Created on 2018-03-18 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

lionel-:
Note that the quosure doesn't embed `quote(.Primitive())`, it embeds the primitive literal... With your patch there's a symbol in CAR position so that makes the quosure dependent on context.

krlmlr:
I see what you mean: the `.Primitive` object is embedded and always accessible. I don't think it's that important, creating formulas shouldn't be relevant for profiling. Thanks anyway!

lionel-:
Btw no need to supply two environments to `eval()`, the third argument only matters if the second is not an environment.

PR #479, opened 2018-04-05
Calltraces

hadley:
This PR implements a new calltrace objects, along with an attractive tree-like print method (inlined from `cli::tree()`). It comes with convenient parameters for subsetting the tree to commonly useful fragments.


R/calltrace.R
```diff
@@ -0,0 +1,212 @@
...
+    class = "calltrace"
+  )
+}
+
+# Methods -----------------------------------------------------------------
+
+#' @export
+format.calltrace <- function(x, simplify = FALSE, dir = getwd(), ...) {
+  if (length(x) == 0) {
+    return("\u2588")
```
jimhester: Do you need a non-unicode fallback for this?

hadley: Oops, yes.

PR #500, opened 2018-05-03
Treat bare functions as exiting handlers

### LINKED ISSUE #499: with_handlers() should accept bare functions
And treat them as exiting handlers

lionel-:
So `with_handlers(expr, error = identity)` is equivalent to `tryCatch(expr, error = identity)`.

Closes #499

### DIFF
--- a/R/cnd-handlers.R
+++ b/R/cnd-handlers.R
@@ -31,10 +31,10 @@
 #' @param .expr An expression to execute in a context where new
 #'   handlers are established. The underscored version takes a quoted
 #'   expression or a quoted formula.
-#' @param ... Named handlers. Handlers should inherit from `exiting`
-#'   or `inplace`. See [exiting()] and [inplace()] for constructing
-#'   such handlers. Dots are evaluated with [explicit
-#'   splicing][tidy-dots].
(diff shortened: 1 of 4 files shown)


R/cnd-handlers.R
```diff
@@ -82,11 +85,9 @@ with_handlers <- function(.expr, ...) {
     abort("All handlers should be functions")
   }
 
-  inplace <- keep(handlers, inherits, "inplace")
-  exiting <- keep(handlers, inherits, "exiting")
-  if (length(handlers) > length(exiting) + length(inplace)) {
-    abort("all handlers should inherit from `exiting` or `inplace`")
-  }
+  is_inplace <- map_lgl(handlers, inherits, "inplace")
```
hadley: Should we now use `as_function()` instead of `is_function()`? That way you can use the formula shortcut for exiting handlers.

lionel-: I wouldn't mind, but then should we deprecate `exiting()` since it would offer no added value (beside being extra precise)?

hadley: I think we could leave it for symmetry with `inplace()`

PR #501, opened 2018-05-03
Take metadata through ... in condition signallers

lionel-:


R/cnd.R
```diff
@@ -279,65 +279,135 @@ muffle <- function(...) NULL
...
+#' [base::warning()] and [base::message()], but make it easy to supply
+#' condition metadata:
+#'
+#' * Supply `type` to create a classed condition. Typed conditions can
+#'   be captured or handled selectively, allowing for finer-grained
+#'   error handling.
+#'
+#' * Supply metadata with named `...` arguments. This data will be
+#'   stored in the condition object and can be examined by handlers,
+#'   print methods, etc.
```
hadley: I think _print_ methods is a little misleading here, maybe just refer to methods generally?

R/cnd.R
```diff
@@ -279,65 +279,135 @@ muffle <- function(...) NULL
...
+#'     # Compute an alternative return value with the data:
+#'     recover_error(err$data)
+#'   }
+#' )
+#'
+#' }
+abort <- function(.msg, .type = NULL, ..., .call = FALSE, msg, type, call) {
+  validate_signal_args(msg, type, call)
+
+  cnd <- error_cnd(.type, ..., .msg = .msg, .call = cnd_call(.call))
```
hadley: This surprises me

R/cnd.R
```diff
@@ -279,65 +279,135 @@ muffle <- function(...) NULL
...
+  if (!is_false(.call)) {
```
hadley: We really should always be assigning something to the `call` component to respect the fields of the parent class.

tests/testthat/helper-cnd.R
```diff
@@ -41,3 +41,56 @@ with_non_verbose_retirement <- function(expr) {
...
+  })
+
+  wngs
+}
+catch_warning_msgs <- function(expr) {
+  wngs <- catch_wngs(expr)
+  flatten_chr(pluck(wngs, "message"))
+}
+
+catch_cnds <- function(expr) {
```
hadley: This should probably go into testthat eventually.

R/cnd.R
lionel-: This would cause the call to be printed on error. Am I missing something?

R/cnd.R
lionel-: What do you mean?

R/cnd.R
hadley: We'd assign `NULL` when `.call` is `FALSE`

R/cnd.R
hadley: Now I don't remember 😭

R/cnd.R
lionel-: Good point, I'll change this separately the ctors need a little work.

PR #505, opened 2018-05-06
Streamline condition ctors and signallers

lionel-:
* Constructors and signallers now take `message` argument instead of `.msg`. This reflects the `message` field `message`, `warning` and `error` conditions. I'm still wondering whether bare conditions should have a `message` field since they don't display any message. The message field can serve as self-documentation for the condition though.

* `cnd_inform()`, `cnd_warn()` and `cnd_abort()` are retired and defunct (I don't think there is anyone using these functions). `cnd_signal()` now dispatches on the relevant signaller based on condition type.

* New `signal()` function to complete the family of `abort()` etc.

* Bare conditions are now always signalled with a muffle restart. This way we get consistency with messages and warnings and we get rid of the `.mufflable` argument. All non-critical conditions signalled with rlang can be muffled.


R/lifecycle-retired.R
```diff
@@ -424,31 +424,31 @@ new_cnd <- function(.type = NULL, ..., .msg = NULL) {
   # Deprecated in 0.1.2
   warning("`new_cnd()` has been renamed to `cnd()` for consistency",
     call. = FALSE)
-  cnd(.type = .type, ..., .msg = .msg)
+  cnd(.type = .type, ..., message = .msg)
 }
 #' @rdname deprecated-cnd
 #' @export
 cnd_error <- function(.type = NULL, ..., .msg = NULL) {
   # Deprecated in 0.1.2
```
hadley: Should we make these defunct now?

PR #509, opened 2018-05-07
Replace rst_muffle() with cnd_muffle()

lionel-:
Includes #506

* No longer generic

* Supports interrupts though it's not necessarily safe to try and muffle an interrupt as there might not be a relevant restart. Maybe we need `cnd_maybe_muffle()` that would not throw an error if a restart is not found?

* `cnd_muffle()` is marked as a calling handler so can be used directly:

```r
fn <- function() {
  warn("msg", "mypkg_mywarning_wng")
  "foo"
)

with_handlers(fn(), mypkg_mywarning_wng = cnd_muffle)
```


NEWS.md
```diff
@@ -1,6 +1,12 @@
 
 # rlang 0.2.0.9000
 
+* The experimental function `rst_muffle()` has been renamed to
+  `cnd_muffle()` and is now defunct. Unlike its predecessor
```
hadley: It's slightly unclear that you're making `rst_muffle` defunct, vs `cnd_muffle`

R/cnd.R
```diff
@@ -435,6 +435,87 @@ validate_signal_args <- function(msg, type, call, env = parent.frame()) {
...
+#' @section Mufflable conditions:
+#'
+#' Not all conditions are signalled with a muffling restart.
+#' `cnd_muffle()` is compatible with:
+#'
+#' * `warning` and `message` conditions. In this case `cnd_muffle()`
+#'   is equivalent to [base::suppressMessages()] and
+#'   [base::suppressWarnings()].
+#'
+#' * Bare conditions signalled with `signal()` or [cnd_signal()].
```
hadley: Note that signalling with `singalCondition()` does _not_ create a muffable condition.

R/cnd.R
```diff
@@ -435,6 +435,87 @@ validate_signal_args <- function(msg, type, call, env = parent.frame()) {
...
+#'
+#' * Interrupts are sometimes signalled with a `resume` restart on
```
hadley: It's a bit unclear how this connects to muffable restarts.

R/cnd.R
```diff
@@ -435,6 +435,87 @@ validate_signal_args <- function(msg, type, call, env = parent.frame()) {
...
+#' Unlike [exiting()] handlers, [calling()] handlers must be explicit
+#' that they have handled a condition to stop it from propagating to
+#' other handlers. Use `cnd_muffle()` within a calling handler (or as
+#' a calling handler, see examples) to prevent any other handlers from
+#' being called for that condition.
+#'
+#'
+#' @section Mufflable conditions:
+#'
+#' Not all conditions are signalled with a muffling restart.
```
hadley: I think this needs a brief lead-in like: Most conditions signalled by base are muffable, although the name of the restart varies. `cnd_muffle()` will automatically call the correct restart for ...

R/cnd.R
```diff
@@ -435,6 +435,87 @@ validate_signal_args <- function(msg, type, call, env = parent.frame()) {
...
+#' * Interrupts are sometimes signalled with a `resume` restart on
+#'   recent R versions.
+#'
+#' If you call `cnd_muffle()` with a condition that is not mufflable
+#' you will cause a new error to be signalled.
+#'
+#' * Errors are not mufflable since they are signalled in critical
+#'   situations where execution cannot continue safely.
+#'
+#' * Conditions captured with [base::tryCatch()], [with_handlers()] or
```
hadley: I don't understand this

R/cnd.R
```diff
@@ -435,6 +435,87 @@ validate_signal_args <- function(msg, type, call, env = parent.frame()) {
...
+#'   my_particular_msg = calling(function(cnd) {
+#'     inform("Dealt with this particular message")
+#'     cnd_muffle(cnd)
+#'   })
+#' )
+#'
+#' # Note how execution of `fn()` continued normally after dealing
+#' # with that particular message.
+#'
+#' # Since cnd_muffle() is a calling handler itself, it can also be
```
hadley: I think this is sufficiently esoteric that you shouldn't put in an example?

R/cnd.R
```diff
@@ -435,6 +435,87 @@ validate_signal_args <- function(msg, type, call, env = parent.frame()) {
...
+#' # passed to with_handlers() directly:
+#' with_handlers(fn(),
+#'   my_particular_msg = cnd_muffle
+#' )
+cnd_muffle <- NULL
+
+# Delay definition of `cnd_muffle()` after load-time because
+# `calling()` uses internal rlang functions
+delayedAssign("cnd_muffle", {
```
hadley: I'm not sure the advantage of making a `cnd_muffle` a calling handler is sufficiently great to make this sort of code gymnastic worthwhile.

R/cnd.R
hadley: I still don't understand. Can you give an example?

lionel-: ok jumping to a wrong restart is harder to trigger than I thought. It cannot happen when using only our signalling functions, the user would have to set up a muffle restart manually:

```r
impl <- function() {
  signal("", "my_cnd")
  "expected return value"
}

fn_correct <- function() {
  with_handlers(impl(), my_cnd = cnd_muffle)
}
fn_wrong <- function() {
  with_handlers(impl(), my_cnd = ~cnd_muffle(.x))
}

with_restarts(muffle = function() "wrong return value",
  fn_correct()
)
#> [1] "expected return value"

with_restarts(muffle = function() "wrong return value",
  fn_wrong()
)
#> [1] "wrong return value"
```

Maybe we could prefix the restart (e.g. `rlang_muffle`) to make this event even less likely.


On the other hand getting a "no restart" error is easy, especially with the formula shortcut interpreted as exiting by default:

```r
# This is fine:
with_handlers(warn(""), warning = cnd_muffle)

# These cause errors:
with_handlers(warn(""), warning = ~cnd_muffle(.x))
#> Error in invokeRestart("muffleWarning") :
#>   no 'restart' 'muffleWarning' found

tryCatch(warning(""), warning = cnd_muffle)
#> Error in invokeRestart("muffleWarning") :
#>   no 'restart' 'muffleWarning' found
```

Maybe we shouldn't use `as_function()` in `with_handlers()` and force users to be explicit with `exiting()` when they want to use formulas

hadley: I think that current behaviour is ok - you can't muffle in an exiting handler.

lionel-: Sounds good. Is this PR now mergeable?

PR #511, opened 2018-05-08
Fix calltrace

lionel-:
With this fix the srcrefs are grabbed from the sys.calls. This way the srcrefs reference the location of the call rather than the location of the definition of the function being called.

I toyed with using the srcrefs to print the calls themselves, which would allow us to get more accurate calls in case primitive functions are involved (we could observe `x <- i()` instead of just `i()`). However I found some issues when function bodies are not wrapped in `{`. I took note of this problem in my backlog of R core issues to fix.

I find there's something odd about `calltrace()`. I feel it lacks an underscore, however I don't want to use `call_` as a prefix here. How about `backtrace()` or `trace_back()`?

### DIFF
--- a/R/calltrace.R
+++ b/R/calltrace.R
@@ -148,7 +148,7 @@ trace_as_tree <- function(x, dir = getwd()) {
   nodes <- c(0, seq_along(x$calls))
   children <- map(nodes, function(id) seq_along(x$parents)[x$parents == id])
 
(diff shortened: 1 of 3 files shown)


hadley:
`trace_back()` seems the best of those options, but I don't have a problem with `calltrace()`

lionel-:
ok let's go with `trace_back()` then. I like it because it's a verb and it introduces a natural `trace_` prefix for trace objects.

PR #512, opened 2018-05-08
Rename calltrace

lionel-:
* `calltrace()` => `trace_back()`
* `new_calltrace()` => `new_trace()`
* `calltrace` class => `rlang_trace`


hadley:
I think it would be good to get @jimhester's opinion on this.

jimhester:
The renaming overall seems like a good idea to me. The only hesitation I see with naming the argument  `to:` is it enforces a specific direction when reading the output. I sometimes read the tracebacks top down and sometimes bottom up, depending on what I am interested in.

But the point is minor, so I am fine changing the argument name as well, it does read nicely.

lionel-:
Thanks! I'm not sure the `to` argument will constrain how we read the output since it's more an argument for the creation of the backtrace.

PR #517, opened 2018-05-10
New format methods for backtraces

lionel-:
* `simplify = "branch"` is equivalent to the old `simplify = TRUE` argument. The tree indentation is suppressed in that case. This method will be used in error messages:

  ```
  f <- function() eval(quote(eval(quote(g()))))
  g <- function() tryCatch(eval(quote(h())), foo = identity, bar = identity)
  h <- function() abort("Oh no!")

  eval(quote(f()))
  #> Error: Oh no!
  #> Backtrace:
  #> █
  #> ├─f()
  #> ├─g()
  #> └─h()
  ```

With `simplify = "collapse"`, a collapsed version of the tree is shown. This would be used in the print method of rlang errors:

  ```
  catch_cnd(eval(quote(f())))
  #> <error>
  #> * Message: "Oh no!
  #> * Class: `rlang_error`
  #> * Backtrace:
  #> █
  #> ├─[ catch_cnd(...) ] ... +5
  #> └─f()
  #>   ├─[ eval(...) ] ... +1
  #>   ├─[ eval(...) ] ... +1
  #>   └─g()
  #>     ├─[ tryCatch(...) ] ... +6
  #>     ├─[ eval(...) ] ... +1
  #>     └─h()
  ```


lionel-:
Also adds a couple of global options and corresponding arguments:

* `rlang_trace_format_srcrefs` to disable srcrefs
* `rlang_trace_top_env` to set top environment of backtraces

lionel-:
@hadley Should `expect_known_output()` have an option to send a non-failing message when output differs? This way we can monitor changes in output without causing CRAN failures when dependencies evolve (here backtrace output with `eval()` differs on old R versions).

hadley:
@lionel- wouldn't it be easier to just skip on other versions of R?

tests/testthat/test-trace-recursive.txt
```diff
@@ -0,0 +1,16 @@
...
+└─f()
+  └─g()
+
+Collapsed:
+█
+├─[ eval_tidy(...) ]
+└─f()
+  └─g()
+
+Branch:
```
hadley: I think this view should not use the tree styling (since it's very far from the tree now)

lionel-:
> wouldn't it be easier to just skip on other versions of R?

Yes but I'm worried about future updates. In general it seems checking output is more fragile than specific value checking.

tests/testthat/test-trace-recursive.txt
lionel-: I like that it reminds the tree view while being sufficiently different and I find it effective at separating the backtrace from the rest of the output. If these backtraces are going to be everywhere they should be easily detectable by eye so you can ignore it and look for the error message straight away. Maybe that's possible to achieve with another kind of output? I think the branch view makes sense though.

hadley: What if they used `-` without the `|`?

PR #529, opened 2018-05-17
Added link in documentation

osorensen:


R/env.R
```diff
@@ -70,7 +70,7 @@
 #'   features.
 #' @param .parent,parent A parent environment. Can be an object
 #'   supported by [as_environment()].
-#' @seealso `scoped_env`, [env_has()], [env_bind()].
```
lionel-: I think the link to `scoped_env` should be removed altogether as the API is uncertain.

lionel-:
Fixed it from outside the PR. Thanks for all the PRs!

osorensen:
Great! And thanks for a very useful package; it makes all the quote() and eval() stuff a lot more comprehensible.

PR #533, opened 2018-05-23
Add experimental support for `.Last.error`

lionel-:
When rlang is loaded `.Last.error` is installed as a delayed binding that offers to enable error capture. If "no" is selected a taskCallback resets the active binding (Luke mentioned on r-devel that it might make sense to provide an official way of resetting delayed bindings but that was not implemented yet AFAICT --- Also an active binding wouldn't work here because top-level handlers have to be installed by top-level calls).

```r
.Last.error
#> Would you like to enable `.Last.error` capture?
#>
#> 1: Yes
#> 2: No

#> Selection: 1
#> Capture of uncaught errors in `.Last.error` is now enabled
```

Once enabled, uncaught errors are recorded in `.Last.error`:

```r
.Last.error
#> NULL

stop("bim")
#> Error: bim

.Last.error
#> <simpleError: bim>

f <- function() stop("bam")
f()
#> Error in f(): bam

.Last.error
#> <simpleError in f(): bam>
```

However errors thrown at top-level reset the calling handler stack, so users have to restore the stack manually. This requires two steps because we can only add top-level handlers from a top-level call such as triggered by a delayed binding.

```r

data[1]
#> Recording of errors is now disabled, please call `rlang:::enable_last_error()`
#> Error in data[1] : object of type 'closure' is not subsettable

.Last.error
#> <simpleError in data[1]: object of type 'closure' is not subsettable>

rlang:::enable_last_error()
#> Now type `.Last.error` at the console to enable error recording

.Last.error
#> Capture of uncaught errors in `.Last.error` is now enabled
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,12 @@
 
 # rlang 0.2.0.9000
 
+* Uncaught error objects can now be recorded in `.Last.error` just
+  like `.Last.value` references the last value returned at top-level.
+  This is experimental and currently fails with errors thrown at top
+  level (i.e. from a primitive function called directly at the
+  console).
+
 * The experimental function `rst_muffle()` is now defunct. Please use
   `cnd_muffle()` instead. Unlike its predecessor `cnd_muffle()` is not
   generic. It is marked as a calling handler and thus can be passed
--- a/R/cnd.R
+++ b/R/cnd.R
@@ -452,12 +452,17 @@ catch_cnd <- function(expr) {
 # This returns an expression that has to be evaluated at top-level
 # because the only way to add a top-level handler is from a delayed
 # binding
-enable_last_error_expr <- function() {
+enable_last_error_expr <- function(ask = TRUE) {
   expr({
-    sel <- utils::menu(
-      c("Yes", "No"),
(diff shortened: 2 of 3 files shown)


gaborcsardi:
Interesting! It is not optimal that top level errors turn it off, but I guess we cannot do much about that.

Did you consider the alternative of having some helper functions for (re)defining `options(error)`? What  are the pros and the cons?

gaborcsardi:
Isn't it true that there will be a `stop()` call on the stack when you inspect it from the `options(error)`  callback?

lionel-:
Oh hmm yes good point it seems like we should be able to get or create a condition object from there.

If we can't make the calling handler work by poking into `RCNTXT R_Toplevel` then perhaps the `options(error)` way would offer a better user experience. You'd have to restore the handler each time you switch to `recovery` but that's still better than having the stack reset at each top-level error.

gaborcsardi:
I think we could provide some api (i.e. a helper function) to call `recover()` or whatever, after we saved the error.

hadley:
I think experimental stuff like this needs to go somewhere other than rlang, and we need to work with Luke to figure out a path going forward.

PR #541, opened 2018-05-29
Add parameter to truncate backtrails

### LINKED ISSUE #489: Display truncated terminal stack
Probably want to automatically truncate long callstacks in error messages. Best to truncate in middle so that see the start and the end.

lionel-:
Closes #489

Displays number of truncated frames in the same style as collapsed frames: `+n`.

```
Full:
 ─f(10)
 ─f(n - 1)
 ─f(n - 1)
 ─f(n - 1)
 ─f(n - 1)
 ─f(n - 1)
 ─f(n - 1)
 ─f(n - 1)
 ─f(n - 1)
 ─f(n - 1)
 ─f(n - 1)

Truncated:
 ─f(10)
 ─f(n - 1)
 ─f(n - 1)
 .
 . +6
 .
 ─f(n - 1)
 ─f(n - 1)
 ─f(n - 1)

Corner case:
 ─f(10)
 .
 . +10
```


lionel-:
I think it's not very clear that the backtrail was truncated with this approach. Do you think it's important to be consistent with collapsed frames in this case?

```
├─[ catch_cnd(...) ] ... +5
└─f()
  └─g()
    └─h()
```

Maybe it should be:

```
 ─f(10)
 ─f(n - 1)
 ─f(n - 1)
 .
 . ... +6
 .
 ─f(n - 1)
 ─f(n - 1)
 ─f(n - 1)
```

I think I prefer the original version though.

hadley:
I don't think it needs to be consistent. I think the `...` is too vague - why not spell out what's happening with words? Also, I'd prefer to have two more calls than two lines that contain only `.`.

hadley:
I like 

```
 ─f(10)
 ─[ eval(...) ] ... +1
 ─f(n - 1)
 ─[ ... +6 calls ]
 ─f(n - 1)
 ─f(n - 1)
```

PR #542, opened 2018-05-29
Edit docs for typos & grammar

pkq:
This PR includes many small edits to the documentation to improve clarity. Edits include removing extraneous words, making pluralization consistent, standardizing comment line break length, and replacing most instances of 'since' with the more grammatically correct 'because'.


DESCRIPTION
```diff
@@ -21,7 +21,7 @@ Suggests:
     pillar,
     testthat (>= 2.0.0),
     covr
-RoxygenNote: 6.0.1
+RoxygenNote: 6.0.1.9000
```
lionel-: Could you downgrade to release roxygen and redocument please.

R/cnd.R
```diff
@@ -378,7 +379,7 @@ interrupt <- function() {
 #' If you call `cnd_muffle()` with a condition that is not mufflable
 #' you will cause a new error to be signalled.
 #'
-#' * Errors are not mufflable since they are signalled in critical
```
lionel-: After reading https://dictionary.cambridge.org/grammar/british-grammar/as-because-or-since it is not clear to me these changes are desired.

pkq: Agreed. In a few cases (such as this one), I think adding a comma after the since- clause helps emphasize that.

pkq:
Thanks for taking a look at this. As requested, I downgraded to the released version of `roxygen`, removed white space changes, and reverted the because/since changes. I agree that, in most cases, emphasizing the result clause is the better choice. Let me know if there are any other changes that would be helpful.

R/cnd.R
lionel-: I don't like this comma :(

batpigandme: While the top one makes sense to me, the bottom comma (though technically correct/stylistic) seems to break things up awkwardly, given that it's a bullet.

pkq: This comma is no more. I also rebased, to include the typo fix from #544. Please let me know if there's anything else you'd like changed.

lionel-: Fewer commas throughout please. Only add a comma if you're absolutely sure it is an improvement that everyone will agree with. In general this advice should hold for all copy-editing changes, i.e. things that are not typos, clear grammatical errors, or that do not clearly improve upon the clarity of the doc.

Also in the future it is better to create multiple PRs that contain only similar changes. This way we can move forward faster. Thanks for bearing with us!

pkq: Thanks for the quick reply. Sounds good; I'll review and revert the comma changes. Quick question about multiple PRs: Should they build on each other (i.e., each as a separate branch that builds on the last), or each a new branch based on master? Thanks!

lionel-: If they are independent it is better to base them on master. If they are dependent it is ok to have them based on one another.

pkq:
I've trimmed this PR down to just minor grammatically edits, including missing and extraneous words, and pluralization and tense inconsistencies. I'll open a separate PR for a pared down set of phrasing suggestions. Thanks again for considering this.

lionel-:
Thanks! I'll take a look at the other PR as well.

PR #547, opened 2018-06-07
Fix typo in docs.

mkcor:
Hello!

There's also "quasiquotation lets you build a complex expressions" in `R/quotation.R` to fix. But I'm hesitating between "... build a complex expression" and "... build complex expressions" (either would work).

Best,
Marianne


lionel-:
Oops this was already fixed in https://github.com/r-lib/rlang/pull/559
Thanks anyway!

PR #550, opened 2018-06-28
Edit examples in new_data_mask

jrnold:
I was attempting to use new_data_mask and found it difficult to figure out exactly what it was doing from the docs. I added an example that walks through creating a multi-environment data mask with pronouns from scratch.


lionel-:
Hi @jrnold, thanks for this!

I have tweaked the examples a bit, does this look good to you?

lionel-:
Merging now but let me now if you have further comments.

jrnold:
Thanks. That looked great!

PR #568, opened 2018-08-13
Evaluate promises & active bindings in env_get()

### LINKED ISSUE #554: env_print incorrectly labels binding as promise
``` r
library(rlang)

power1 <- function(exponent) {
  function(x) {
    x ^ exponent
  }
}

square <- power1(2)
square(2)
#> [1] 4

env <- get_env(square)
env_print(env)
#> <environment: 0x7ff5c3cf7988>
#>   parent: <environment: global>
#>   bindings:
#>    * exponent: <promise>

env$exponent
#> [1] 2
env_print(env)
#> <environment: 0x7ff5c3cf7988>
#>   parent: <environment: global>
#>   bindings:
#>    * exponent: <promise>
```

Created on 2018-07-03 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

hadley:
Fixes #554

### DIFF
--- a/src/internal/env-binding.c
+++ b/src/internal/env-binding.c
@@ -40,7 +40,7 @@ sexp* rlang_env_get(sexp* env, sexp* nm) {
     r_abort("Internal error: `rlang_env_get()` should have failed earlier");
   }
 
-  if (r_typeof(out) == r_type_promise || r_typeof(out) == r_type_function) {
+  if (r_typeof(out) == r_type_promise) {
     out = r_eval(out, r_empty_env);
   }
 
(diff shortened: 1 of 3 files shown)


src/internal/env-binding.c
```diff
@@ -40,5 +40,9 @@ sexp* rlang_env_get(sexp* env, sexp* nm) {
     r_abort("Internal error: `rlang_env_get()` should have failed earlier");
   }
 
+  if (r_typeof(out) == r_type_promise || r_typeof(out) == r_type_function) {
```
lionel-: Won't the second alternative evaluate bindings to a function? I think active bindings have promise type.

hadley: Oooh good point; looks like I should be using this: https://github.com/wch/r-source/blob/496adb0085703362b2dee63537f0d89820f5afbb/src/include/Defn.h#L430

hadley: But you're right that it is a subclass of promise

hadley: Fixed now

PR #576, opened 2018-08-17
Implement exec()

### LINKED ISSUE #536: Replace invoke() with exec()
```
library(rlang)

sym_protect <- function(x) {
  if (is_symbolic(x)) {
    call2(quote, x)
  } else {
    x
  }
}

exec <- function(f, ..., .env = caller_env()) {
  if (is.character(f)) {
    f <- sym(f)
  }

  args <- rlang::list2(...)
  args[-1] <- lapply(args[-1], sym_protect)

  call <- as.call(args)
  eval_bare(call, .env)
}
```

This is less flexible than `do.call()` but the flexibility only matters when you really care about the generated calls. In that case, you are best off generating (with `expr()` or similar) and evaluating yourself.

`invoke()` needs to be soft-deprecated/retired.

hadley:
Fixes #536


R/exec.R
```diff
@@ -0,0 +1,48 @@
...
+#'
+#' * To call every function stored in a list (in conjunction with `map()`/
+#'   [lapply()])
+#'
+#' @param f A function, or function name as a string.
+#' @param ... Arguments to function.
+#'
+#'   These dots support [tidy-dots] features.
+#' @param  .env Environment in which to evaluate the call. This will be
+#'   most useful if `f` is a string, or the function as side-effects.
```
lionel-: has side effects.

R/exec.R
```diff
@@ -0,0 +1,48 @@
...
+#'
+#' fs <- list(a = function() "a", b = function() "b")
+#' lapply(fs, exec)
+#'
+#' # Compare to do.call it will not automatically inline expressions
+#' # into the evaluated call.
+#' x <- 10
+#' args <- exprs(x1 = x + 1, x2 = x * 2)
+#' exec(list, !!!args)
+#' do.call(list, args)
```
lionel-: Does it need an example to show how supplying a string creates better backtraces / recorded calls?

And then show how to construct and evaluate a call manually when it's important to mask the arguments as well. Maybe taking `lm()` as example:

```r
f <- disp ~ cyl

# Horrible call
exec("lm", f, data = mtcars)

# Nice call
eval(expr(lm(!!f, data)), list(data = mtcars))
```

hadley:
Rebased against master and improvement documentation.

lionel-:
Thanks!

Issue #563, opened 2018-07-27
rlang:::new_lines makes illegal pairlist, not terminated by NULL


BillDunlap:
rlang:::new_lines maintains a linked list, indent_status, whose elements are made with new_node(FALSE, 0L).  The second argument to new_node (eventually to Rf_cons) should be a linked list (pairlist), not 0L.  I noticed this because TERR complains about the illegal second argument to Rf_cons, but when I used R to trace the calls to new_node, with trace(rlang::new_node, quote({str(car),str(cdr)})), R crashed.  E.g.,
  > library(rlang)
  > str(new_node(FALSE, 0L))  
  
   *** caught segfault ***
  address 0x21006009, cause 'memory not mapped'

I believe the following patch will keep things legal without changing the logic of your code much.  All the testthat tests pass after the change.

```
--- /homes/bill/CRAN/rlang/R/deparse.R  2018-02-14 23:32:56.000000000 -0800
+++ ./rlang/R/deparse.R 2018-07-27 12:03:28.463628000 -0700
@@ -159,10 +159,10 @@
       status <- node_car(self$indent_status)

       if (self$next_indent_sticky) {
-        node_poke_cdr(status, inc(node_cdr(status)))
+        node_poke_cadr(status, inc(node_cadr(status)))
       } else {
         self$indent <- self$indent + 2L
-        self$indent_status <- new_node(new_node(FALSE, 0L), self$indent_status)
+        self$indent_status <- new_node(pairlist(FALSE, 0L), self$indent_status)
         self$next_indent_sticky <- TRUE
       }

@@ -176,7 +176,7 @@
       }

       reset <- node_car(status)
-      n_sticky <- node_cdr(status)
+      n_sticky <- node_cadr(status)

       # Decrease indent level only once for all the openers that were
       # on a single line
@@ -187,7 +187,7 @@
       }

       if (n_sticky >= 1L) {
-        node_poke_cdr(status, dec(n_sticky))
+        node_poke_cadr(status, dec(n_sticky))
       } else {
         self$indent_status <- node_cdr(self$indent_status)
         self$next_indent_sticky <- FALSE


```

lionel-:
This is an experiment in using cons cells as data structures. These cells do crash with `str()` but they have basic `print()` support:

```r
print(rlang::new_node(1, 2))
#> [[1]]
#> [1] 1
#> 
#> . 
#> 
#> [1] 2
```

[...]

PR #579, opened 2018-08-20
Fix handling of indent_status in deparse()

akbertram:
The increase_indent() function was actually creating an invalid
object via the expression `new_node(FALSE, 0L)`.

The `new_node` function is a wrapper for the `cons()` function, which
expects two arguments:

* car: the value of the node to be created
* tail (or cdr): a pairlist (or NULL) which continues the list after the  new node created

The expression `new_node(FALSE, 0L)` then actually passes an integer vector
as the cdr or tail instead of a pairlist, creating an illegal object.

You can verify this by evaluating the following expressions:

     x <- new_node(FALSE, 0L)
     length(x[[1]])

Which will crash your R session.

The code does work at the moment because GNU R does not actually check
that the cdr argument to the cons() function is valid, and the value is
only accessed via the CDR macros, which also do not verify the expression.

Other interpreters (e.g. Renjin) are not so forgiving however.

### DIFF
--- a/R/deparse.R
+++ b/R/deparse.R
@@ -159,10 +159,10 @@ new_lines <- function(width = peek_option("width"),
       status <- node_car(self$indent_status)
 
       if (self$next_indent_sticky) {
-        node_poke_cdr(status, inc(node_cdr(status)))
+        node_poke_cadr(status, inc(node_cadr(status)))
       } else {
         self$indent <- self$indent + 2L
-        self$indent_status <- new_node(new_node(FALSE, 0L), self$indent_status)
+        self$indent_status <- new_node(new_node(FALSE, new_node(0L, NULL)), self$indent_status)
         self$next_indent_sticky <- TRUE
       }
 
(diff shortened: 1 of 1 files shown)


lionel-:
Discussion in https://github.com/r-lib/rlang/issues/563

lionel-:
Thanks!

Merged because it's causing trouble to non GNU implementations, but I disagree that non-list cells are invalid objects. Sure as long as they make R crash they shouldn't be exposed to the user, but this was used internally only.

In my opinion the proper place for a runtime check is in the cons cell getters and setters.

akbertram:
Thanks @lionel- for merging the fix.

It's hard to say what is "valid" or "invalid" without a language specification, in the absence of which we can only define validity as what runs without crashing the current version of GNU R :-)

But I have to say that my impression was that the definition of a pairlist is pretty close to holy. It's one of the few basic R data types, and there are hundreds of thousands of lines of code that rely on the type of the CDR pointer.

Maintaining these preconditions is more than just runtime checks- it's about being able to reason about the language. For example, Renjin's Just-in-time compiler uses the definition of a pairlist to reason about pairlists and operations on pairlists in order to infer types of expressions within a loop body or an sapply function.

Our compiler is just one example -- exploiting loop holes in the current implementation make future innovation across the whole R ecosystem more difficult!

PR #582, opened 2018-08-22
Add parentheses to call CAR when printing if required (#581)

### LINKED ISSUE #581: Printing quosures with calls sometimes lacks required parentheses
Printing quosures does not always have parentheses around the CAR of a call when it's necessary. For example, if an anonymous function expression is unquoted into the CAR slot of a call, the required parentheses are missing and produce a strange print result for the expression:

``` r
f <- quote(function(x) x + 1)
rlang::quo((!!f)(x))
#> <quosure>
#>   expr: ^function(x) x + 1(x)
#>   env:  global
```

Extracting the expression shows that it's only the printing that's incorrect:

``` r
q <- rlang::quo((!!f)(x))
rlang::quo_get_expr(q) # correct expression
#> (function(x) x + 1)(x)
```

Another example with the same issue:

``` r
g <- quote(f + g)
rlang::quo((!!g)(x))
#> <quosure>
#>   expr: ^f + g(x)
#>   env:  global
as.call(c(g, quote(x))) # expected
#> (f + g)(x)
```

I think this could be fixed by adding a check to `call_deparse()` before deparsing the CAR to see if parentheses are needed, and insert them if so.

https://github.com/r-lib/rlang/blob/2cf70c8f41513831a02878384421e417169719a8/R/deparse.R#L432-L435

I have a local branch implementing this, and could submit a PR if that sounds OK.

mikmart:
Fixes #581 by adding a check to `call_deparse()` to see if the expression in the CAR slot of the call should be wrapped in parentheses for printing.

I picked the expressions that should have parentheses using the switch statement from `op_deparse()` and eyeballing which ops should have parentheses and which not.

### DIFF
--- a/R/deparse.R
+++ b/R/deparse.R
@@ -430,8 +430,67 @@ args_deparse <- function(x, lines = new_lines()) {
   lines$get_lines()
 }
 call_deparse <- function(x, lines = new_lines()) {
-  lines$deparse(node_car(x))
+  car <- node_car(x)
+  if (car_needs_parens(car)) {
+    car <- call("(", car)
+  }
+  lines$deparse(car)
   args_deparse(node_cdr(x), lines)
+}
+
+car_needs_parens <- function(x) {
+  if (typeof(x) != "language") {
+    return(FALSE)
+  }
+
+  op <- which_operator(x)
+  switch (op,
+    `function` = ,
+    `while` = ,
+    `for` = ,
+    `repeat` = ,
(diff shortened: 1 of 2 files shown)


lionel-:
Thanks, that was a good approach! I changed it a bit to use prefix notation when possible because it's better to have a lossless deparse-parse roundtrip. For functions that's not possible because the prefix notation doesn't work well.

PR #591, opened 2018-08-28
Implement bang()

lionel-:


jimhester:
I think we should consider a name which would give you an idea of what the function does.

Even having the knowledge that `!` can be pronounced bang and that `!!` as used in rlang is bang bang it gives no insight at what this function does (I certainly could not guess it). The only way to determine this is by looking at the documentation.

I think one of the key things we should strive for is to make things as self-describing as possible. Ideally someone would be able to roughly inuit what a piece of code is doing even having no prior knowledge of the packages used.

I unfortunately don't have a good alternative name to suggest.

lionel-:
Yup that function has been on hold for 6 months because of the naming issue. Self-description is of course good, but has to be balanced with concision. My first suggestion was `with_qq()` because it has the semantics of a `with_` function. Maybe `with_bang()` would be good.

How concise it should be (i.e. do we need a 4-5 letters 1 particle name) depends on the place this function will take in the toolkit. It's currently unclear:

1.  It's useful as a quoting replacement for do.call / exec. But is it confusing to have both exec and bang? `exec()` can be used as a regular function. It also has slightly different semantics with symbolic arguments:

    ```r
    x <- exprs(foo, bar)
    exec("list", !!!x)
    #> [[1]]
    #> foo
    #>
    #> [[2]]
    #> bar

    bang(list(!!!x))
    #> Error in eval_bare(enexpr(expr), caller_env()): object 'foo' not found
    ```

    Because it evaluates symbolic objects, `bang()` will also be useful for building complex calls with symbolic masking, so you don't end up with data inlined:

    ```r
    f <- disp ~ cyl
    data <- mtcars
    args <- list(weights = NULL, na.action = NULL)

    bang(lm(!!f, data, !!!args))
    #> Call:
    #> lm(formula = disp ~ cyl, data = data, weights = NULL, na.action = NULL)
    #>
    #> Coefficients:
    #> (Intercept)          cyl
    #>      -156.6         62.6
    ```

    I should probably add that example in the doc.


2.  Building functions. `expr()` is not appriate here because the function needs to be evaluated to record the environment. That's a rather subtle difference that will be hard to teach about in a natural and simple way.

    One major downside of building functions with quasiquotation is that the srcrefs get in the way: you can't see the result of `!!` and `!!!`. This will become more troublesome if R enables package srcrefs by default, which Tomas Kalibera is looking into (there's a performance issue). This will impact the function factories in purrr. Possible solutions are modifying the srcrefs or just strip them.


3.  Building fomulas. The issue with formulas is that `bang()` might not be very helpful because the reducing has to be done manually. I've been thinking about a `f()` constructor that would automatically do the reducing, with `+` as default:

    ```r
    x <- exprs(a, b, c)
    f(~!!!x)
    #> ~a + b + c

    f(out ~ !!!x)
    #> out ~ a + b + c

    f(out ~ !!!x, op = "*")
    #> out ~ a * b * c
    ```

    I have implemented a draft but it isn't quite right as users will probably expect the reducing to happen deep in the call tree:

    ```{r}
    f(out ~ 1 + 2 + !!!x)
    #> Error: Can't use `!!!` at top level
    ```

    To implement this we'd need to make the qq engine in rlang more flexible.

Since it isn't quite clear yet how important will `bang()` be, maybe we should just mark it as experimental for now until we figure it out?

hadley:
`exec()` is `call2()` is as this function is to `expr()`.

lionel-:
I strongly feel `eval` functions have to evaluate twice, not once. Relatedly, the semantics of this function would be too different from `eval()` to qualify as a `2` function.

lionel-:
How about `value()`? This is related to `eval()` but distinct enough, and mirrors `expr()` nicely.

* `call2()` vs `exec()`
* `expr()` vs `value()`

vspinu:
My 2c. I think the name should be short, very short. This function is likely to be used everywhere and should be minimal on typing and space consumption. 

I would call it either `e` or `ee` because it's essentially a shortcut for `eval(expr(...))`.

lionel-:
@hadley @jimhester What do you think about `expand()`?

```r
args <- list(1:2, 3:4)

expand(rbind(!!!args))
#>      [,1] [,2]
#> [1,]    1    2
#> [2,]    3    4
```

"expand" should be suggestive of expanding the call with lists of arguments or inlined arguments.

vspinu:
`unbang`  maybe.

lionel-:
@krlmlr pointed out that `expand()` is a tidyr verb, which might be a problem since that function has the potential to be end-user facing.

krlmlr:
Perhaps `enexpr_eval()` (because it's a combo of enexpr + eval), and `enquo_eval()` for a corresponding function that uses enquo + eval?

lionel-:
These functions need end-user naming, not programmer naming.

krlmlr:
I'd expect these functions to be useful primarily for learning; when implementing own functions, `eval_tidy()` seems more useful. Where do you see use cases in data analysis?

There's `qq_show()`, maybe `qq_eval()` (but how to distinguish the expr variant)?

lionel-:
> I'd expect these functions to be useful primarily for learning

That might be the case for `with_data()`, but `expand()` / `value()` / `bang()` is generally useful for calling functions with programmatically defined arguments. I use it in several packages. Also I think it should be the preferred way of creating functions and formulas programmatically, instead of `expr_interp()`.

> maybe qq_eval()

It can't be an `eval` function, because these evaluate twice (one time to get the argument, and then another time). Here the function evaluates only once.

lionel-:
@wch suggests `inject()`

wch:
I've been writing code with the (placeholder) name `blast()`, and as I've been writing comments, I find that I say things like, "Use `blast()` to inject a quosure into the function call." So I think `inject()` is good, since that's the verb I'm using to actually describe what is happening.

lionel-:
I think I like it!

PR #595, opened 2018-08-31
Better trace and trail collapsing

lionel-:
* Collapse eval/evalq frames
* Collapse magrittr implementation
* Unroll magrittr calls in backtrails


tests/testthat/test-trace-collapse-children.txt
```diff
@@ -0,0 +1,17 @@
...
+█
+└─f()
+  └─eval(quote(g()), env())
+    └─eval(quote(g()), env())
+      └─g()
+
+Collapsed:
+█
+└─f()
+  └─[ eval(...) ] ... +1
```
hadley: I wonder if it would be better to do `eval()` here, to eliminate one set of `...`. It feels like it might be confusing currently

tests/testthat/test-trace-collapse-children.txt
```diff
@@ -0,0 +1,17 @@
...
+    └─g()
+
+Trail:
+ ─f()
+ ─[ eval(...) ] ... +1
```
hadley: `... +1` is very concise. I think I bought this up before, but do you think we could use some words here? Like `... and 1 more call`.

hadley: tibble does "... with 53,930 more rows" so maybe borrow that language for familiarity?

tests/testthat/test-trace-collapse-children.txt
lionel-: But then it might be ambiguous with empty `f()` calls. If we use the more verbose "with 1 more call" there would only be one ellipsis.

tests/testthat/test-trace-collapse-children.txt
lionel-: ok let's try it out. Any objections to make this configurable?

lionel-:
@hadley After spending the day implementing pipe unrolling I went to bed thinking that actually it would be better not to unroll :/ It'd be more consistent to hide it entirely because all those past calls have already been evaluated. A backtrace / backtrail shows the calls that haven't been completed yet and to which you could `recover()` into. Maybe we should show `... %>% mutate(., foo)`?

tests/testthat/test-trace-collapse-children.txt
hadley: Yeah, sounds good.

tests/testthat/test-trace-collapse-children.txt
hadley: The downside of making anything configurable is that we now have to test under both configurations. So the added complexity has to have some clear payoff.

hadley:
Hmmmm, I think that might be an overly technical explanation - I think the point of backtrail is to show you the sequence of calls that lead to the failure. In other words, while technically correct, I think your interpretation will be less helpful to most R users.

PR #600, opened 2018-09-05
Simplify quosure evaluation

lionel-:


lionel-:
@hadley I refactored quosure eval to simplify the code and make it easier to fix a bug, but we got a nice perf boost as well. Not sure what was causing all those gc.

```r
bench::mark(iterations = 50000,
  `No mask` = eval_tidy(expr),
  `With mask` = eval_tidy(expr, mtcars)
)
```

```r
### Quosure refactor

# A tibble: 2 x 14
  expression    min   mean median   max `itr/sec` mem_alloc  n_gc n_itr
  <chr>      <bch:> <bch:> <bch:> <bch>     <dbl> <bch:byt> <dbl> <int>
1 No mask    15.8µs 23.4µs 18.9µs 688µs    42801.        0B    13 49987
2 With mask  19.1µs 28.3µs 22.8µs 451µs    35336.      560B    16 49984


### 3.0 dev

# A tibble: 2 x 14
  expression     min    mean median      max `itr/sec` mem_alloc  n_gc n_itr
  <chr>      <bch:t> <bch:t> <bch:> <bch:tm>     <dbl> <bch:byt> <dbl> <int>
1 No mask     42.7µs  60.6µs   49µs 938.87µs    16513.        0B    33 49967
2 With mask  239.9µs 349.8µs  282µs   4.67ms     2859.      560B   142 49858
# ... with 5 more variables: total_time <bch:tm>, result <list>,
#   memory <list>, time <list>, gc <list>
```

src/internal/eval-tidy.c
```diff
@@ -379,8 +377,17 @@ sexp* rlang_eval_tidy(sexp* expr, sexp* data, sexp* frame) {
 const char* data_pronoun_c_names[4] = { "src", "lookup_msg", "read_only", NULL };
 
 void rlang_init_eval_tidy() {
-  tilde_thunk_fmls = rlang_constants_get("tilde_thunk_fmls");
-  tilde_thunk_body = rlang_constants_get("tilde_thunk_body");
+  tilde_fn = r_parse_eval(
```
hadley: Is there a reason these have to be created in C code? Why can't we store in the package?

hadley:
Nice!

src/internal/eval-tidy.c
lionel-: It's completely equivalent but I think storing on the R side makes things a bit more complicated overall, especially as you start using this pattern more generally. It is nice to see the code inline.

Where it makes sense, I think we should replace the dynamic creation of language objects with cons/lang by parsing at load-time. It is faster, less error prone, and more flexible. In C this has to be done in an init function and stored in a global variable but in C++ you can dynamically initialise in a static-local variable, which is even nicer.

Too bad C / C++ don't have multiline literal strings though.

lionel-: But we optimise for reading the code? In this case I don't want to create these objects at load-time particularly. But being able to do it at load-time means that this is a general solution that even works in header-only libraries that don't have a namespace (esp. with C++).

I think parsing also has advantages when writing the code. It's easy to parse inline small pieces of code and when performance matters you move it to a static variable. Here's another example: https://github.com/r-lib/rlang/pull/601/commits/760f6fce2a5e73420ea0a920b7c7d49948beed19

You may be right that in this case it's better to build them in the namespace. It obviously depends on the size of the function, whether you have your own namespace, etc. As all programming there are tradeoffs, I'm just experimenting with this new idiom that I like so I can develop a feel for it.

PR #601, opened 2018-09-06
Lifecycle of UQE() and namespaced UQ()/UQS() calls

lionel-:
@hadley How about giving examples in the warning?

```r
expr(rlang::UQ("foo"))
#> [1] "foo"
#> Warning message:
#> Prefixing `UQ()` with the rlang namespace is deprecated as of rlang 0.3.0.
#> Please use the non-prefixed form or `!!` instead.
#>
#>   # Bad:
#>   rlang::expr(mean(rlang::UQ(var) * 100))
#>
#>   # Ok:
#>   rlang::expr(mean(UQ(var) * 100))
#>
#>   # Good:
#>   rlang::expr(mean(!!var * 100))
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,23 @@
 
 # rlang 0.2.2.9000
 
+* Calling `UQ()` and `UQS()` with the rlang namespace qualifier is
+  deprecated as of rlang 0.3.0. Just use the unqualified forms
+  instead:
+
+  ```
+  # Bad
+  rlang::expr(mean(rlang::UQ(var) * 100))
+
+  # Ok
+  rlang::expr(mean(UQ(var) * 100))
+
+  # Good
+  rlang::expr(mean(!!var * 100))
+  ```
+
+  Although soft-deprecated since rlang 0.2.0, `UQ()` and `UQS()` can still be used for now.
+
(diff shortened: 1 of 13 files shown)


hadley:
I like the idea, but I'm slightly worried that it's going to lead to a _lot_ of warnings if you're using a package that uses `rlang::UQ()`

lionel-:
I was wondering if we should display deprecation warnings only once per top-level call.

lionel-:
The warning is now only displayed once per session. This is mentioned in the message:

```r
exprs(rlang::UQS(1:2))
#> [[1]]
#> [1] 1
#>
#> [[2]]
#> [1] 2
#>
#> Warning message:
#> Prefixing `UQS()` with the rlang namespace is deprecated as of rlang 0.3.0.
#> Please use the non-prefixed form or `!!!` instead.
#>
#>   # Bad:
#>   rlang::expr(mean(rlang::UQS(args)))
#>
#>   # Ok:
#>   rlang::expr(mean(UQS(args)))
#>
#>   # Good:
#>   rlang::expr(mean(!!!args))
#>
#> This warning is only displayed once per session.

exprs(rlang::UQS(1:2))
#> [[1]]
#> [1] 1
#>
#> [[2]]
#> [1] 2
```

PR #603, opened 2018-09-11
Update data mask API

### LINKED ISSUE #571: Using eval_tidy() with environments is not straightforward / well documented
library(rlang)
    myquo <- quo((Temp - 32) / 1.8)
    eval_tidy(myquo, data = as_data_mask(datasets::airquality)) # works
    e <- as_env(datasets::airquality, parent = global_env())
    eval_tidy(myquo, data = as_data_mask(list(), parent = e))   # error

I expected `Temp` to be found in `e`.

I have R version 3.5.0 and tested this with both latest CRAN and GitHub versions of {rlang}.

***
PS: I've posted this on SO: https://stackoverflow.com/q/51830920/6103040

lionel-:
* Deprecate bare environments passed to `eval_tidy()` and `as_data_mask()` Passing bare environments is confusing because their parents is silently changed to the current evaluation env. (There is a similar confusion in `base::eval()` except that it goes the other way around, the current evaluation env is silently ignored.)

* Deprecate `parent` arguments of `as_data_mask()` and `new_data_mask()`.

The idea behind these two changes is that the parent of the data mask is undetermined at time of creation. The parent changes during evaluation to the `env` argument of `eval_tidy()`, or to quosure environments when applicable. Given these semantics, allowing parents at data mask creation and supporting bare environments were confusing.

Closes #571

### DIFF
--- a/R/eval-tidy.R
+++ b/R/eval-tidy.R
@@ -21,6 +21,11 @@
 #'
 #' `eval_tidy()` is stable.
 #'
+#' **rlang 0.3.0**
+#'
+#' Passing an environment to `data` is deprecated. Please construct an
+#' rlang data mask with [new_data_mask()].
+#'
 #' @param expr An expression to evaluate.
 #' @param data A data frame, or named list or vector. Alternatively, a
 #'   data mask created with [as_data_mask()] or [new_data_mask()].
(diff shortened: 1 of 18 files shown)


R/eval-tidy.R
```diff
@@ -251,16 +295,19 @@ delayedAssign(".data", as_data_pronoun(list()))
...
-#' # upper level:
-#' try(eval_tidy(quote(a + b), data = bottom))
-as_data_mask <- function(data, parent = base_env()) {
-  .Call(rlang_as_data_mask, data, parent)
+as_data_mask <- function(data, parent = NULL) {
+  if (!is_null(parent)) {
+    id <- "rlang::as_data_mask() - parent argument"
+    msg <- paste(sep = "\n",
+      "The `parent` argument of `as_data_mask()` is deprecated.",
+      "The parent of the data mask is determined from either:\n",
```
hadley: I think this would be clearer if you added an extra `""` and got rid of the `\n` (and maybe defined `paste_lines()` in utils

PR #608, opened 2018-09-14
Add last_error()

lionel-:
This is an approximation to saving unhandled errors for later retrieval. Ideally we'd use a top-level handler. Since we can't do that, we save the errors when they are printed at the console. This normally only happens when the error goes unhandled.

There are some exceptions, for example a verbose `try()` calls `conditionMessage()`, or errors demoted to warnings in the namespace loading code. That seems alright because in general the conditionMessage method is called when the error message is printed at the console and then it makes sense to provide a means of retrieving it.

cc @gaborcsardi @jimhester


gaborcsardi:
I think this is nice and simple. The only thing I am not sure about is whether it is worth it, considering that top-level handlers  will be the ultimate solution. It might very well be that by the time we are using rlang to throw errors in our packages, we'll also have top level handlers, which can catch all errors.

R/cnd.R
```diff
@@ -582,8 +590,14 @@ print.rlang_error <- function(x,
...
 
+# Last error to be returned in last_error()
+last_error_env <- new.env(parent = emptyenv())
+last_error_env$cnd <- NULL
+
 #' @export
 conditionMessage.rlang_error <- function(c) {
+  last_error_env$cnd <- c
+
   lines <- c$message
```
gaborcsardi: Shouldn't you call NextMethod  somehow to get the message? I do have some packages that override`conditionMessage()` to create  a custom message. It would be great if `last_error()` worked for these cases as well.

lionel-:
I think it's worth it because it's simple and we're already using `abort()` in a few tidyverse packages.

Also there'll be `with_abort()` to convert base errors to rlang errors so that you can rethrow with a parent errors and compartmentalised backtraces. And we'll provide an `options(error = )` instrumenter as well.

R/trace.R
```diff
@@ -501,6 +501,18 @@ trace_simplify_collapse <- function(trace) {
   trace[rev(path)]
 }
 
+
+#' Last `abort()` error
+#'
+#' This returns the last error thrown with [abort()]. The error is
+#' printed with a backtrace.
```
gaborcsardi: Maybe mention that handled errors might also set the last error.

R/cnd.R
lionel-: What are your use cases for `conditionMessage()`? To get some idea of the variety of purposes for it.

gaborcsardi: https://github.com/r-lib/async/blob/widestack/R/widestack.R#L334

Although right now this is not  in master....  it prints more information on an async error.

gaborcsardi: Basically, because `print.error()` is not used when printing the error message, any time you want  to include extra information, you need to use  `conditionMessage()`.

hadley: But why not generate the complete message when you throw the error?

gaborcsardi: You could. It is just easier and less expensive to just put the data in, and format the message on  demand, typically only if the error was not caught. 

In the particular case of async, the error is just a normal error, the async backend puts there the async information later, piece by piece. It could (re)format the message every time, but it is much cheaper and simpler just to collect everything. Especially because AFAICT the errors are often caught here, so we might not even need the message.

Anyway, we can deal with this later, not very important, and actually not even strictly related to this PR.

lionel-: Perhaps the way to do it is to update the `message` field before calling `NextMethod()`? This way it happens lazily and the error message can be expanded until it reaches the final method.

lionel-: There's a tension here because normally specific methods expand around base methods (e.g. to reconstruct), but here we'd like to have it the other way around. In fact we are really after what is called `:around` method combination in Common Lisp: a method implementation that wraps all other methods.

Since S3 is so limited, we are now trying to find a workaround that reverses the natural order of method combination just so that our base rlang error method can expand last instead of first.

gaborcsardi: Strictly speaking, an error object does not even have to include a `message` entry,  the API is  `conditionMessage()`

The other thing is that for better or worse I do this: 
```r
class(err)  <- unique("async_rejected_error", class(err))
```
(FIXME if there is a better idiom.) As I understand `rlang_error` will be added similarly to the  class. So basically, `async_rejected_error` and `rlang_rejected_error` may appear in arbitrary order.

All this said, this is not a common use case, and we can deal with it later.

gaborcsardi: > I don't think this is a good way of adding a class as the ordering is very important for method combination.

What is a  better way?

lionel-: We have to discover it. Figuring out proper class hierarchies and method combinations within the constraints of S3 is a WIP :) Perhaps there is no good way.

hadley: I suspect you're better off composing rather than subclassing, but that probably requires a bunch of infrastructure in order to forward most methods.

PR #609, opened 2018-09-14
Make backtrace display optional

lionel-:
Includes #608

* Backtraces are no longer displayed by default.

* Introduces the experimental option `rlang__backtrace_on_error`. It can be set to `reminder` (to remind the user to call `rlang::last_error()` to see a backtrace), `branch`, `collapse` or `full` (to control the verbosity of the backtrace).

    See [test-on-error-message-options.txt](https://github.com/r-lib/rlang/compare/master...lionel-:impl-onerror-option?expand=1#diff-1b9d45add965cd1603af2dff672624ad) for the different outputs.

The reminder is displayed with `crayon::silver()` as an approximation to `crayon::blurred()` which doesn't work in RStudio. This makes the reminder appear dimmer than the error message. Perhaps the whole backtrace should get this treatment?

In packages, `abort()` can be wrapped like this:

```r
my_abort <- function(...) {
  with_options(
    rlang__backtrace_on_error = "reminder",
    abort(...)
  )
}

my_abort("foo")
#> Error: foo
#> Call `rlang::last_error()` to see a backtrace
```

Is this sufficient or do we still need the extra argument @hadley?


hadley:
I don't think we need the argument.

But I'd like to get @jennybc's opinion on this, because I think she has a better sense of how users struggle with error messages.

jennybc:
Shall I just take this out for a spin or is there something specific I should contemplate? E.g. the argument you two are discussing above.

lionel-:
The last commit adds an unexported error handler `add_backtrace()` to promote base errors to rlang errors, cc @jimhester @gaborcsardi. When this handler is set:

* All errors are saved and can be inspected with `rlang::last_error()`
* All errors display a backtrace according to the `rlang__backtrace_on_error` option.

This should make it easier to test the backtrace stuff. @jennybc Could you install this branch and play with the backtrace-on-error options please?

This will display a simplified backtrace on error:

```r
options(
  rlang__backtrace_on_error = "branch",
  error = rlang:::add_backtrace
)
```

This displays a reminder to call `rlang::last_error()` to see a backtrace:

```r
options(
  rlang__backtrace_on_error = "reminder",
  error = rlang:::add_backtrace
)
```

gaborcsardi:
Yeah, I think this is good. My main issue with `options(error)` handlers is that you can only have one. I have one for tracer, another for gitty, another for prompt....

lionel-:
> My main issue with options(error) handlers is that you can only have one. I have one for tracer, another for gitty, another for prompt....

I've been thinking for a while that function hooks like in Emacs would be useful, especially when we flesh out the events API. With hooks you could do

```r
options(error = function() hook_run(on_error_hook))
```

They'd come with a C API so you could easily add dtors:

```r
hook <- new_hook()
on.exit(hook_run(hook))

.Call(my_c_function, hook)
```

gaborcsardi:
Yeah, hooks would be awesome.

PR #610, opened 2018-09-14
 Implement str() method for rlang errors 

lionel-:
Includes #608.

Compared to the print method which displays a simplified backtrace, `str()` displays the full call tree as well as the condition fields. This makes it easy to print the full backtrace of the last error:

```
str(last_error())
```

See [test-cnd-error-str.txt](https://github.com/r-lib/rlang/pull/610/files#diff-049078f9aaca010092da520fab9d7269) for an example output.


hadley:
Is `str()` the right method for this? I would've thought `summary()` (which confusingly tends to display more information than `print()`)

lionel-:
I thought about `str()` because it prints the structure of the backtrace. `summary()` sounds good as well.

lionel-:
Perhaps `str()` would make sense if the subsetting methods actually subsetted the tree, e.g. `trace[[2]]` would select the second sibling branch. So `summary()` does make more sense.

PR #625, opened 2018-09-25
Improve empty punning idioms for trace subsetting

lionel-:
* `seq2()` recycles to length zero over its arguments.
* Use `x[vec %0% expr()]` idiom to escape from empty punning while subsetting.
* Draft `match2()`, an empty punning version of `match()` with better argument ordering.

### DIFF
--- a/R/vec-utils.R
+++ b/R/vec-utils.R
@@ -128,6 +128,16 @@ seq2_along <- function(from, x) {
   seq2(from, length(x))
 }
 
+# Better argument ordering and empty punning when no match
+match2 <- function(vec, elt) {
+  idx <- match(elt, vec)
+  if (identical(idx, na_int)) {
+    int()
+  } else {
+    idx
+  }
+}
+
 first <- function(x) {
   .subset2(x, 1L)
 }
--- a/R/vec-utils.R
+++ b/R/vec-utils.R
@@ -106,6 +106,16 @@ modify <- function(.x, ...) {
 #'
 #' seq2_along(10, letters)
 seq2 <- function(from, to) {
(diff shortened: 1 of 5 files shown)


R/vec-utils.R
```diff
@@ -106,6 +106,16 @@ modify <- function(.x, ...) {
 #'
 #' seq2_along(10, letters)
 seq2 <- function(from, to) {
+  n_from <- length(from)
```
hadley: This seems a bit weird to be since I thought `seq2()`  was `seq2(<integer>[1], <integer>[1])` — making it length 0 or 1 feels odd

R/vec-utils.R
```diff
@@ -118,6 +128,16 @@ seq2_along <- function(from, x) {
   seq2(from, length(x))
 }
 
+# Better argument ordering and empty punning when no match
+match2 <- function(vec, elt) {
```
hadley: Would this be better as `match1()` since I think you're implying the `elt` is length 1?

R/vec-utils.R
lionel-: While not vectorised on its inputs, this function is used in vectorised context as it returns a vector. Perhaps I should implement this in a different unexported function so we can get a feel for it?

R/vec-utils.R
lionel-: Good point I hadn't considered the vector elements case.

R/vec-utils.R
hadley: I feel quite negatively towards it right now. The type feels very weird.

Issue #598, opened 2018-09-04
Forbid `<-` calls in `list2()`


lionel-:
To catch errors when user confused `=` and `<-`. It'd still be possible to do assignments by wrapping in `(`.

lionel-:
Now optional.

PR #632, opened 2018-09-26
Add `.check_assign` argument to dots collectors

lionel-:
`list2()` and functions collecting dots with `enquos()` and `enexprs()` now warn when `<-` is passed as argument:

```r
list2(a <- 1)
#>Warning: Using `<-` as argument is often a mistake.
#>Do you need to use `=` to match an argument?
#>
#>If you really want to use `<-`, please wrap in parentheses:
#>
#>  # Bad:
#>  fn(a <- 1)
#>
#>  # Good:
#>  fn(a = 1)     # Match 1 to parameter `a`
#>  fn((a <- 1))  # Assign 1 to variable `a`
#>[[1]]
#>[1] 1
```

`exprs()` and `quos()` do not warn.

The behaviour is hardcoded for `list2()`, `exprs()` and `quos()`, but can be configured in `dots_list()`, `enexprs()` and `enquos()`.


hadley:
What's the motivation for this change?

lionel-:
Catch this kind of errors: https://stackoverflow.com/questions/45383616/how-to-interpret-error-elements-must-be-named-when-sourcing-an-r6-class

hadley:
I'm slightly worried about turning this on by default, but maybe it's not important.

lionel-:
We could reconsider the default if we see that it causes many revdep problems (I don't think it will but you never know).

lionel-:
Closing for now, we'll revisit after the release.

ping #598

PR #633, opened 2018-09-27
Fix !!! input types

### LINKED ISSUE #572: list2(!!!x) should not work with language lists
```r
list2(!!!~foo)
#> [[1]]
#> `~`
#> 
#> [[2]]
#> foo
```

lionel-:
Closes #572

Changes for `list2()`:

* Accept pairlists
* Fail with language objects
* Call `base::as.list()` for classed vector objects

Changes for `exprs()` and `quos()`:

* Non-vector non-language objects like environments and functions now fail. Before they were automatically wrapped in a list to allow splicing.

* Call `base::as.list()` for classed vector objects.

* Soft-deprecate splicing of language objects, including scalar quosures. This makes this kind of code throw an error: https://github.com/tidyverse/tidyr/commit/82ef03af587b5ad57e6edf185a0ca5af8c969d4b#diff-65e6a5a29a28bb3768b7a3c192f952f1R147.

    ```r
    #> Warning message:
    #> Unquoting language objects with `!!!` is soft-deprecated as of rlang 0.3.0.
    #> Please use `!!` instead.
    #>
    #>   # Bad:
    #>   dplyr::select(data, !!!enquo(x))
    #>
    #>   # Good:
    #>   dplyr::select(data, !!enquo(x))    # Unquote single quosure
    #>   dplyr::select(data, !!!enquos(x))  # Splice list of quosures
    #>
    #> This warning is displayed once per session.
    ```

    The soft-deprecation won't be very verbose because there's no reliable way of detecting the caller env, so the warning only occurs when rlang is attached. However we can't deprecate right away because of the tidyr bug (dplyr and tidyselect are fine).

With these changes, and once splicing scalar language objects is defunct, the semantics of `!!!` for `list2()` and `exprs()`/`quos()` will be very similar. I think it'd be a good idea to make them identical. This involves supporting quoted `{` calls in `list2()`:

```r
list2(!!!quote({ foo; bar; }))
#> [[1]]
#> foo
#>
#> [[2]]
#> bar
```

Should we make this change?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,9 +8,12 @@
   and in quoting functions. `!!!` accepts these types:
 
   - Lists, pairlists, and atomic vectors. If they have a class, they
-    are converted with `base::as.list()` to allow S3 and S4 dispatch.
+    are converted with `base::as.list()` to allow S3 dispatch.
     Following this change, objects like factors can now be spliced
     without data loss.
+
+  - S4 objects. These are converted with `as(obj, "list")` before
+    splicing.
 
   - Quoted blocks of expressions, i.e. `{ }` calls
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,29 @@
 
 # rlang 0.2.2.9000
 
+* The input types of `!!!` have been standardised. `!!!` is generally
+  defined on vectors: it takes a vector (typically, a list) and
+  unquotes each element as a separate argument. The standardisation
+  makes `!!!` behave the same in functions taking dots with `list2()`
+  and in quoting functions. `!!!` accepts these types:
+
+  - Lists, pairlists, and atomic vectors. If they have a class, they
+    are converted with `base::as.list()` to allow S3 and S4 dispatch.
+    Following this change, objects like factors can now be spliced
+    without data loss.
+
+  - Quoted blocks of expressions, i.e. `{ }` calls
+
+  `!!!` disallows:
+
+  - Any other objects like functions or environments, but also
+    language objects like formula, symbols, or quosures.
+
+  Quoting functions used to automatically wrap language objects in
+  lists to make them spliceable. This behaviour is now soft-deprecated
+  and it is no longer valid to write `!!!enquo(x)`. Please unquote
+  scalar objects with `!!` instead.
+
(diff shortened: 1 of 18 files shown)


hadley:
The tidyr case seems clearly just like a typo/think-o on my behalf, right? I can't imagination why I'd try and unquote-splice a single element.

hadley:
So to be clear, in `!!!x`, `x` must now be an atomic vector, a list, or a pairlist. If `is.object()` is true, then `as.list()` will be called to convert to a list?

hadley:
Finally, what's our policy on expression objects? Should they be spliced, or throw an error?

lionel-:
Currently expression vectors are an error because generally not supported in rlang, but it would be easy to convert them to list. I recently realised expression vectors might be useful in some cases because they carry srcrefs around.

What do you think about making `list2(!!!x)` and `expr(foo(!!!x))` completely uniform and thus supporting quoted `{` in `list2()`?

hadley:
I don't think that's something that you'd ever actually use, but if it simplifies the code, I don't see a reason not to do it.

lionel-:
I just added support for S4 and deep splicing. Dots-splicing now follows a unique code path. The news item is now:

```
* The input types of `!!!` have been standardised. `!!!` is generally
  defined on vectors: it takes a vector (typically, a list) and
  unquotes each element as a separate argument. The standardisation
  makes `!!!` behave the same in functions taking dots with `list2()`
  and in quoting functions. `!!!` accepts these types:

  - Lists, pairlists, and atomic vectors. If they have a class, they
    are converted with `base::as.list()` to allow S3 dispatch.
    Following this change, objects like factors can now be spliced
    without data loss.

  - S4 objects. These are converted with `as(obj, "list")` before
    splicing.

  - Quoted blocks of expressions, i.e. `{ }` calls

  `!!!` disallows:

  - Any other objects like functions or environments, but also
    language objects like formula, symbols, or quosures.

  Quoting functions used to automatically wrap language objects in
  lists to make them spliceable. This behaviour is now soft-deprecated
  and it is no longer valid to write `!!!enquo(x)`. Please unquote
  scalar objects with `!!` instead.
```

PR #635, opened 2018-09-28
Add call_print_type()

### LINKED ISSUE #176: Add is_infix_call()

lionel-:
Closes #176


R/call.R
```diff
@@ -225,6 +225,111 @@ is_call <- function(x, name = NULL, n = NULL, ns = NULL) {
...
+    `!!!` =
+      "prefix",
+    `function` = ,
+    `while` = ,
+    `for` = ,
+    `repeat` = ,
+    `if` = ,
+    `(` = ,
+    `{` =
+      "control-flow",
```
hadley: Could we make it match https://adv-r.hadley.nz/functions.html#function-forms ?  I don't love the characterisation of `[[` as postfix, since there's still the closing `]]`. And I think "special" is general enough for most purposes.

lionel-: Makes sense, done.

Issue #325, opened 2017-12-21
Should set_names() recycle length-1 names vectors?


lionel-:
cc @jennybc

lionel-:
And do we need `set_names_at()` that initialises missing names to `""` instead of `NA`?

From @jennybc:

> To summarize what I find weird with base handling:
> Inconsistency around the lack of a name: `NULL` vs `NA` vs “”
> Recycling behavior (or lack thereof, actually) wrt assignment that feels inconsistent with how R usually works

lionel-:
Or support only `""` and `NA` as it is better for names to be unique. I.e. `set_names(x, "")` would reinitialise the names with an empty string vector.

lionel-:
@hadley mentioned that names should generally be unique because they are typically used as element-wise identifiers. But sometimes they are group identifiers. In this case it might make sense to recycle length-1 names:

```r
vec <- c(set_names(vec1, "group1"), set_names(vec2, "group2"))
tibble::enframe(vec)
```

cc @jennybc

jennybc:
I seem to run into cases where names are non-unique fairly often. For example, in forming the query part of an API call. I'll make a point of noticing some IRL examples.

krlmlr:
I'm thinking about using a named vector with non-unique names in `tibble::tidy_names()` for https://github.com/tidyverse/tibble/issues/406. Here I'd like _rlang_ to keep all `NA` and `""` unchanged, but I'm not sure if this requirement is useful in other cases.

lionel-:
It now occurs to me the recycling could go both ways as a handy way of creating vectors along names:

```r
set_names(NA, c("foo", "bar"))
#> foo bar 
#>  NA  NA 
```

hadley:
I think this is not possible because our [standard recycling rules](https://github.com/tidyverse/principles/issues/13) give length-0 output when any input is length-0. That conflicts with the existing use of `set_names(x, NULL)` to remove names.

PR #639, opened 2018-10-02
Add .homonyms argument to dots collectors

lionel-:
```r
mylist <- function(...) {
  dots_list(..., .homonyms = "first")
}

mylist(1, a = 2, b = 3, 4, b = 5, b = 6, 7, a = 8)
#> [[1]]
#> [1] 1
#>
#> $a
#> [1] 2
#>
#> $b
#> [1] 3
#>
#> [[4]]
#> [1] 4
#>
#> [[5]]
#> [1] 7
```

```r
mylist <- function(...) {
  dots_list(..., .homonyms = "last")
}

mylist(1, a = 2, b = 3, 4, b = 5, b = 6, 7, a = 8)
#> [[1]]
#> [1] 1
#>
#> [[2]]
#> [1] 4
#>
#> $b
#> [1] 6
#>
#> [[4]]
#> [1] 7
#>
#> $a
#> [1] 8
```

```r
mylist <- function(...) {
  dots_list(..., .homonyms = "error")
}

mylist(1, a = 2, b = 3, 4, b = 5, b = 6, 7, a = 8)
#> Error: Arguments can't have the same name. We found these problems:
#> * Multiple arguments named `a` at positions 2 and 8
#> * Multiple arguments named `b` at positions 3, 5 and 6

mylist(1, a = 2, b = 3, 4, b = 5, b = 6, 7)
#> Error: Arguments can't have the same name.
#> We found multiple arguments named `b` at positions 3, 5 and 6
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,12 @@
 
 # rlang 0.2.2.9000
 
+* `dots_list()`, `enexprs()` and `enquos()` gain a `.homonyms`
+  argument to control how to treat arguments with the same name.
+  The default is to keep them. Set it to `"first"` or `"last"` to keep
+  only the first or last occurrences. Set it to `"error"` to raise an
+  informative error about the arguments with duplicated names.
+
 * Automatic naming of expressions now uses `quo_name()` instead of
   `quo_text()`. This makes it compatible with all object types,
   prevents multi-line names, and ensures `name` and `.data[["name"]]`
--- a/R/compat-purrr.R
+++ b/R/compat-purrr.R
@@ -50,7 +50,7 @@ pluck_cpl <- function(.x, .f) {
 }
 
 map2 <- function(.x, .y, .f, ...) {
-  mapply(.f, .x, .y, MoreArgs = list(...), SIMPLIFY = FALSE)
+  set_names(mapply(.f, .x, .y, MoreArgs = list(...), SIMPLIFY = FALSE), names(.x))
 }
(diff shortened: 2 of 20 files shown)


NEWS.md
```diff
@@ -1,6 +1,12 @@
 
 # rlang 0.2.2.9000
 
+* `dots_list()`, `enexprs()` and `enquos()` gain a `.homonyms`
+  argument to control how to treat arguments with duplicated names.
```
hadley: duplicated -> "the same"

R/compat-purrr.R
```diff
@@ -50,7 +50,7 @@ pluck_cpl <- function(.x, .f) {
 }
 
 map2 <- function(.x, .y, .f, ...) {
-  mapply(.f, .x, .y, MoreArgs = list(...), SIMPLIFY = FALSE)
+  set_names(mapply(.f, .x, .y, MoreArgs = list(...), SIMPLIFY = FALSE), names(.x))
```
hadley: I have a vague recollection that setting names unconditionally like this is a bit risky, but I can't remember why

lionel-: Perhaps because `set_names()` does not recycle whereas `mapply()` does? I think `set_names()` should recycle, it makes sense in some situations, and is handy for `set_names(x, "")`: https://github.com/r-lib/rlang/issues/325

However I see that that `purrr::map2()` discards the names when it recycles. So for consistency I'll check for length.

PR #642, opened 2018-10-03
Support large double values in is_integerish()

### LINKED ISSUE #578: is_integerish() and large numbers
``` r
rlang::is_integerish(1e10)
#> Warning in rlang::is_integerish(1e+10): NAs introduced by coercion to
#> integer range
#> [1] NA
```

> Created on 2018-08-20 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

Do we need to use `trunc()` instead of `as.integer()` ?

lionel-:
Closes #578

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 
 # rlang 0.2.2.9000
 
+* `is_integerish()` now supports large double values (#578).
+
 * The sentinel for removing arguments in `call_modify()` has been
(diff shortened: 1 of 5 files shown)


hadley:
Also need to consider the case of the double that equals the smallest integer, because R uses that value for `NA_INTEGER`

hadley:
This doesn't involve coercion, so it's not important. Need to note that even if `is.integerish()` is all `TRUE`, `is.integer()` might produce NAs

PR #646, opened 2018-10-04
Implement zap sentinel for removing elements

lionel-:
Creating zap objects:

```r
zap()
#> <zap>

zaps(3)
#> [[1]]
#> <zap>
#>
#> [[2]]
#> <zap>
#>
#> [[3]]
#> <zap>

zaps(c("foo", "bar"))
#> $foo
#> <zap>
#>
#> $bar
#> <zap>
```

Zapping arguments:

```r
call <- quote(foo())

call <- call_modify(call, bar = NULL, baz = )
call
#> foo(bar = NULL, baz = )

call_modify(call, bar = zap())
#> foo(baz = )

call_modify(call, !!!zaps(c("foo", "bar", "baz")))
#> foo()
```

Zapping bindings:

```r
env <- env()

env_bind(env, foo = )
env$foo
#>

env_bind(env, foo = NULL)
env$foo
#> NULL

env_has(env, "foo")
#>  foo
#> TRUE

env_bind(env, foo = zap())
env$foo
#> NULL

env_has(env, "foo")
#>   foo
#> FALSE
```


hadley:
`zaps()` feels a bit too special purpose to me, so I'd prefer not to export it yet.

lionel-:
@hadley Added `rep_named()` in https://github.com/r-lib/rlang/pull/646/commits/dbf1480cb45ab57a6a3da57da6232f8c29af351c.

lionel-:
cc @vspinu, I think you're the only user of `call_modify()` ;)

We changed the meaning of `arg = NULL`, it will now create a `NULL` argument.

PR #647, opened 2018-10-04
Fix CRAN dplyr

lionel-:
`dplyr:::tidy_text()` uses `quo_text()` instead of `quo_name()`:

https://github.com/cran/dplyr/blob/fe5edb1b273e1c6f091f88bd64f627a0a2aee6f6/R/utils-expr.R#L45

This is problematic now that `quo_text()` deparses non-syntactic symbols with backticks and it causes all sorts of issues with default names. This PR implements a dynamic compat to ease out the transition.


R/quotation.R
```diff
@@ -413,6 +413,12 @@ endots <- function(call,
 #'   argument and `width` as the second argument.
 #' @export
 exprs_auto_name <- function(exprs, width = NULL, printer = expr_name) {
+  if (is_installed("dplyr") &&
+        utils::packageVersion("dplyr") < "0.7.99" &&
```
krlmlr: The `is_reference()` call is two orders of magnitude faster than `utils::packageVersion()` .

Perhaps we could just check if `printer` evaluates to `dplyr:::tidy_text` (if that exists), and make sure we rename in dev dplyr?

hadley:
This seems ok to me, but it seems like a high risk change — how can we check that all four combinations of dev and cran, dplyr and rlang work?

lionel-:
I have checked manually.

The idea of numbering the 0.8 branch as 0.7.99 is helpful regarding future releases. If we fix it in 0.7.8, we can use `< "0.7.8"` instead.

R/quotation.R
lionel-: we could do that,`tidy_text()` should be removed from dplyr

lionel-:
@krlmlr The check is now computed only once and lazily.

Did you mean 0.7.7 instead of 0.7.8? The check is now `< "0.7.6"`, is that correct or is a 0.7.7 already on its way to CRAN?

PR #660, opened 2018-10-15
Add namespace prefixes to backtrace calls

lionel-:
* Exported functions are prefixed with `::`

* Unexported functions are prefixed with `:::`

* Functions inheriting from the global env are prefixed with `global::`.

* Namespace is determined with the topenv rather than the parent env. This way we handle private closures such as `base:::tryCatchOne()`.

```r
library("dplyr")

myfunction <- function() {
  mtcars %>%
    group_by(cyl) %>%
    select(last_col(Inf))
}

myfunction()
#> Error: `offset` must be smaller than the number of columns
#> Call `rlang::last_error()` to see a backtrace

rlang::last_error()
#> <error>
#> * Message: "`offset` must be smaller than the number of columns"
#> * Class: `rlang_error`
#> * Backtrace:
#>  ─global::myfunction()
#>  ─tidyselect::last_col(Inf)
#> Call `summary(rlang::last_error())` to see the full backtrace

summary(rlang::last_error())
#> <error>
#> * Message: "`offset` must be smaller than the number of columns"
#> * Class: `rlang_error`
#> * Fields: `message`, `trace` and `parent`
#> * Backtrace:
#> █
#> ├─global::myfunction()
#> │ └─mtcars %>% group_by(cyl) %>% select(last_col(Inf))
#> │   ├─base::withVisible(eval(quote(`_fseq`(`_lhs`)), env, env))
#> │   └─base::eval(quote(`_fseq`(`_lhs`)), env, env)
#> │     └─base::eval(quote(`_fseq`(`_lhs`)), env, env)
#> │       └─global::`_fseq`(`_lhs`)
#> │         └─magrittr::freduce(value, `_function_list`)
#> │           ├─base::withVisible(function_list[[k]](value))
#> │           └─function_list[[k]](value)
#> │             ├─dplyr::select(., last_col(Inf))
#> │             └─dplyr:::select.grouped_df(., last_col(Inf)) /Users/lionel/Dropbox/Projects/R/hadley/dplyr/R/manip.r:513:2
#> │               └─dplyr:::.select_grouped_df(.data, !!!quos(...), notify = TRUE) /Users/lionel/Dropbox/Projects/R/hadley/dplyr/R/grouped-df.r:212:2
#> │                 └─tidyselect::vars_select(names(.data), !!!quos(...)) /Users/lionel/Dropbox/Projects/R/hadley/dplyr/R/grouped-df.r:205:2
#> │                   └─tidyselect:::vars_select_eval(.vars, quos)
#> │                     └─purrr::map_if(ind_list, is_helper, eval_tidy)
#> │                       └─purrr::map(.x[sel], .f, ...)
#> │                         └─rlang:::.f(.x[[i]], ...)
#> └─tidyselect::last_col(Inf)
```

I added an unexported version of `call_print_type()` that provides a more rigorous breakdowns of calls. I'm still bothered by the current enum returned by `call_print_type()`. Should we unexport it for the time being? Alternatively, it could take a `type = "coarse"` argument that could be set to `"fine"` or `"strict"`.


R/fn.R
```diff
@@ -345,10 +345,15 @@ is_primitive_lazy <- function(x) {
...
+  if (is_primitive(fn)) {
+    return(ns_env("base"))
+  }
+
+  if(is_closure(fn)) {
+    return(environment(fn))
   }
-  environment(fn)
+
+  abort("`fn` is not a function")
```
hadley: Maybe put this in an `else` block, and then eliminate the explicit `return()`s

hadley:
I'm fine with leaving `call_print_type()` unexported for now.

lionel-:
Merged to move on with the PRs but let me know if you'd rather not have tho `global::` prefix.

jennybc:
I have a weak preference for `global::foo()`, in the absence of other cues to contemplate where each function is being found.

jennybc:
The real life example that got me involved in this in the first place featured two packages with internal functions with the same name, which I wasn't even considering. So I barked up the wrong tree for quite a while in my debugging. I can imagine someone doing this to themselves with the global environment.

PR #663, opened 2018-10-16
Allow subsetting .data with symbols

lionel-:
I noticed that lists and data frames can be subsetted with symbols. It makes sense to allow it for data pronouns as well, then we get:

```r
quoting_fn <- function(data, var) {
  var <- ensym(var)
  mutate(data, .data[[var]])
}
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -55,6 +55,10 @@ transform the warning to a failure.
   .data[["name"]])` and `group_by(df, name)` produce the same column
   name.
 
+* `.data[[` now supports symbols, just like you can use symbols to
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -55,6 +55,10 @@ transform the warning to a failure.
   .data[["name"]])` and `group_by(df, name)` produce the same column
   name.
 
+* `.data[[` now supports symbols, just like you can use symbols to
```
hadley: How do you subset a regular vector with a symbol?

hadley: That seems like a pretty weird quirk to recreate.

PR #664, opened 2018-10-16
Add `trim` argument to trace_back()

### LINKED ISSUE #491: calltrace should have strip = 1L argument
To replace current line:

```
trace <- trace[-length(trace)] # remove call to self
```

Making it easier to embed inside other functions that you don't want to capture on the trace.

### LINKED ISSUE #521: Should trace_back() trim one layer of sibling calls?
This would make it easier to use. Unlike `cst()` it's not meant to understand the call stack but to capture an useful backtrace. The following should all produce an empty trace:

```r
trace_back()
#> █

identity(trace_back())
#> █
#> └─identity(trace_back())

identity(identity(trace_back()))
#> █
#> ├─[ identity(...) ]
#> └─identity(trace_back())
```

lionel-:
~Includes #660 to avoid merge conflicts with the namespaced backtraces.~

Closes #491
Closes #521

### DIFF
--- a/R/trace.R
+++ b/R/trace.R
@@ -11,6 +11,11 @@
 #'   larger context, for example in tests or inside an RMarkdown
 #'   document where you don't want all of the knitr evaluation mechanisms
 #'   to appear in the backtrace.
+#' @param trim The number of parent contexts to trim from the bottom
+#'   of the backtrace, starting from the rightmost leaf (the newest
+#'   call frame in the tree). Contexts containing several frames
+#'   (these appear as sibling nodes in the backtrace tree) are fully
+#'   trimmed from the backtrace.
 #' @examples
 #' # Trim backtraces automatically (this improves the generated
 #' # documentation for the rlang website and the same trick can be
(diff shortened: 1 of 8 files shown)


tests/testthat/test-trace-collapse-eval.txt
```diff
@@ -4,17 +4,13 @@ Full:
   ├─base::eval(quote(g()))
   │ └─base::eval(quote(g()))
   └─rlang:::g()
```
hadley: Can you explain why these are now trimmed?

R/trace.R
```diff
@@ -11,6 +11,11 @@
 #'   larger context, for example in tests or inside an RMarkdown
 #'   document where you don't want all of the knitr evaluation mechanisms
 #'   to appear in the backtrace.
+#' @param trim The number of parent contexts to trim from the bottom
+#'   of the backtrace, starting from the rightmost leaf (the newest
+#'   call frame in the tree). Contexts containing several frames
+#'   (these appear as sibling nodes in the backtrace tree) are fully
+#'   trimmed from the backtrace.
```
hadley: It would be helpful if you could briefly explain when you'd want to set this, and how you find the correct value.

lionel-: I have renamed `to` to `top` and `trim` to `bottom`. Both arguments now take frame environments. This way it's more consistent and easier to document. I just mention that `bottom` can also be an integer passed to `caller_env()`, then we don't need to mention siblings. Documentation reviewed accordingly with a little more motivation.

tests/testthat/test-trace-collapse-eval.txt
lionel-: Because these are sibling nodes:

```
█
└─global::f()
  ├─base::eval(quote(g()))
  │ └─base::eval(quote(g()))
  └─global::g()
    ├─base::eval(quote(trace_back(e, 0)))
    │ └─base::eval(quote(trace_back(e, 0)))
    └─rlang::trace_back(e, 0)
```

I have set `bottom` to 0 to preserve the original test output.

PR #690, opened 2018-11-23
Update lifecycle API

lionel-:
* Move lifecycle badge generation and deprecation signallers in a self-contained `compat-lifecycle.R` file. Draft documentation. I haven't exported these functions yet so we can keep experimenting without constraints. To use this file, packages need rlang in their DESCRIPTION Imports but they don't need to import rlang.

* Change verbosity strategy. Instead of checking the package of the retired function is attached, we now check that the caller of the retired function inherits from the package being tested. The idea is that only people who can do something about the retired function should get a warning (the end user if called from the global env, or the package developer if called during tests).

  To implement this, I check the `TESTTHAT_PKG` environment variable. It isn't set with `devtools::test()` yet but there is an issue opened at https://github.com/r-lib/testthat/issues/787.

* The options have been reviewed. `lifecycle_disable_warnings` disables retirement warnings and `lifecycle_verbose_soft_deprecation` causes soft-deprecated functions to be treated like deprecated functions. A new option `lifecycle_repeat_warnings` causes repeated deprecation warnings instead of once per session.

  To ensure warnings in all cases, we need these three options:

  ```r
  with_options(
    lifecycle_disable_warnings = FALSE,
    lifecycle_verbose_soft_deprecation = TRUE,
    lifecycle_repeat_warnings = TRUE,
    ...
  )
  ```

  ~We could add a `lifecycle_force_warnings` option to set all three above, but it would be a bit confusing to have both this option and `lifecycle_disable_warnings`. It's probably better to add wrappers around `with_options()` etc instead, to get this effect.~

  Added helpers `scoped_lifecycle_warnings()` and `with_lifecycle_warnings()` to force warnings.


lionel-:
* Added `scoped_lifecycle_silence()` and `scoped_lifecycle_errors()` (and `with_` variants).
* Renamed `abort_defunct()` to `stop_defunct()`.

PR #704, opened 2018-12-18
Add entrace()

### LINKED ISSUE #699: Export rlang:::add_backtrace()
Possibly with a different name, so that you can do `option(error = rlang_backtrace)` to get nice rlang tracebacks everywhere.

### LINKED ISSUE #702: Tools to convert conditions to errors
Something like this: 

```R
library(rlang)

message2error <- function(code) {
  withCallingHandlers(code, message = function(cnd) {
    trace <- trace_back(bottom = sys.frame(-1))
    abort(cnd$message, trace = trace)
  })
}
```

(But generalised to work with user supplied condition name)

The advantage over `options(warn = 2)` is that you get a nicer back trace.

lionel-:
* Add `classes` parameter to `with_abort()`, similar to the new parameter in `catch_cnd()`. Closes #702.

* Rename `add_backtrace()` to `entrace()`. It now supports conditions as well, and so can be used both as a calling handler and as an `options(error = )` handler. `entrace()` now powers `with_abort()`. Closes #699.

* `entrace()` takes `top` and `bottom` arguments, passed to `trace_back()`. By default, it uses an improved detection mechanism for `bottom`. In particular, warnings promoted to errors should now properly be captured.


tests/testthat/test-with-abort.txt
```diff
@@ -46,8 +45,8 @@ fields:  `message`, `trace`, `parent` and `error`
 backtrace:
     █
  1. ├─rlang::with_abort(f())
- 2. │ └─base::withCallingHandlers(...)
- 3. └─rlang:::f()
- 4.   └─rlang:::g()
- 5.     └─rlang:::h()
- 6.       └─base::stop("Low-level message")
+ 2. │ ├─rlang::eval_bare(rlang::expr(withCallingHandlers(expr, !!!handlers)))
```
hadley: Maybe it's worth generating the expression in a separate line just to make this slightly cleaner?

tests/testthat/test-with-abort.txt
```diff
@@ -13,10 +13,9 @@ message: Low-level message
 class:   `rlang_error`
 backtrace:
  1. rlang::with_abort(f())
- 3. rlang:::f()
- 4. rlang:::g()
- 5. rlang:::h()
- 6. base::stop("Low-level message")
+ 4. rlang:::f()
+ 5. rlang:::g()
+ 6. rlang:::h()
```
hadley: Is it confusing to not include the final call to `stop()`? I could go either way.

lionel-: One advantage is that it is consistent with backtraces created by `abort()`.

tests/testthat/test-with-abort.txt
lionel-: Now also `.Call()`s into `Rf_eval()` to make that line disappear completely.

tests/testthat/test-with-abort.txt
hadley: I think I'd definitely prefer to see the error message in tree call tree

hadley: OTOH while `traceback()` shows them, neither `where` nor `recover` does

PR #705, opened 2018-12-20
Export `is_interactive()`

lionel-:


jennybc:
Wouldn't `rlang.interactive` and `local_interactive()` be more consistent with other option names and withr, respectively?

lionel-:
We use underscores for all separators in rlang and other packages. I agree some standardisation would be nice, but we don't have a definite convention AFAIK. rlang also uses `scoped_` instead of `local_` for other functions like `scoped_bindings()` or `scoped_options()`.

Issue #636, opened 2018-09-29
Deparsing API


lionel-:
Meta-issue about the deparsing API:

**Naming**

* I'm not sure why we duplicated the API with `expr_` and `quo_` functions. Isn't it simpler to just treat quosures as any other language type and deal with them in `expr_` functions as well?

* Perhaps relevant for naming: These functions are about pretty printing, while `expr_deparse()` is about transferring R objects as text. Though the latter also has pretty printing features such as omitting the last *n* elements of vectors.

* The naming of `expr_text()`, `expr_name()` etc generally suggests they are accessors. In reality they are actions (deparse variants).

* Are `expr_name()` and `quo_name()` misnomers? It seems it should only be used to provide default names for quoted arguments that might be complex expressions. I often see it used for transforming symbols to strings, which should better be done with `as_string()` or `as.character()` (the former works better on Windows with foreign characters).

    Perhaps `expr_pretty_name()` would be more suggestive that it's about default names?

**Features**

* Do we need a `_line()` variant? We are using `_name()` for backtraces, it seems we are really after line. It should try to deparse as much of an expression as possible while keeping it on a same line and within `peek_option("width")`. It should take a column position as argument to be substracted from width.

hadley:
Maybe `_abbr`, `_line`, `_full`

lionel-:
Squash quosures with the label variant but represent them with `^` with the other variants?

krlmlr:
I was searching for `quo_text()` but didn't find this issue. Good it's linked from the docs ;-)

Is it safe to use `quo_text()` for now?

lionel-:
If you need a multi-line deparser yes. It'll go through a long lifecycle of deprecation.

lionel-:
Discussion continues in #1173.

PR #706, opened 2018-12-20
Export `as_label()`

lionel-:
This is about a replacement for `quo_name()`. I found out that this function is misused by many users for transforming quoted symbols to strings. This is not appropriate because `quo_name()` is a lossy operation and no assumptions should be made about the resulting string. For instance, long symbols might be truncated in the future. The right operator for deterministic symbol-to-string conversion is `as_string()`.

I think the problem stems partly from the function name which suggests it is a getter for "variable names", rather than a lossy transformation. To fix this, the PR introduces a new function `as_label()` to replace `quo_name()`, which I hope will be more suggestive of its purpose. I have also improved the documentation to make it clearer what function to use and when.

This is also relevant to the new naming repair framework for data frames because labellisation is the first step before repairing names because `as_label()` is used internally for automatic naming of captured quosures. This is how unnamed columns in `tibble()` get their names for instance.

I would have liked a greater period of experimentation with this function name but it is likely I will need it for my RStudio conf talk.

Part of #636.


lionel-:
I just realised `as_string()` should be `as_name()`. I think this is the right name for it since people were already using `quo_name()` instead of `as_string()`. It is also more specific in the intention, and could gain argument suchs as `syntactic = ` in the future.

Thus users would have a choice between `as_label()` and `as_name()` for converting symbols to strings. I think they will naturally be drawn to the latter when they need to _refer_ to an object by name, and the former when they need to _represent_ an object e.g. in a plot.

lionel-:
`as_name()` is now implemented in https://github.com/r-lib/rlang/pull/712.

PR #712, opened 2019-01-03
Add `as_name()`

lionel-:
* I have reverted the automatic unwrapping of quosures by `as_string()`. This function is now a low-level symbol to string converter.

* The new `as_name()` function is a higher-level variant of `as_string()` which automatically unwraps closures. When rlang gains name-repairing features, `as_name()` should probably get a `syntactic` parameter.

Tidy eval users thus have the choice between `as_label()` to _represent_ an object as a short description, and `as_name()` to _refer_ to an object by its name.


R/deparse.R
```diff
@@ -868,3 +870,50 @@ as_label <- function(x) {
     }
   )
 }
+
+#' Extract names from symbols
+#'
+#' @description
```
hadley: I think you need to emphasise that `as_name()` is the _opposite_ of `as.name()`

Issue #735, opened 2019-03-04
Results of as_function() should inherit from "function"


richierocks:
Functions created using `as_function()` do not have a class that includes `"function"`, which means that `inherits(fn, "function")` confusingly returns `FALSE`.

``` r
library(rlang)
# An example function, created with as_function()
is_positive <- as_function(~ . > 0)
# It is a function, as expected
is.function(is_positive)
#> [1] TRUE
# It's also a lambda, as expected 
is_lambda(is_positive)
#> [1] TRUE
# It doesn't inherit from function, which isn't expected
inherits(is_positive, "function")
#> [1] FALSE
```

A possible fix would be to change this line

https://github.com/r-lib/rlang/blob/master/R/fn.R#L436

from

```r
structure(fn, class = "rlang_lambda_function")
```

to

```r
structure(fn, class = c("rlang_lambda_function", "function"))
```

lionel-:
Agreed, would you like to send a PR for this?

PR #736, opened 2019-03-04
as_function() return value inherits from function

richierocks:
closes https://github.com/r-lib/rlang/issues/735


lionel-:
Could you add a NEWS bullet referring to that issue and your github handle please? Oh and a unit test that checks rlang lambdas inherit from `function`.

lionel-:
Thanks!

PR #742, opened 2019-03-21
Add curly curly operator

lionel-:


lionel-:
`\alias{{{}` in an Rd file chokes the parser :/

> The only thing that might need discussion is {{ x }} vs {{x}}

I find the former much easier to read and it makes the interpolation stand out in wrapper functions.

PR #752, opened 2019-04-01
Add `r_on_exit_callback()`

lionel-:
`rlang_on_exit_callback` is an exported C callable that pushes an `on_exit()` event in the current call frame (typically the `.Call()` wrapper) to call back a C function with data.

The C library exposes it as `r_on_exit_callback()`.

It could be renamed to `r_on_exit()` if we find a better name for the version that takes R code to evaluate. In general it seems simpler to pass a C callback because you don't need to worry about scoping of the R code then, so the callback variant might end up more useful.

### DIFF
--- a/R/rlang.R
+++ b/R/rlang.R
@@ -17,3 +17,7 @@ is_same_body <- NULL
 .onUnload <- function(lib) {
   .Call(rlang_library_unload)
 }
+
+callback <- function(fn, data) {
+  .Call(rlang_callback, fn, data)
+}
--- a/src/Makevars
+++ b/src/Makevars
@@ -37,6 +37,7 @@ internal-files = \
         internal/expr-interp-rotate.c \
         internal/internal.c \
         internal/quo.c \
+        internal/stack.c \
         internal/utils.c
 
 export-files = \
--- a/src/export/exported-tests.c
+++ b/src/export/exported-tests.c
@@ -4,6 +4,17 @@ sexp* rlang_r_string(sexp* str) {
   return STRING_ELT(str, 0);
 }
 
+static void cb(void* data) {
+  const char* msg = (const char*) data;
+  r_inform(msg);
+}
+
+sexp* rlang_test_onexit() {
+  r_on_exit_callback(cb, "foo");
+  r_on_exit_callback(cb, "bar");
+  r_abort("tilt");
+}
+
 
 // chr.c
 
--- a/src/export/init.c
+++ b/src/export/init.c
@@ -112,6 +112,8 @@ extern sexp* rlang_env_inherits(sexp*, sexp*);
 extern sexp* rlang_eval_top(sexp*, sexp*);
 extern sexp* rlang_attrib(sexp*);
 extern sexp* rlang_named(sexp*, sexp*);
+extern sexp* rlang_callback(sexp*, sexp*);
+extern void rlang_on_exit_callback(void (*fn)(void*), void* data);
 
 // Library initialisation defined below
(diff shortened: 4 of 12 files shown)


gaborcsardi:
Yeah, makes sense to keep the data in R.

There are some practical issues, though.

One is that you would need to use `on.exit(..., after = FALSE)`, otherwise the API is not so convenient, because usually you want to run the exit handlers in reverse order. `after = FALSE` is only available from 3.5.0, so this is not compatible with older R versions.

Two, which is less important, is that if one exit handler throws for some reason, then the rest of them are not executed on older R versions (before R 3.5.0). This is probably not something we need to worry about, unless the exit handlers use the R API, which they should not, probably.

Three, calling back to `on.exit()` will be slower, just pushing the addresses in a C stack can even be macro, so much faster. But yeah, you can always collect your resources into a struct, so....

lionel-:
Good points about reverse order and backward compatibility. This can be worked around on earlier versions of R by managing our own stack of exit expression.

> Three, calling back to on.exit() will be slower, just pushing the addresses in a C stack can even be macro, so much faster.

Agreed but with the C stack you need to wrap each of them in `R_ExecWithCleanup()` in case one of them jumps, which causes some overhead as well. In general when performance is an issue, pooling resources to clean up in a single object might be the preferred way as you mention.

gaborcsardi:
I guess this is not a priority for you, but it would be nice to have a non-rlang implementation as well, which I can just copy into packages that are low level, e.g. processx.

gaborcsardi:
> This can be worked around on earlier versions of R by managing our own stack of exit expression.

Yeah, that's why I created the stack. And if you already maintain the stack, then you don't really need to call back to R...

But I like that the `on.exit()` solution does not need to know about native routine registration, etc. So I think this is the way forward, and we should create a non-rlang implementation that can be ultimately added to base R, assuming they are interested, which seems possible.

gaborcsardi:
Btw. are there any calls in C that create frames? E.g. `R_ExecWithCleanup()` and similar do, maybe? 

It might be fine if there are, but you need to be aware of these when you write the C code, i.e. that the exit handlers and not necessarily on the last `.Call()`.

lionel-:
Sure, it should be easy to rewrite as normal C API. Or processx could depend on rlang and pick up the C callable to avoid duplicating implementations, if that makes sense.

By the way if the goal is to include something like safecall in R, one possible implementation is to offer a `.Call2()` which creates a special context within which you can push callbacks before calling `do_dotcall()`. Maybe `CTXT_CCODE` can be reused for this. Then the stack of callbacks is stored in `RCNTXT.cdata` and invoked from `RCNTXT.cend`. Could also expose `R_ExecWithExit()` to create the context manually and push exits from normal `.Call()`. I think it makes sense to have `R_onExit()` in the C API, even if we can't use it right away because of compatibility.

> Btw. are there any calls in C that create frames?

I don't think so. For instance exec-with-cleanup creates a CTXT_CCODE frame, `on.exit()` only pushes to `CTXT_FUNCTION` frames.

gaborcsardi:
So, how about an email to R-devel or Luke, giving a couple of possible implementations, and asking if they would be happy with one of them in base?

And until 3.7 will be oldrel we can use rlang, or embed an implementation.

lionel-:
Sounds good.

lionel-:
If the patch gets accepted I will rewrite this PR as a drop-in compat file for `R_onExit()`.

lionel-:
Closing this since if favour of cleancall.

PR #759, opened 2019-04-12
Add pairlist2()

lionel-:
* New `pairlist2()` function to create pairlists with splicing support. It's mainly meant to support `call2()` but may also be useful for `new_function()`. It preserves empty arguments since those are useful in calls and formals lists.

* `call2()` is reimplemented in C and uses `pairlist2()` for capturing dots. This brings its performance closer to that of `base::call()`:

  ```r
  bench::mark(
    base = call("fn"),
    before = call2_before("fn"),
    after = call2("fn")
  )
  #>   expression     min    mean  median    max `itr/sec` mem_alloc  n_gc n_itr
  #>   <bch:expr> <bch:t> <bch:t> <bch:t> <bch:>     <dbl> <bch:byt> <dbl> <int>
  #> 1 base          48ns  76.8ns    59ns 14.7µs 13013347.        0B     0 10000
  #> 2 before      11.2µs  14.5µs  13.4µs 94.3µs    68865.        0B     3  9997
  #> 3 after        698ns 867.4ns   806ns 24.9µs  1152811.        0B     0 10000
  
  args <- list(foo = missing_arg(), bar = 2)
  bench::mark(
    base = as.call(c(list(as.name("fn")), args)),
    before = call2_before("fn", !!!args),
    after = call2("fn", !!!args)
  )
  #>   expression     min    mean  median     max `itr/sec` mem_alloc  n_gc n_itr
  #>   <bch:expr> <bch:t> <bch:t> <bch:t> <bch:t>     <dbl> <bch:byt> <dbl> <int>
  #> 1 base         777ns  1.02µs   892ns  15.2µs   982750.        0B     0 10000
  #> 2 before        13µs 16.76µs  15.7µs 104.7µs    59670.        0B     2  9998
  #> 3 after        1.4µs  1.74µs   1.6µs  28.5µs   573699.        0B     1  9999
  ```

  Like `pairlist2()`, it preserves empty arguments:

  ```
  call2("[", quote(x), , j = 2)
  #> x[, j = 2]
  ```

* `new_function()` is now implemented in C as well.


NEWS.md
```diff
@@ -1,6 +1,19 @@
 
 # rlang 0.3.4.9000
 
+* `call2()` now uses `pairlist2()` internally. This brings its
+  performance closer to that of `base::call()`. This also means it now
```
hadley: I think this oversells it a bit given your benchmarks - maybe just say "considerably improves performance"?

src/internal/call.c
```diff
@@ -0,0 +1,76 @@
...
+    return true;
+  default:
+    return false;
+  }
+}
+
+static sexp* rlang_call2(sexp* fn, sexp* args, sexp* ns) {
+  if (r_typeof(fn) == r_type_character) {
+    if (r_length(fn) != 1) {
+      r_abort("`.fn` must be a length 1 string");
```
hadley: Can we just say "must be a string"? (i.e. a string is a length 1 character vector)

src/internal/fn.c
```diff
@@ -0,0 +1,26 @@
...
+
+sexp* rlang_new_function(sexp* args, sexp* body, sexp* env) {
+  if (r_typeof(env) != r_type_environment) {
+    r_abort("`env` must be an environment");
+  }
+
+  args = KEEP(r_vec_coerce(args, r_type_pairlist));
+
+  sexp* node = args;
+  while (node != r_null) {
```
hadley: Why do you prefer a `while` here instead of a `for`? I like that the `for` bounds the scope of the iterator variable, and separates the iteration from the action.

PR #764, opened 2019-04-18
Emphasize `.x` (over `.`) in as_function() docs

jennybc:
Some recent discussions elsewhere have affirmed my sense that `.x` is better to emphasize in anonymous function, as opposed to `.`. Specifically, in dplyr pipelines, I think it's best if people associate `.` with the primary data frame that is being piped and use `.x` when writing a lambda in scoped verbs.

### DIFF
--- a/R/fn.R
+++ b/R/fn.R
@@ -379,9 +379,9 @@ fn_env <- function(fn) {
 #'
 #' \Sexpr[results=rd, stage=render]{rlang:::lifecycle("stable")}
 #'
-#' * `as_function()` transform objects to functions. It fetches
-#'   functions by name if supplied a string or transforms formulas to
-#'   function.
+#' * `as_function()` transforms an object into a function. It fetches
+#'   the function by name, for string input, or turns a one-sided formula
+#'   into a function.
 #'
 #' * `as_closure()` first passes its argument to `as_function()`. If
 #'   the result is a primitive function, it regularises it to a proper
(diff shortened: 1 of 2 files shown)


R/fn.R
```diff
@@ -379,9 +379,9 @@ fn_env <- function(fn) {
 #'
 #' \Sexpr[results=rd, stage=render]{rlang:::lifecycle("stable")}
 #'
-#' * `as_function()` transform objects to functions. It fetches
-#'   functions by name if supplied a string or transforms formulas to
-#'   function.
+#' * `as_function()` transforms an object into a function. It fetches
+#'   the function by name, for string input, or turns a one-sided formula
+#'   into a function.
```
jennybc: It seems better to write in the singular here.

R/fn.R
```diff
@@ -392,17 +392,25 @@ fn_env <- function(fn) {
 #'   If a **function**, it is used as is.
 #'
 #'   If a **formula**, e.g. `~ .x + 2`, it is converted to a function
-#'   with two arguments, `.x` or `.` and `.y`. This allows you to
+#'   with up to two arguments: `.x` (single argument) or `.x` and `.y`
+#'   (two arguments). The `.` placeholder can be used instead of `.x`.
+#'   This allows you to
```
jennybc: I did this awful line break to make the diff smaller. Also, I think you wrap at a different column than I do.

lionel-: I think it's ok to create a larger diff because of rewrapping. I usually wrap at 72 because this way two pages fit nicely on 13in laptop screens with a legible font size.

No worries though.

PR #765, opened 2019-04-23
Support functions and environments in `ns_env()` and `ns_env_name()`

lionel-:
And remove support for default argument.

### DIFF
--- a/R/env-special.R
+++ b/R/env-special.R
@@ -197,9 +197,12 @@ current_env <- function() {
 #' These functions are experimental and may not belong to the rlang
 #' package. Expect API changes.
 #'
-#' @param x The name of a package or an environment. If a function,
-#'   the environment of that function is searched for a namespace with
-#'   [base::topenv()].
+#' @param x
+#'   * For `ns_env()`, the name of a package or an environment as a
+#'     string.
+#'   * An environment.
+#'   * A function. In that case, the environment of that function is
+#'     searched for a namespace with [base::topenv()].
 #'
 #' @seealso [pkg_env()]
 #' @keywords internal
--- a/R/env-special.R
+++ b/R/env-special.R
@@ -200,7 +200,7 @@ current_env <- function() {
 #' @param x
 #'   * For `ns_env()`, the name of a package or an environment as a
 #'     string.
-#'   * An environment.
+#'   * An environment (the current environment by default).
 #'   * A function.
 #'
 #'   In the latter two cases, the environment ancestry is searched for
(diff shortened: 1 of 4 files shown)


R/env-special.R
```diff
@@ -197,44 +197,51 @@ current_env <- function() {
...
-    if (isNamespace(env)) {
-      env
-    } else {
-      NULL
-    }
-  } else {
-    asNamespace(pkg)
+  env <- switch(typeof(pkg),
+    builtin = ,
+    special =
```
hadley: It looks weird to indent the switch in this way — I'd do `character = if (is_string(pkg)) asNamespace(pkg)` so that every case can fit on one line.

R/env-special.R
```diff
@@ -197,44 +197,51 @@ current_env <- function() {
...
+#' @param env A namespace environment.
 #' @export
 ns_env_name <- function(pkg = NULL) {
-  if (is_null(pkg)) {
-    pkg <- with_env(caller_env(), ns_env())
-  } else if (is_function(pkg)) {
-    pkg <- get_env(pkg)
-  }
-  unname(getNamespaceName(pkg))
+  env <- switch(typeof(pkg),
```
hadley: Why not use `ns_env()` here?

R/env-special.R
```diff
@@ -197,44 +197,51 @@ current_env <- function() {
 #' These functions are experimental and may not belong to the rlang
 #' package. Expect API changes.
 #'
-#' @param pkg The name of a package. If `NULL`, the surrounding
-#'   namespace is returned, or an error is issued if not called within
-#'   a namespace. If a function, the enclosure of that function is
-#'   checked.
+#' @param pkg The name of a package or an environment. If a function,
```
hadley: Should we rename to `x`? 

I think a bulleted list would also make it more clear what the set of possible values is.

R/env-special.R
lionel-: I like this indent because it's a little more scalable, e.g. you can add `{ }` braces when a couple of expressions are needed and it's not worth moving them to a function.

Your style works here though.

R/env-special.R
lionel-: Because `ns_env()` has a different error message.

R/env-special.R
lionel-: Good idea.

lionel-:
@hadley With the last commit we also take the `topenv()` for environments, to be consistent with functions.

lionel-:
I restored the default to be the current environment. This way we don't break tibble which uses this. And actually the UI makes sense I think. The new implementation of default arguments is cleaner than before.

PR #787, opened 2019-06-12
Add support for weak references

wch:
This PR adds an R interface for weak references. I tried to use the conventions in rlang, but please let me know if things should be done differently.

I'm not sure if this is correct, but I used `@keywords experimental`.


R/weakref.R
```diff
@@ -0,0 +1,87 @@
...
+#'
+#' When the key becomes unreachable, the key and value in the weak reference
+#' object are replaced by `NULL`, and the finalizer is scheduled to execute.
+#'
+#' @param key The key for the weak reference. Must be a reference object -- that
+#'   is, an environment or external pointer.
+#' @param value The value for the weak reference. This can be `NULL`, if you
+#'   want to use the weak reference like a weak pointer.
+#' @param finalizer A function that is run after the key becomes unreachable.
+#' @param onexit Should the finalizer be run when R exits?
```
lionel-: "onexit" is evocative of `on.exit()` handlers. Maybe `ensure`? Should it be `TRUE` by default?

R/weakref.R
```diff
@@ -0,0 +1,87 @@
...
+#'
+#' @keywords experimental
+#' @seealso [is_weakref()], [weakref_get_key()] and [weakref_get_value()].
+#' @export
+#' @examples
+#' e <- new.env()
```
lionel-:
```suggestion
#' e <- env()
```

R/weakref.R
```diff
@@ -0,0 +1,87 @@
...
+#' # When the regular reference (the `e` binding) is removed and a GC occurs,
+#' # the weak reference will not keep the object alive.
+#' rm(e)
+#' gc()
+#' identical(weakref_get_key(w), NULL)
+#'
+#'
+#' # A weak reference with a key and value. The value contains data about the
+#' # key.
+#' k <- new.env()
```
lionel-:
```suggestion
#' k <- env()
```

R/weakref.R
```diff
@@ -0,0 +1,87 @@
...
+}
+
+#' Get key/value from a weak reference object
+#'
+#' @param x A weak reference object.
+#'
+#' @seealso [is_weakref()] and [new_weakref()].
+#'
+#' @export
+weakref_get_key <- function(x) {
```
lionel-: `wref_key()`?

R/weakref.R
```diff
@@ -0,0 +1,87 @@
...
+  .Call(rlang_weakref_get_key, x)
+}
+
+#' @rdname weakref_get_key
+#' @export
+weakref_get_value <- function(x) {
```
lionel-: `wref_value()`?

NAMESPACE
```diff
@@ -408,6 +409,7 @@ export(new_quosure)
 export(new_quosures)
 export(new_raw)
 export(new_raw_along)
+export(new_weakref)
```
lionel-: Also needs `NEWS` bullet.

R/weakref.R
wch: I was thinking of the `onexit` from `reg.finalizer()`, but if you think that's too confusing with the `on.exit()` function, I could change it. `ensure` seems a little weird to me though.

In my experience, it's rarely necessary to run a finalizer on exit, because I've used them for managing stuff that's all in R. But I don't feel strongly about the default either way.

lionel-: How about `on_quit`? This is consistent with `quit()`. (It needs the underscore so we don't introduce a new casing style in the UI.)

Keeping the default as `FALSE` sounds good.

wch: Sure, `on_quit` sounds good to me.

lionel-:
Thanks!

PR #806, opened 2019-07-05
Allow %|% to work when y is of same length as x

rcannood:
These changes allow `x %|% y` when `y` is of same length as `x`. For example:
```r
> c(1L, NA, 3L, NA, NA) %|% 0L # usual use case
[1] 1 0 3 0 0
> c(1L, NA, 3L, NA, NA) %|% (6L:10L) # modified use case
[1]  1  7  3  9 10
```

Changes: 
 * `R/operators.R`: I modified the documentation of %|% to reflect the changes in behaviour of this function. I also modified the `stopifnot()` checks accordingly.
 * `src/lib/replace-na.c`: Rather than looking at the length of `y` in `replace_na_()`, I created a separate function `replace_na_vec_()` and call the correct function depending on the length of `y`, as not to sacrifice speed by having more if/else statements.
 * `tests/testthat/test-operators.R`: I added tests to check whether `x %|% y` is working correctly when `length(y) == length(x)`, and also that it produces an error when it is not.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 
 # rlang 0.4.0.9000
 
+## Bugfixes and small improvements
+
+* `x %|% y` now also works when `y` is of same length as `x` (@rcannood, #806).
+
 # rlang 0.4.0
 
 ## Tidy evaluation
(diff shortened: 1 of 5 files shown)


lionel-:
Nice, thanks! Can you add a NEWS item please? It should end with:

> (@rcannood, #806).

rcannood:
Done! :slightly_smiling_face:

lionel-:
Thanks!

Please note that all vector functions will eventually move to the vctrs or funs packages.

PR #809, opened 2019-07-09
changed outdated doc reference to fn_fmls

burchill:
The `is_function()` documentation refers to `fn_fmls()` as a "function that returns a representation of formal arguments for primitive functions", which is no longer true as of rlang v 0.4.0 (but was true for rlang v0.3.1). `fn_fmls()` currently throw an error for primitives. 

From the `fn_fmls()` doc page: 

> Unlike `formals()`, these helpers throw an error with primitive functions instead of returning NULL.

As a side note, I'm a little salty about this loss of functionality. Does `rlang` now not have a way of accessing primitive function arguments?

### DIFF
--- a/R/fn.R
+++ b/R/fn.R
@@ -224,15 +224,13 @@ fn_body_node <- function(fn) {
 #' partial matching, etc). One practical consequence of the special
 #' way in which primitives are passed arguments is that they
 #' technically do not have formal arguments, and [formals()] will
-#' return `NULL` if called on a primitive function. See [fn_fmls()]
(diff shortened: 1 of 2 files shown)


lionel-:
Thanks!

> As a side note, I'm a little salty about this loss of functionality. Does rlang now not have a way of accessing primitive function arguments?

It seems safer to fail and force the caller to handle primitive functions specially. The goal of taking formals is likely meta-programming, and primitives are tricky in terms of argument matching.

Note that rlang is not a stable package, except for a few core functionality like tidy eval, `abort()`, etc.

PR #817, opened 2019-07-30
Simplify branch backtraces further

lionel-:
This increases the simplification of branch backtraces by keeping track of lexical contexts of local closures. If a call frame A is hosted by a closure that was created on the stack at frame B, the latter frame is treated as the parent of A.

The main effect of this change is to merge all frames implicated in the invokation of the local closure, as well as the frame of the closure itself. It's as if the code was run from the function that created the closure:

```r
f <- function() lapply(NA, function(x) trace_back())
trace <- f()[[1]]

trace
#>     █
#>  1. └─global::f()
#>  2.   └─base::lapply(NA, function(x) trace_back())
#>  3.     └─FUN(X[[i]], ...)

print(trace, simplify = "branch")
#>  1. global::f()

```

This behaviour is recursive:

```r
f <- function() lapply(NA, function(x) lapply(NA, function(y) trace_back()))
trace <- f()[[1]][[1]]

trace
#>     █
#>  1. └─global::f()
#>  2.   └─base::lapply(NA, function(x) lapply(NA, function(y) trace_back()))
#>  3.     └─FUN(X[[i]], ...)
#>  4.       └─base::lapply(NA, function(y) trace_back())
#>  5.         └─FUN(X[[i]], ...)

print(trace, simplify = "branch")
#>  1. global::f()
```

This also applies to named local closures:

```r
f <- function() {
  g <- function(x) {
    h <- function(y) trace_back()
    h()
  }
  g()
}
trace <- f()

trace
#>     █
#>  1. └─global::f()
#>  2.   └─g()
#>  3.     └─h()

print(trace, simplify = "branch")
#>  1. global::f()
```

I think ideally we'd preserve the call frames of named closures, however R does not allow distinguishing between named and anonymous functions.

This automatically simplifies magrittr backtraces in a range of cases, though not all of them. I have removed the complex, difficult to maintain and buggy unrolling code in favour a simpler frame trimming to take care of the remaining cases of complex magrittr backtraces. This way we can experiment with terser backtraces, and if unrolling turns out better we could tackle it as part of tidyverse/magrittr#175.

### DIFF
--- a/R/trace.R
+++ b/R/trace.R
@@ -524,6 +524,17 @@ is_eval_call <- function(call) {
   is_call(call, c("eval", "evalq"), ns = c("", "base"))
 }
 
+lexical_parent <- function(parents, id) {
+  parent <- parents[[id]]
+
+  # `eval()` frame cause XXX
+  if (is.na(parent) || parent >= id) {
+    NA
+  } else {
+    parent
+  }
+}
+
 trace_simplify_branch <- function(trace) {
   parents <- trace$parents
   lexical_parents <- trace$clo_parents
@@ -543,7 +554,7 @@ trace_simplify_branch <- function(trace) {
       id <- next_id
     }
 
-    while (!is.na(lexical_parent <- lexical_parents[[id]])) {
+    while (!is.na(lexical_parent <- lexical_parent(lexical_parents, id))) {
       id <- lexical_parent
     }
 
--- a/R/trace.R
+++ b/R/trace.R
@@ -82,7 +82,6 @@ trace_back <- function(top = NULL, bottom = NULL) {
   calls <- as.list(sys.calls()[idx])
 
   calls <- map(calls, call_fix_car)
-  calls <- add_pipe_pointer(calls, frames)
   calls <- map2(calls, seq_along(calls), maybe_add_namespace)
 
   parents <- normalise_parents(parents)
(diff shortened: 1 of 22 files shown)


hadley:
@jennybc could you please take a look at this? I'm worried that we might be simplifying too much, so that it'll harder for people to map back to what we see in the code.

jennybc:
Just saw this (behind on my gh notifications). Sure I will test drive.

jennybc:
Just from reading @lionel-'s narrative (which is super helpful) I do agree with this:

> I'm worried that we might be simplifying too much, so that it'll harder for people to map back to what we see in the code.

Of course, everyone want the "perfect" amount of info, but I think it's better to err on the side of too much than too little.

That said, the idea of eliminating extraneous info in the important special case of magrittr is very exciting. It might be interesting to add some of these examples here:

> This automatically simplifies magrittr backtraces in a range of cases, though not all of them.

lionel-:
I think the main downside is that even in the case of `lapply()` / `map()` we lose precious information, for instance line 2 is important here:

```r
 1. global::f()
 2. purrr::map(1:3, ~abort("foo"))
 3. .f(.x[[i]], ...)
```

Also it's probably surprising to have the structure of the backtrace change when the lambda is extracted into a named function.

Regarding magrittr, it is already taken care of in current rlang (although with complex and buggy code), but it was interesting to see some of the heavy backtraces vanish on their own with this scheme.

lionel-:
Likely a bad idea.

PR #826, opened 2019-08-20
Fix typo

krlmlr:
in `?entrace` .


krlmlr:
`add_backtrace()` is still in the index, do we still need it there?

lionel-:
Good catch, I don't think so.

PR #828, opened 2019-08-23
Use `value` not `data[[nm]]` in default binder

### LINKED ISSUE #827: `env_bind()` performance fix
I believe this is related to: https://github.com/r-lib/tidyselect/issues/111

`env_bind_impl()` has a default `binder` function as:
https://github.com/r-lib/rlang/blob/09fbc8618bb0a886d2d4546e8cbd378b1c178f65/R/env-binding.R#L139

The `data[[nm]]` result there should really just be `value`. This has dreadful performance implications as is, but still works.

I'll submit a PR

DavisVaughan:
Closes #827

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
 
 # rlang (development)
 
+* `env_bind()`'s performance has been significantly improved by fixing a bug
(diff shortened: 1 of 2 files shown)


lionel-:
Thank you!

PR #834, opened 2019-09-13
Document caveat when using quasiquotation with as_function()

### LINKED ISSUE #833: Corner case: ensym() with dot-dot argument ..1 from ellipsis
Do we expect `ensym(..1)` to return the same as `ensyms(...)[[1]]`, similarly for `enquo()` and `enexpr()` ? (Should we update `as_function()` to use mustaches?)

``` r
library(rlang)

f <- function(...) ensym(..1)
f(a)
#> Error in (function (x) : object '..1' not found

f <- function(...) ensyms(...)[[1]]
f(a)
#> a

f <- function(x, .x = {{ x }}) ensym(.x)
f(a)
#> a

f <- function(..., .x = {{ ..1 }}) ensym(.x)
f(a)
#> Error in (function (x) : object '..1' not found
```

<sup>Created on 2019-09-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

krlmlr:
Closes #833.

### DIFF
--- a/R/fn.R
+++ b/R/fn.R
@@ -378,9 +378,13 @@ fn_env <- function(fn) {
 #'   If a **formula**, e.g. `~ .x + 2`, it is converted to a function
 #'   with up to two arguments: `.x` (single argument) or `.x` and `.y`
 #'   (two arguments). The `.` placeholder can be used instead of `.x`.
-#'   This allows you to create very compact anonymous functions with up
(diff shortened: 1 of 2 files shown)


lionel-:
Thanks.

The other thing is that it doesn't really make sense to use quasiquotation with the first argument of lambdas, if that's a lambda that is being mapped. The first argument is always forced in that case, and that's like getting an unquoted object.

PR #842, opened 2019-10-02
Stop formatting rlang errors from `conditionMessage()` method

lionel-:
This changes the strategy for formatting rlang errors and saving unhandled errors to `rlang::last_error()`. Currently all the logic is in the `conditionMessage()` method for rlang errors. This has several drawbacks:

* It is harder to implement a condition-message method for subclasses of rlang-error.

* Some functions like `try()` invoke the condition-message method, which causes backtraces (or reminders to call `last_error()`) to leak unexpectedly.

This PR moves the logic inside `abort()` with a new strategy, thanks to an idea from @gaborcsardi:

* To detect unhandled errors, we first `signalCondition()` the error object. If the error is handled this will cause a jump out of `abort()`. This can be overridden by setting the `rlang_force_unhandled_error` global option to `TRUE` (useful for unit tests).

* If we're still in `abort()` after signalling the condition, we save the error and format the message.

* To avoid logging errors multiple times from calling handlers, we `stop()` with a bare condition object that inherit only from `"condition"`.


gaborcsardi:
@hadley thanks!

PR #847, opened 2019-10-11
Add `cnd_message()`, `cnd_issue()`, and `cnd_bullets()`

lionel-:
`cnd_message()` is a function that assembles an error message from two generics, `cnd_issue()` and `cnd_bullets()`. Methods for `cnd_issue()` should return a single line. Methods for `cnd_bullets()` should return a character vector of lines.

The bullets are prefixed with `*`, `x`, or `i` depending on the names of the character vector.

- Elements named `"i"` are prefixed with a blue "info" symbol.
- Elements named `"x"` are prefixed with a red "cross" symbol.
- Unnamed elements are prefixed with a "*" symbol.

`cnd_message()` is called automatically by the `rlang_error` method for `conditionMessage()`, so that error classes only need to implement these two methods. The default methods are equivalent to just plucking `cnd$message`, so they can be left unimplemented as well.

If the error object contains a `.bullets` field containing a function, that method is called instead. This makes it possible to write very precise methods for generating the error bullets without having to overspecify the class hierarchy.


lionel-:
Example of usage at r-lib/vctrs#613.

lionel-:
One nice perk of this approach is that it is easy to override the `cnd_issue()` method for a whole classes of errors. For instance tibble and tidyselect can catch all `vctrs_error_index` errors and rethrow a subclass for which the `cnd_issue()` method mentions data frames and columns. The bullets parts are independent and don't need to be changed as long as they are written in a generic fashion.

PR #850, opened 2019-10-16
Don't recreate active bindings in `env_clone()`

lionel-:


lionel-:
cc @ltierney and @kalibera. A point release of rlang is already planned so this should make it to CRAN soon.

PR #858, opened 2019-10-29
Switch to dynamic dots terminology

lionel-:


R/call.R
```diff
@@ -22,7 +22,7 @@
 #' @param .fn Function to call. Must be a callable object: a string,
 #'   symbol, call, or a function.
 #' @param ... Arguments to the call either in or out of a list. These dots
-#'   support [tidy dots][tidy-dots] features. Empty arguments are preserved.
+#'   are [dynamic][dyn-dots] and preserve empty arguments.
```
hadley: Maybe a little more context: These are [dynamic dots][dyn-dots] which support `!!!` and `:=`, and preserve empty arguments.

R/call.R
```diff
@@ -99,12 +99,12 @@ call2 <- function(.fn, ..., .ns = NULL) {
 }
 #' Create pairlists with splicing support
 #'
-#' This pairlist constructor supports [tidy dots][tidy-dots] features
-#' like `!!!`. Use it to manually create argument lists for calls or
-#' parameter lists for functions.
+#' This pairlist constructor supports [dynamic dots][dyn-dots]. Use
```
hadley:
```suggestion
#' This pairlist constructor uses [dynamic dots][dyn-dots]. Use
```

R/dots.R
```diff
@@ -1,26 +1,82 @@
...
+#'   along the way to function parameters.
 #'
-#' - You can unquote names by using the [unquote][quasiquotation]
-#'   operator `!!` on the left-hand side of `:=`.
+#' - __Collect__ arguments inside data structures, e.g. with [c()] or
+#'   [list()].
 #'
-#' - Trailing commas are ignored, making it easier to copy and paste
-#'   arguments.
+#' Dynamic dots offer a few more additional features:
```
hadley:
```suggestion
#' Dynamic dots offer a few additional features:
```

lionel-:
> I don't like "these dots are dynamic" because it implies you could say "this arg is dynamic

I thought it would be fine to say a quasiquoted argument is dynamic since you can force part of it with runtime values, though we wouldn't use this description ourselves.

R/call.R
lionel-: Would this be just for rlang? I think in dplyr or tidyverse packages it might be best to leave off all the scary looking terms or names.

lionel-:
We now use a clickable `[Dynamic]` button to describe the semantics.

dyn-dots

It's quite nice, maybe we could use other buttons, for example to describe named dots arguments as pairs of keys and values, like `[Named, Dynamic]`, or `[UniquelyNamed, Dynamic]`.

lionel-:
Maybe also a button to indicate data variables are accepted?

```
var [DataMasked] - The variable to group by.
... [Dynamic, DataMasked] - Variables summarised with `mean()`.
```

R/call.R
```diff
@@ -21,8 +21,8 @@
 #'
 #' @param .fn Function to call. Must be a callable object: a string,
 #'   symbol, call, or a function.
-#' @param ... Arguments to the call either in or out of a list. These dots
-#'   support [tidy dots][tidy-dots] features. Empty arguments are preserved.
+#' @param ... [[Dynamic][dyn-dots]] - Arguments for the function
```
hadley: I'd be tempted by something like:

```R
#' @param ... <[dynamic][dyn-dots]> Arguments for the function call...
```

lionel-: That does look nice as well:

dyn-dots2

I have a slight preference for the square brackets, which one do you prefer?

hadley: I prefer `<dynamic>` because it looks more like a type specification (which it kind of is), and I think at the level of documentation reader trying to make the distinction more clear will result in more net confusion.

I'm imagining we'll eventually get to something like this:

```R
#' @param df <data.frame()>
#' @param x <in_mask, vector()>
#' @param flag <logical(1)> 
#' @param name <character(1)>
```

Where the goal is not to fully capture the exact type, but give hints to the educated user in a form that they can internalise.

hadley: Maybe the way to think of it is not exactly as a type specification, but as a set of modifiers that imply something about how each argument is interpreted.

hadley: OTOH since we write types as prototypes, and prototypes can be expressed as function calls, maybe we should consider the same approach for data docs? OTOOH maybe the lack of `()` would help to distinguish between types and other modifiers.

lionel-: I agree we don't need to make the distinction clear in general purpose documentation. However it's still a worthwhile distinction to maintain in general. "Dynamic" refers to a language effect and describes how values are produced. "Type" refers to the set of values accepted by the function. One is about syntax, the other about semantics. Together they describe the _kind_ of inputs accepted by a function (e.g. what kind of R code the user is allowed to type), which is why it makes sense to treat them similarly at a high level.

I was leaning towards `[` because it looks like a button. From an aesthetic point of view I also like having CamelCase prose for topic descriptions, instead of code-based snake_case. One issue with code is that it's not clear how to convey type parameters such as size, and it's not clear whether constructors are uniform enough to make sure they are easy to parse by pressed users. Maybe we should use type abbreviations instead?

Could something like this also work?

```r
@param ... [Dynamic, DataMasked, Recycled, Type:lgl, Size:nrow] Variables to filter by.
```

lionel-: With specifiers like `Recycled`, we not only have a description of the kind of accepted inputs, but also a description of input validation or processing. Coercion semantics would also be of this nature.

hadley: Let's do a poll to decide this

lionel-: (After having considered what `nse-force` would look like in CamelCase I'm no longer attracted to that casing for topics...)

lionel-: Now using angular brackets:

```r
#> @param ... <[dynamic][dyn-dots]> A list of named attributes.
```

PR #859, opened 2019-10-31
Switch to "force" terminology

lionel-:


R/quasiquotation.R
```diff
@@ -1,61 +1,80 @@
...
-#'   unquoted in place.
+#'   ```
+#'   vars <- syms(c("height", "mass"))
+#'
+#'   # Force-splicing is equivalent to supplying the elements separately
+#'   starwars %>% select(!!!vars)
+#'   starwars %>% select(height, mass)
+#'   ```
+#'
+#' - The curly-curly operator `{{ }}` for function arguments is a bit
```
hadley: I'd suggest something like: this is useful when you have a data-variable in a env-variable (i.e. a function argument)

PR #860, opened 2019-10-31
Switch to "defuse" terminology

lionel-:


R/quotation.R
```diff
@@ -1,109 +1,95 @@
...
 #'
-#' Quotation is a mechanism by which an expression supplied as
-#' argument is captured by a function. Instead of seeing the value of
-#' the argument, the function sees the recipe (the R code) to make
-#' that value. This is possible because R [expressions][is_expr] are
-#' representable as regular objects in R:
+#' When a function argument is defused, R doesn't return its value
+#' like it normally would. Instead, it returns the R expression
+#' describing how to make the value as an inert object. These defused
+#' expressions are like blueprints for computing values.
```
hadley: I think you need a parenthetical here saying something like "(This is also known as quoting, and is what is done by `substitute()`"), i.e. something to sign post to the more experience reader that this is the content that they're looking for.

R/quotation.R
```diff
@@ -1,109 +1,95 @@
...
 #'
-#' * Calls represent the action of calling a function to
-#'   compute a new value. Evaluating a call causes that value to be
-#'   computed. Calls typically involve symbols to reference R objects.
+#' The main purpose of preventing evaluation of an expression is to
```
hadley: Is that really the main purpose?

R/quotation.R
```diff
@@ -1,109 +1,95 @@
...
-#' We call objects containing calls and symbols [expressions][is_expr].
-#' There are two ways to create R expressions. First you can **build**
-#' calls and symbols from parts and pieces (see [sym()], [syms()] and
-#' [call2()]). The other way is by *quotation* or *quasiquotation*,
-#' i.e. by intercepting an expression instead of evaluating it.
 #'
+#' @section Types of defused expressions:
 #'
-#' @section User expressions versus your expressions:
+#' * __Calls__ represent the action of calling a function to compute a
```
hadley:
```suggestion
#' * __Calls__, like `f(1, 2, 3)`, `1 + 1`, or ...,  represent the action of calling a function to compute a
```

R/quotation.R
```diff
@@ -1,109 +1,95 @@
...
+#'   new value, such as a vector.
 #'
-#' There are two points of view when it comes to capturing an
-#' expression:
+#' * __Symbols__ represent named objects. When the object pointed to
```
hadley:
```suggestion
#' * __Symbols__, like `x` or `df`, represent named objects. When the object pointed to
```

R/quotation.R
lionel-: Data masking isn't the main purpose?

hadley: I think you need to rewrite the sentence to be more like: "The main purpose of defusing evaluation to allow data-masking, where an expression is evaluated in the context of a data frame so that you can do `var` instead of `df$var`.

PR #865, opened 2019-11-12
Print messages to `stdout` by default

### LINKED ISSUE #852: Should `rlang::inform()` use `stdout()` by default?
One issue with `base::message()` is that it outputs to `stderr()`, which while it works great for terminal usage, in RStudio and in the Windows GUI causes the output to be in red.

I just did a [workaround for this in vroom](https://github.com/r-lib/vroom/commit/55a94bb9dfd18eac30c4771f3c48dc6af6c55f86#diff-90a167c269c93fd663f1c3c690147343R460-R472). Maybe we should do this by default in `rlang::inform()`? 

Currently some packages (such as usethis) use `cat()` for user messaging to avoid the red text issue, but they really should use message conditions, so users can use `suppressMessages()` to suppress the messages.

lionel-:
Closes #852

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,12 @@
 
 # rlang 0.4.2
 
+* `inform()` now prints messages to the standard output by default in
+  interactive sessions. This makes them appear more like normal output
+  in IDEs such as RStudio. In non-interactive sessions, messages are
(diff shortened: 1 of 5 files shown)


jimhester:
LGTM, thanks!

PR #870, opened 2019-11-19
Fix condition message generics

lionel-:
* `cnd_message()` now assembles the message based on three generics `cnd_header()`, `cnd_body()`, and `cnd_footer()`.

* `cnd_body()` can be overridden by storing a `body` field in the condition object.

* `cnd_issue()` and `cnd_bullets()` are removed from the package. `format_bullets()` is kept as an experimental function.

* If a condition object contains a `bullets` field, this overrides `cnd_body()` similarly to the `body` field. Unlike `body`, the bullets are formatted with `format_bullets()`. Supplying both `body` and `bullets` is an error.


lionel-:
on second thought the `bullets` field thing doesn't seem useful since it's just:

```r
abort("header", body = ~ format_bullets(bullets))
```

I'll remove it.

NEWS.md
```diff
@@ -1,8 +1,15 @@
 
-# rlang (development version)
+# rlang 0.4.2
 
+* New `cnd_header()`, `cnd_body()` and `cnd_footer()` generics. These
+  are automatically called by the default `conditionMessage()` for
```
hadley: Maybe called by `conditionMessage.rlang_error()`?

R/cnd-message.R
```diff
@@ -1,121 +1,126 @@
...
-#' `cnd_message()`, according to the following scheme:
 #'
-#' - Elements named `"i"` are prefixed with a blue "info" symbol.
-#' - Elements named `"x"` are prefixed with a red "cross" symbol.
-#' - Unnamed elements are prefixed with a "*" symbol.
+#' @section Overriding `cnd_body()`:
 #'
-#' While you are free to lay out the bullets in the order that you
-#' like, "x" bullets should usually precede "i" bullets.
+#' Sometimes the contents of an error message depends on the state of
```
hadley: This idea slightly freaks me out, but it seems to makes sense.

R/cnd-message.R
```diff
@@ -1,121 +1,126 @@
...
 #'
 #' @description
 #'
-#' \Sexpr[results=rd, stage=render]{rlang:::lifecycle("experimental")}
-#'
-#' `cnd_message()` assembles an error message from two components:
+#' `cnd_message()` assembles an error message from three generics:
 #'
-#' - The `cnd_issue()` generic. Methods should return a single line.
+#' - `cnd_header()`
```
hadley: Should probably define exactly what these should return

R/cnd-message.R
```diff
@@ -1,121 +1,126 @@
...
+#' - `cnd_body()`
+#' - `cnd_footer()`
 #'
-#' - The `cnd_bullets()` generic. Methods should return a named vector
-#'   of lines. These lines are automatically prefixed with a bullet by
-#'   `cnd_message()` (see the section on error statements).
+#' The default method for the error header returns the `message` field
+#' of the condition object. The default methods for the body and
+#' footer return empty character vectors.
```
hadley: Oh, I thought the default footer inserted the ``Run `rlang::last_error()`...`

R/cnd-message.R
```diff
@@ -1,121 +1,126 @@
...
+  chr()
+}
+
+#' Format bullets for error messages
+#'
+#' @description
+#'
+#' \Sexpr[results=rd, stage=render]{rlang:::lifecycle("experimental")}
+#'
+#' `format_bullets()` takes a character vector and returns a single
```
hadley: Maybe `format_error_bullets()`?

NEWS.md
lionel-: Now:

> These are automatically called by `conditionMessage.rlang_error()`, the default method for all rlang errors.

R/cnd-message.R
lionel-: This overriding was how we could achieve lazy errors + complicated type checking + shallow error hierachy in vctrs.

In general I think the idea of extending S3 to be truly OO (i.e. the object itself decides its implementation, rather than the lexical scoping or the subclassing hierarchy) is worth exploring.

R/cnd-message.R
lionel-: I think that's a good idea but we should try it after the CRAN release because changing the mechanism of how this reminder is displayed is a bit tricky (what happens in knitted document etc).

R/cnd-message.R
lionel-: This section now starts with an `experimental` badge.

PR #881, opened 2020-01-03
update is_call description

TylerGrantSmith:
Removed the final sentence in the description of `is_call` which makes reference to nonexistent functions  `is_unary_call()` and `is_binary_call()`.


lionel-:
Thanks!

PR #882, opened 2020-01-04
Add glue unquoting

lionel-:


lionel-:
@jimhester Can you also take a look at `glue_unquote()` please?

R/rlang.R
```diff
@@ -64,8 +64,15 @@ base_pkg_env <- NULL
     is_same_body <<- is_reference
   }
 
+  # glue depends on rlang so we can't detect it is installed right
+  # away. Instead we do it when glue has finished loading.
+  setHook(
```
hadley: Does this work if glue has is already loaded?

NEWS.md
```diff
@@ -1,6 +1,45 @@
 
 # rlang (development version)
 
+* You can now use glue syntax to unquote on the LHS of `:=`. This
```
hadley: Maybe mention that if you use this in a package you'll also need to import glue?

R/nse-force.R
```diff
@@ -84,25 +84,42 @@
...
+#' ```
+#'
 #' Currently, forcing names with `:=` only works in top level
 #' expressions. These are all valid:
 #'
 #' ```
-#' exprs(!!nm := x)
-#' tibble(!!nm := x)
-#' list2(!!nm := x)
+#' exprs(!!name := x)
```
hadley: Maybe remove this example now?

R/rlang.R
lionel-: oh I confused myself into thinking that glue depended on rlang, but it doesn't.

I have extracted `defer_if_unloaded()` from the downstream dep check routine. It should work without side effects in all cases. Better use this function no matter the dependency structure, to avoid mistakes.

R/rlang.R
```diff
@@ -5,18 +5,24 @@ NULL
 is_same_body <- NULL
 
 
+defer_if_unloaded <- function(pkg, fn, action = NULL) {
```
hadley: I was expecting something more like this:

```R
on_pkg_load <- function(pkg, code) {
  if (isNamespaceLoaded(pkg)) {
    code
  } else {
    f <- function() code
    setHook(packageEvent(pkg, "onLoad"), f)
  }
}
```

I find that a bit easier to reason about, rather than a hook to re-call the current function plus a non-local return.

jimhester:
Is there any performance impact in strings without any glue markers? If I read the code correctly we unconditionally call `glue()` twice if the glue package is loaded for all LHS strings.

Also do we want to issue an informative error if people try to use this syntax and _don't_ have glue loaded? This seems an easy mistake to make in a package and it might not be immediately obvious to people what is wrong.

A cheap scan for `{` not followed by another `{` first in the string might be a way to handle both issues.

lionel-:
Thanks @jimhester. I'm not too worried about performance in this case, but I think it's a good idea to perform a first pass so we can throw an informative error when glue isn't there and a `{` is detected.

One breaking change here is that `{` must now be escaped twice so if you had `"foo{bar}"` you now need `"foo{{{{bar}}}}"` (or use a symbol rather than a string, or unquote the string). So it seems we can throw an error whenever we see `{` and glue is not installed.

PR #889, opened 2020-01-13
Use signal() instead of cnd_signal()

### LINKED ISSUE #887: A wrong example on catch_cnd()
Just happened to see this on https://rlang.r-lib.org/reference/catch_cnd.html. Maybe this should be something like `catch_cnd(signal(message = "a condition", class = "my_condition"))`?

```r
catch_cnd(cnd_signal("my_condition", .msg = "a condition"))
#> <simpleError in cnd_signal("my_condition", .msg = "a condition"): unused argument (.msg = "a condition")>
```

yutannihilation:
Fix #887


lionel-:
Thanks!

PR #896, opened 2020-01-17
Return invisibly from inform()

krlmlr:
Also for `warn()`?


lionel-:
Thanks! I guess `warn()` is ok because of `warning()`. It would be difficult to test for its return value without causing CMD check warnings.

PR #911, opened 2020-01-30
Be more explicit about what happens if knitr, an RStudio notebook or testthat is in progress

Bisaloo:
This is just a suggestion since it wasn't immediately clear to me what would happen in this case. I had to test it to understand.


lionel-:
Thanks!

PR #913, opened 2020-01-31
Suggest typo fixes in `arg_match()`

### LINKED ISSUE #798: Suggest candidate in case of small fuzzy match in `arg_match()`
Currently `arg_match()` suggest candidate value only if there is a partial match. However, if values are quite long, input mismatch can occur because of a typo inside a word. For example:

``` r
input <- "continuuos"
rlang::arg_match(input, c("discrete", "continuous"))
#> `input` must be one of "discrete" or "continuous"
```

<sup>Created on 2019-07-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I think it would be helpful in this case to have 'Did you mean "continuous"?' suggestion.

This can be achieved by first checking for partial matching with `pmatch()` (as is currently done), and in case of a failure check fuzzy matching with `agrep()`.

jonkeane:
Fixes #798


jonkeane:
I noticed a few warnings about partial matching the `regex` arg on eg https://github.com/r-lib/rlang/pull/913/files#diff-b9d4c08ada30b6dd5ba55b928f867f85R7 I'm happy to clean those up, but didn't want to confuse any issues with the actual fix.

hadley:
@jonkeane since you're touching that file anyway, it would be fine to fix the partial arg warnings too.

tests/testthat/test-arg.R
```diff
@@ -17,6 +17,14 @@ test_that("informative error message on partial match", {
   )
 })
 
+test_that("informative error message on a typo", {
+  myarg <- "continuuos"
+  expect_error(
+    arg_match(myarg, c("discrete", "continuous")),
```
hadley: I think you could inline `myarg` here?

R/arg.R
```diff
@@ -52,6 +53,14 @@ arg_match <- function(arg, values = NULL) {
     i_partial <- pmatch(arg, values)
     if (!is_na(i_partial)) {
       candidate <- values[[i_partial]]
+    }
+
+    i_close <- adist(arg, values) / nchar(values)
```
hadley: I'd like to see a few more tests on short strings — I'd like to see how this deals with false positives too.

tests/testthat/test-arg.R
```diff
@@ -17,6 +17,14 @@ test_that("informative error message on partial match", {
   )
 })
 
+test_that("informative error message on a typo", {
+  myarg <- "continuuos"
+  expect_error(
```
hadley: I _think_ it would be a little nice to use `verify_output()` here, and then provide quite a few tests.

tests/testthat/test-arg.R
jonkeane: For posterity, inlining returns `Error: Internal error: `arg_match()` expects a symbol`

R/arg.R
jonkeane: I've added a few more, as I was running them it looked (to me) like my original threshold of 0.3 was too strict for short strings (though 0.5 might be too lenient for longer ones).

tests/testthat/test-arg.R
jonkeane: TIL about `verify_output()`

jonkeane:
Changes pushed

R/arg.R
hadley: I'd wonder if just using a constant (e.g. edit distance of 3) might be better? Would you mind trying that out?

jonkeane: Yeah, I just tried that out, and things get a bit funky with small strings:

```
> myarg <- "a"
> arg_match(myarg, c("b", "c"))
Error: `myarg` must be one of "b" or "c"
Did you mean "b"?
```

```
> myarg <- "zz"
> arg_match(myarg, c("aa", "dd"))
Error: `myarg` must be one of "aa" or "dd"
Did you mean "aa"?
```

We could also disable the typo-checking for short strings with a constant, but that might not be better than the ratio. At the risk of overcomplicating: a max distance + a ratio might be a strike a balance here (something like `any(constant <= 3 & ratio <= 0.5)`).

hadley: Ok, lets stick with what you have and we can tweak it in the future, if needed.

PR #915, opened 2020-02-01
Add centralized documentation for global rlang options.

### LINKED ISSUE #899: Centralized documentation for options that affect rlang behaviour
Is there a central place to find all of rlang's options and what they do?

This is motivated by my recent journey into options that affect back traces (re: knitr, reprex, etc.).

If not, can there be? And would it be a good issue for Tidyverse dev day?

I made a [help topic like this for gargle](https://gargle.r-lib.org/reference/gargle_options.html). I realize options might play a different role here in rlang, but maybe the basic idea is still sound.

smingerson:
Added a centralized doc location for global options, accessible through `?rlang::rlang_options`. It is pretty brief, and points to more detailed documentation, if it exists. 

Fixes #899


R/rlang-options.R
```diff
@@ -0,0 +1,20 @@
+#' Global options for rlang
+#'
+#' @description
+#' rlang has several options which may be set globally to control
+#'  behavior. A brief description of each is given here. If any functions
```
lionel-: Can you remove the extra leading space please?

R/rlang-options.R
```diff
@@ -0,0 +1,20 @@
...
+#'  are referenced, refer to their documentation for additional details.
+#'
+#' * rlang_interactive: A logical value used by [is_interactive()]. This
```
lionel-: The option should be backticked so they appear like code.

R/rlang-options.R
```diff
@@ -0,0 +1,20 @@
...
+#'    can be set to `TRUE` to test interactive behavior in unit tests,
+#'    for example.
+#' * rlang_backtrace_on_error: A character string which controls whether
```
lionel-: Needs newlines between bullet items to improve readability.

R/rlang-options.R
```diff
@@ -0,0 +1,20 @@
...
+#'    backtraces are displayed with error messages, and the level of
+#'    detail they print. See [rlang_backtrace_on_error] for the possible option values.
+#' * rlang_trace_format_srcrefs: A logical value used to control whether
+#'    srcrefs are printed as part of the backtrace.
+#' * rlang_trace_top_env: An environment which will be treated as the
+#'    top-level environment when printing traces. See [trace_back()]
+#'    for examples.
+#' @name rlang_options
```
lionel-: I think the topic could be called `faq-options`, this will ensure it's included in the pkgdown reference along other faq items.

smingerson: I've left the file as `rlang-options.R`, but am happy to change that as well.

lionel-: Thanks, `faq.R` might be appropriate.

smingerson: The rlang pkgdown does not have a FAQ section, though it does have  "session state", which includes `matches("option")`. I had thought that had been the section you were referring to. 

Would you like me to update the pkgdown so it does have a FAQ section, similar to what [tidyselect](https://tidyselect.r-lib.org/reference/index.html#section-faq) has?

lionel-: That'd be great, thanks!

_pkgdown.yml
```diff
@@ -1,277 +1,281 @@
...
+      - new_data_mask
+      - as_data_pronoun
+  - title: Condition handlers
+    contents:
+      - with_handlers
+      - calling
+  - title: Session state
+    contents:
+      - matches("interactive")
+      - matches("^[^(faq)]option")
```
smingerson: That regex doesn't actually do what I want. I think the `matches("option")` should be changed to `ends_with("option")` to pick up all of the options documentation without faq-options. Unless we want the global options to be documented under both session state and FAQ

lionel-: Yes it seems fine to document in both sections.

Can you double check your changes to the pkgdown file please? It seems you have reindented everything. The commit should not contain any whitespace changes.

smingerson: All fixed up, sorry about that.

lionel-:
That's great, thanks!

PR #917, opened 2020-02-03
Repair arg_match failure

jonocarroll:
#914 introduced a check which appears to be too strict - on debug I see that the ordering of the values can be changed. This PR appears to resolve a lot of issues (related: https://github.com/r-lib/vctrs/issues/790) I've been seeing.

Fixes https://github.com/r-lib/vctrs/issues/790

### DIFF
--- a/R/arg.R
+++ b/R/arg.R
@@ -42,7 +42,7 @@ arg_match <- function(arg, values = NULL) {
   if (!is_character(arg)) {
     abort(paste0(chr_quoted(arg_nm), " must be a character vector"))
   }
-  if (length(arg) > 1 && !identical(arg, values)) {
+  if (length(arg) > 1 && !setequal(arg, values)) {
     abort(paste0(chr_quoted(arg_nm), " must be a character vector of length 1."))
   }
 
--- a/R/arg.R
+++ b/R/arg.R
@@ -43,7 +43,7 @@ arg_match <- function(arg, values = NULL) {
     abort(paste0(chr_quoted(arg_nm), " must be a character vector"))
   }
   if (length(arg) > 1 && !setequal(arg, values)) {
-    abort(paste0(chr_quoted(arg_nm), " must be a character vector of length 1."))
+    abort(paste0(chr_quoted(arg_nm), " must be a character vector of length 1 or setwise-match caller values."))
   }
 
   arg <- arg[[1]]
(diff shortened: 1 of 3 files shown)


DavisVaughan:
I'm wondering if, rather than making `arg_match()` more flexible, we should have `tibble:::repaired_names()` call `arg_match()` first and then pass the result on to `vec_as_names()`.

https://github.com/tidyverse/tibble/blob/10a25b9a1447901b469364f39081bfa15a7f270e/R/repair-names.R#L7

The core problem here is that `tibble::tibble()` uses a different ordering of the inlined name repair values than `vec_as_names()` does, but then it relies on the `arg_match()` call in `vec_as_names()`, when I think it should really use its own

CC @krlmlr

lionel-:
I tend to agree with @DavisVaughan. We had a similar problem in #914. @AliciaSchep and I concluded it's better to arg-match early.

We could always relax this later if needed. Worth noting that CRAN tibble doesn't use vctrs yet, so it won't be affected by the new `arg_match()` checks.

lionel-:
I've now come around on this. I don't think there's much advantage to being this strict, and it does makes it harder to wrap around arg-matching functions with different defaults. The additional strictness doesn't seem worth the friction it brings, since users are unlikely to write typos or wrong arguments this way. So I'm inclined to merge this PR. What do you think @DavisVaughan?

DavisVaughan:
That sounds fine!

lionel-:
@jonocarroll Could you add a unit test for this please?

jonocarroll:
Can do.

jonocarroll:
I've altered the error message when `!setequal(arg, values)` to help indicate that this may be what is not happening.

Tests to cover both positive and negative cases.

lionel-:
hmm I think it's better to leave the original error message. Mentioning the semantics for the default feels like a distraction that might confuse users.

lionel-:
I have changed it so that it now mentions the valid possible strings.

lionel-:
Thanks!

By the way I have squashed the PR. Since you've been developing from your `master` branch, this means the upstream master and your local master have now diverged. You can fix it by using `git fetch upstream && git reset --hard upstream/master` from your master branch (assuming the upstream rlang repo is called `upstream`).

jonocarroll:
No worries. I'm not working on any `rlang` changes outside of this so I'll just burn down my fork :fire: but thank you.

PR #930, opened 2020-02-27
Implement `set_names()` in C

DavisVaughan:
To do this, I've had to implement a few other things in C:

- `is_function()`

   - While I was there, I went ahead and did `is_closure()`, `is_primitive()`, `is_primitive_eager()` and `is_primitive_lazy()` too

- `is_formula()`

- `names2()`

I've created a new file, `internal/attr.c` to house the C implementations of `set_names()` and `names2()`.

For `names2()`, we fall back to the R level `names()` when `x` is an object to allow it to dispatch. We also handle pairlists and language objects specially, like `getAttrib()` does.

For `set_names()`, we always eventually set the names with `names<-()` rather than with `r_poke_names()`. The reason for this is that `do_namesgets()` has the ability to shallow copy _attributes_ using an ALTREP wrapper with `R_shallow_duplicate_attr()`. This isn't exposed to us yet from C, but calling `names<-` and getting a wrapper object is way faster than calling `r_maybe_duplicate()`, which will duplicate `x`.

I have allowed the new `r_as_function()` to call out to `rlang::as_function()` for simplicity. I have seen that you worked on a partial implementation of a C `as_function()` in vctrs, but the quosure handling here looks tricky to me. Feel free to take over and implement that if you want 😬 

```r
library(rlang)

x <- 1:50000 + 0L
a <- rep("x", 50000L)
```

```r
bench::mark(set_names(x, a), iterations = 100000)

# master
#> # A tibble: 1 x 6
#>   expression           min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>      <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 set_names(x, a)    6.7µs   9.32µs   105179.     121KB     72.6

# this PR
#> # A tibble: 1 x 6
#>   expression           min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>      <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 set_names(x, a)   1.03µs    1.5µs   649682.    7.81KB     110.
```

```r
bench::mark(is_function(1))

# master
#> # A tibble: 1 x 6
#>   expression          min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>     <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 is_function(1)   1.81µs   2.14µs   452624.      18KB     90.5

# this PR
#> # A tibble: 1 x 6
#>   expression          min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>     <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 is_function(1)    264ns    354ns  2491092.    3.86KB        0
```

### DIFF
--- a/R/attr.R
+++ b/R/attr.R
@@ -167,7 +167,7 @@ has_name <- function(x, name) {
 #' set_names(head(mtcars), paste0, "_foo")
 set_names <- function(x, nm = x, ...) {
   mold <- x
-  .External2(rlang_set_names, x, mold, nm)
+  .Call(rlang_set_names, x, mold, nm, environment())
 }
 
 #' Get names of a vector
--- a/R/attr.R
+++ b/R/attr.R
@@ -195,7 +195,7 @@ set_names <- function(x, nm = x, ...) {
 #' x <- set_names(1:3, c("a", NA, "b"))
 #' names2(x)
 names2 <- function(x) {
-  .Call(rlang_names2, x)
+  .Call(rlang_names2, x, environment())
 }
 
 # Avoids `NA` names on subset-assign with unnamed vectors
--- a/R/attr.R
+++ b/R/attr.R
@@ -166,37 +166,8 @@ has_name <- function(x, name) {
 #' # `...` is passed to the function:
 #' set_names(head(mtcars), paste0, "_foo")
 set_names <- function(x, nm = x, ...) {
-  set_names_impl(x, x, nm, ...)
-}
-
-# FIXME: This can be simplified once the `_along` ctors are defunct
-set_names_impl <- function(x, mold, nm, ...) {
-  if (!is_vector(x)) {
-    abort("`x` must be a vector")
-  }
-
-  if (is_function(nm) || is_formula(nm)) {
-    if (is_null(names(mold))) {
-      mold <- as.character(mold)
-    } else {
-      mold <- names2(mold)
-    }
-    nm <- as_function(nm)
-    nm <- nm(mold, ...)
-  } else if (!is_null(nm)) {
-    if (dots_n(...)) {
-      nm <- as.character(c(nm, ...))
-    } else {
-      nm <- as.character(nm)
-    }
-  }
-
-  if (!is_null(nm) && !is_character(nm, length(x))) {
-    abort("`nm` must be `NULL` or a character vector the same length as `x`")
-  }
-
-  names(x) <- nm
-  x
+  mold <- x
+  .External2(rlang_set_names, x, mold, nm)
 }
 
 #' Get names of a vector
@@ -224,15 +195,7 @@ set_names_impl <- function(x, mold, nm, ...) {
 #' x <- set_names(1:3, c("a", NA, "b"))
 #' names2(x)
 names2 <- function(x) {
-  if (typeof(x) == "environment") {
-    abort("Use `env_names()` for environments.")
-  }
-  nms <- names(x)
-  if (is_null(nms)) {
-    rep("", length(x))
-  } else {
-    nms %|% ""
-  }
+  .Call(rlang_names2, x)
 }
 
 # Avoids `NA` names on subset-assign with unnamed vectors
--- a/src/internal/attr.c
+++ b/src/internal/attr.c
@@ -66,14 +66,8 @@ static inline sexp* r_as_character(sexp* x);
 static inline sexp* r_as_function(sexp* x);
 static inline sexp* r_set_names_dispatch(sexp* x, sexp* nm);
 
-sexp* rlang_set_names(sexp* call, sexp* op, sexp* args, sexp* env) {
+sexp* rlang_set_names(sexp* x, sexp* mold, sexp* nm, sexp* env) {
   int n_kept = 0;
-
-  args = r_node_cdr(args);
-
-  sexp* x = KEEP_N(r_node_car(args), n_kept); args = r_node_cdr(args);
-  sexp* mold = KEEP_N(r_node_car(args), n_kept); args = r_node_cdr(args);
-  sexp* nm = KEEP_N(r_node_car(args), n_kept); args = r_node_cdr(args);
 
   sexp* dots = KEEP_N(rlang_dots(env), n_kept);
 
--- a/src/internal/attr.c
+++ b/src/internal/attr.c
@@ -1,0 +1,189 @@
+#include <rlang.h>
+#include "internal.h"
+
+static inline sexp* r_language_names(sexp* x);
+static inline sexp* r_names_dispatch(sexp* x);
+
+sexp* rlang_names2(sexp* x) {
+  const enum r_type type = r_typeof(x);
+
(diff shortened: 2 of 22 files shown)


DavisVaughan:
I'm not sure I understand the errors. I seem to be able to reproduce the one in `test-fn.R` line 273 if I run that test locally (even on master). If I run a full `test()` it doesn't show up, but if I run it manually it does. Odd

src/internal/attr.c
```diff
@@ -0,0 +1,189 @@
...
+    r_chr_fill(nms, r_empty_str, n);
+  } else {
+    nms = KEEP(rlang_replace_na(nms, r_shared_empty_chr));
+  }
+
+  FREE(2);
+  return nms;
+}
+
+static inline sexp* r_language_names(sexp* x) {
```
lionel-: Can you rename to `r_node_names()` please?

src/internal/attr.c
```diff
@@ -0,0 +1,189 @@
...
+  sexp** p_out = STRING_PTR(out);
+
+  int i = 0;
+
+  for(; x != r_null; x = r_node_cdr(x), ++i) {
+    sexp* tag = r_node_tag(x);
+
+    if (tag == r_null) {
+      p_out[i] = r_empty_str;
+    } else if (r_is_symbolic(tag)) {
```
lionel-: `r_is_symbolic()` returns true for either symbols and calls. This should be `r_typeof(x) == r_type_symbol`.

But I think this branch and the next one can be removed and you can assume a `TAG` is either `NULL` or a symbol. It's an internal error otherwise, like taking the CDR of a non-node object.

src/internal/attr.c
```diff
@@ -0,0 +1,189 @@
...
+      p_out[i] = PRINTNAME(tag);
+    } else {
+      r_abort("Invalid type %s for tag", r_type_as_c_string(r_typeof(tag)));
+    }
+  }
+
+  FREE(1);
+  r_mark_shared(out);
```
lionel-: Why is this marked shared?

src/internal/attr.c
```diff
@@ -0,0 +1,189 @@
...
+  FREE(n_kept);
+  return x;
+}
+
+
+static inline sexp* r_eval_fn_with_x(sexp* fn, sexp* x) {
+  sexp* args = KEEP(r_new_node(x, r_null));
+
+  sexp* call = KEEP(r_new_call(fn, args));
+
```
lionel-:
```suggestion
```

src/internal/attr.c
```diff
@@ -0,0 +1,189 @@
...
+  FREE(2);
+  return out;
+}
+
+static inline sexp* r_eval_fn_with_x_y(sexp* fn, sexp* x, sexp* y) {
+  sexp* args = KEEP(r_new_node(y, r_null));
+  args = KEEP(r_new_node(x, args));
+
+  sexp* call = KEEP(r_new_call(fn, args));
+
```
lionel-:
```suggestion
```

src/internal/attr.c
```diff
@@ -0,0 +1,189 @@
...
+// TODO: Replace with C implementation of `as_function()`
+static sexp* as_function_fn = NULL;
+static inline sexp* r_as_function(sexp* x) {
+  return r_eval_fn_with_x(as_function_fn, x);
+}
+
+// Use `names<-()` rather than setting names directly with `r_poke_names()`
+// for genericity and for speed. `names<-()` can shallow duplicate `x`'s
+// attributes using ALTREP wrappers, which is not in R's public API.
```
lionel-: Nice!

src/internal/attr.c
```diff
@@ -0,0 +1,189 @@
...
+
+  sexp* call = KEEP(r_new_call(fn, args));
+
+  sexp* out = r_eval(call, r_global_env);
+
+  FREE(3);
+  return out;
+}
+
+static inline sexp* r_eval_fn_with_x_dots(sexp* fn, sexp* x, sexp* dots) {
```
lionel-: * This function has slightly different semantics than other `r_eval_with_` functions which mask the symbols. Can you mask `fn` and `x` please?

* It should take the eval env as argument.

* You can move this function to the lib with the other variants.

src/internal/attr.c
```diff
@@ -0,0 +1,189 @@
...
+
+  sexp* call = KEEP(r_new_call(fn, args));
+
+  sexp* out = r_eval(call, r_global_env);
+
+  FREE(2);
+  return out;
+}
+
+static inline sexp* r_eval_fn_with_x_y(sexp* fn, sexp* x, sexp* y) {
```
lionel-: Same remark as below.

R/attr.R
```diff
@@ -166,37 +166,8 @@ has_name <- function(x, name) {
...
-  }
-
-  if (!is_null(nm) && !is_character(nm, length(x))) {
-    abort("`nm` must be `NULL` or a character vector the same length as `x`")
-  }
-
-  names(x) <- nm
-  x
+  mold <- x
+  .External2(rlang_set_names, x, mold, nm)
```
lionel-: Can we use `.Call(..., environment())` instead? I now prefer avoid using External2. Especially since it can lead to surprising visibility behaviour when we evaluate R code from the C routine, which is the case here.

src/internal/attr.c
DavisVaughan: The `is_symbol` check came from looking at what base R did, but ill remove
https://github.com/wch/r-source/blob/3d8c9fcaa927eedf0ec77738b1492030aebf75eb/src/main/attrib.c#L124-L145

src/internal/attr.c
DavisVaughan: Same reasoning as above, looking at what base R does
https://github.com/wch/r-source/blob/3d8c9fcaa927eedf0ec77738b1492030aebf75eb/src/main/attrib.c#L142

DavisVaughan: Based on this description of `PRINTNAME()`, my guess is that their reasoning is that this directly gives you the name node of the symbol, without any duplication, so the marking as shared makes it so that you can't accidentally overwrite that? Wouldn't it make more sense for the underlying CHARSXP node to be the thing that is marked as shared? I'm not sure

https://github.com/wch/r-source/blob/355079cacf64365cb0e1e785a9051a4660be273d/doc/manual/R-ints.texi#L433-L438

lionel-: `PRINTNAME()` returns the `CHARSXP` wrapped by the symbol. I don't think the contents of the character vectors explain the decision to mark it as shared.

The only explanation I can find is that it prevents the caller from thinking they can mutate the names and have it reflected in the pairlist.

DavisVaughan: I've left it in for now, but feel free to remove

PR #934, opened 2020-03-02
Simplify Makevars

DavisVaughan:
Do you need to keep track of the files like this? This simple version actually seems to work


DavisVaughan:
Ooh, it looks like you don't do this to actually make it compile, you do it to ensure that when you change something in `lib/<file.c>`, it recompiles the `lib.c` file

PR #941, opened 2020-03-11
Autogenerate rlang version

krlmlr:
for checks between R package and DLL mismatch.

Are `grep` and `cut` available everywhere?


lionel-:
I would prefer to try this in another package first to check if this is portable enough.

krlmlr:
Fair enough, I'll link here.

lionel-:
I think it would be better to do it automatically from the usethis helpers when there exists a `version.h` file in `src`.

krlmlr:
I'm not sure, your call.

krlmlr:
I'm not sure if this is the responsibility of a `./configure` script or of `Makevars`, it can work either way. Hard-coding the logic in an external helper adds a dependency that's hard to track.

lionel-:
This is now handled by `usethis::use_version()` which updates version.c.

PR #977, opened 2020-05-18
Update docs around `class` (vs `.subclass`)

jennybc:
I think there are places in the docs that suggest `type` --> `.subclass` is the most recent change. But in fact there is also `.subclass` --> `class`. I'm not entirely sure I have the wording right, so feel free to refine this PR or close and make the changes separately. But I am pretty sure this is an improvement.


lionel-:
Thanks!

PR #984, opened 2020-05-28
Add an example for each highlighted feature of dynamic dots

jennybc:


R/dots.R
```diff
@@ -33,6 +33,23 @@
...
 #' @name dyn-dots
 #' @aliases tidy-dots
+#'
+#' @examples
+#' f <- function(...) {
+#'   out <- list2(...)
+#'   rev(out)
+#' }
+#'
+#' # splice
```
lionel-, line 43: Comments need to be sentenced cased.

R/dots.R
```diff
@@ -33,6 +33,23 @@
...
+#' x <- list(alpha = "first", omega = "last")
+#' f(!!!x)
+#'
+#' # unquote a name
+#' nm <- "key"
+#' f(!!nm := "value")
```
lionel-, line 49: Actually this should now probably be `"{nm}" := "value"`? Perhaps add an example with a constant prefix/suffix too?

PR #991, opened 2020-06-08
Vectorize `is_installed()`

salim-b:
I'd like to have a vectorized `is_installed()` so I just went ahead and created this PR.

### DIFF
--- a/R/env-special.R
+++ b/R/env-special.R
@@ -270,17 +270,14 @@ is_namespace <- function(x) {
 #' If installed, the packages will be loaded but not attached.
 #'
 #' @param pkg The package names.
-#' @return A named logical vector of the same length as `pkg`. `TRUE` for each
-#'   installed package and `FALSE` for each package not found in any of the
-#'   libraries.
+#' @return `TRUE` if _all_ package names provided in `pkg` are installed,
+#'   `FALSE` otherwise.
 #' @export
 #' @examples
-#' is_installed(c("graphics", "utils"))
-#' is_installed("ggplot5")
+#' is_installed("utils")
+#' is_installed(c("base", "ggplot5"))
 is_installed <- function(pkg) {
-  vapply(pkg,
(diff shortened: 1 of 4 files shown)


lionel-:
I think all `is_` functions should return a single non-missing `TRUE` or `FALSE` value, i.e. they should be safe to use in `if ()` conditions.

salim-b:
Ok. But before this change, `is_installed()` silently tested the first element of the `pkg` argument only. So it always returned `TRUE` if the _first_ package was installed regardless of the remaining ones:

```r
rlang::is_installed(pkg = c("base", "no.notarealpackage"))
#> [1] TRUE
```

<sup>Created on 2020-06-08 by the <a href="https://reprex.tidyverse.org">reprex package</a> (v0.3.0)</sup>

Would it be fine to reduce the return value to a logical scalar using `all()`?

lionel-:
I think that makes sense.

R/env-special.R
```diff
@@ -264,19 +264,23 @@ is_namespace <- function(x) {
...
+#'   installed package and `FALSE` for each package not found in any of the
+#'   libraries.
 #' @export
 #' @examples
-#' is_installed("utils")
+#' is_installed(c("graphics", "utils"))
 #' is_installed("ggplot5")
 is_installed <- function(pkg) {
-  is_true(requireNamespace(pkg, quietly = TRUE))
+  vapply(pkg,
```
lionel-, line 281: Can you use `map_lgl()` and follow the local style please?

salim-b: done

salim-b:
PR updated :)

lionel-:
Can you also add a NEWS bullet with your handle please?

salim-b:
> Can you also add a NEWS bullet with your handle please?

Ok, done.

lionel-:
Thanks!

PR #998, opened 2020-06-14
Extract arg_match0()

### LINKED ISSUE #997: Faster version of arg_match()
because it's very slow. Example: more than half of the overhead in `vctrs::vec_as_location2()` comes from `arg_match()`.

Downstream: https://github.com/tidyverse/tibble/issues/780.

## vctrs dev

``` r
library(vctrs)
library(magrittr)

bench::mark(
  vec_as_location2("x", 1L, "x", missing = "propagate", arg = "x"),
  iterations = 1000
) %>%
  dplyr::select(-expression)
#> # A tibble: 1 x 5
#>        min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1   78.7µs   88.8µs     9451.     322KB     18.9
```

<sup>Created on 2020-06-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

## https://github.com/r-lib/vctrs/pull/1145

``` r
library(vctrs)
library(magrittr)

bench::mark(
  vec_as_location2("x", 1L, "x", missing = "propagate", arg = "x"),
  iterations = 1000
) %>%
  dplyr::select(-expression)
#> # A tibble: 1 x 5
#>        min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1   25.7µs   34.9µs    26573.     172KB     26.6
```

<sup>Created on 2020-06-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

krlmlr:
The idea is to offer two variants: `arg_match()` has the magic (we could even deprecate the `values` argument), `arg_match_val()` is bare-bones and fast (one call to `match()`, one check).

## Comparison

``` r
library(rlang)

foo <- function(x) {
  arg_match(x, c("arg1", "arg2"))
}

bar <- function(x) {
  arg_match_val(x, c("arg1", "arg2"))
}

bench::mark(foo("arg1"), bar("arg1"))
#> # A tibble: 2 x 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 foo("arg1")  10.49µs  12.81µs    71594.     142KB     28.6
#> 2 bar("arg1")   1.84µs   2.06µs   422295.        0B     42.2
```

<sup>Created on 2020-06-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

## rlang dev + vctrs dev

``` r
library(vctrs)
library(magrittr)

bench::mark(
  vec_as_location2("x", 1L, "x", missing = "propagate", arg = "x"),
  iterations = 1000
) %>%
  dplyr::select(-expression)
#> # A tibble: 1 x 5
#>        min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1   78.7µs   88.8µs     9451.     322KB     18.9
```

<sup>Created on 2020-06-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

## This PR + https://github.com/r-lib/vctrs/pull/1146

``` r
library(vctrs)
library(magrittr)

bench::mark(
  vec_as_location2("x", 1L, "x", missing = "propagate", arg = "x"),
  iterations = 1000
) %>%
  dplyr::select(-expression)
#> # A tibble: 1 x 5
#>        min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1   34.8µs   44.1µs    22065.     232KB     22.1
```

<sup>Created on 2020-06-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Happy to tweak as necessary.

Fixes #997.


R/arg.R
```diff
@@ -47,9 +47,23 @@ arg_match <- function(arg, values = NULL) {
...
+  arg_match_val(arg, values, arg_nm)
+}
+
+#' @description
+#' `arg_match_val()` is a bare-bones version if performance is at a premium.
+#' It requires a string as `arg` and explicit `values`.
+#' @param arg_nm The label to be used for `arg` in error messages.
+#' @rdname arg_match
+#' @export
+arg_match_val <- function(arg, values, arg_nm = as_label(substitute(arg))) {
```
lionel-, line 59: The `as_label(substitute(arg))` will cause lots of weird error messages.

krlmlr: Do you have an example?

lionel-: Internal calls and internal symbols (even worse if the internal call is complex). I guess it's going to be rare to call this from internal helpers though, so I agree that this default will work well most of the time. I think I like the idea of a bare bones variant, but then it makes sense to have all arguments mandatory? Probably should be called `arg_match0()`? This way we have a general suffix for naming bare bones variants.

krlmlr: I would not make `arg_nm` mandatory. It's bare-bones in the sense that it's fast, not that we need to repeat ourselves in the code that uses it if we don't have to.

krlmlr:
Let's do another iteration. I deprecated the `values` argument to `arg_match()`, we can discuss.

R/arg.R
```diff
@@ -47,9 +46,31 @@ arg_match <- function(arg, values = NULL) {
...
+#' @examples
+#'
+#' # Use the bare-bones version with explicit values for speed:
+#' arg_match0("bar", c("foo", "bar", "baz"))
+arg_match0 <- function(arg, values, arg_nm = as_label(substitute(arg))) {
+  # Use primitive checks to avoid function calls
+  if (!.Call(rlang_is_string, arg, NULL)) {
+    abort(paste0(chr_quoted(arg_nm), " must be a string."))
+  }
+  if (!.Call(rlang_is_character, arg, NULL)) {
```
lionel-, line 67: Is this tested? This should be `values`. If we're bypassing the public API this definitely needs tests (but I'd prefer to use the proper function calls as mentioned above).

krlmlr:
Good catch.

There's another problem with the current implementation: we need to allow `identical(arg, values)` even in the bare-bones version. 

We're paying a lot for the two function calls, see below. I wouldn't mind doing it in C altogether, the error handling would remain in R:

- Convenient C interface for use by rlang, vctrs and other packages
- Correct UI
- As fast as it can get

What do you think?

## `.Call()`

``` r
library(rlang)
letters
#>  [1] "a" "b" "c" "d" "e" "f" "g" "h" "i" "j" "k" "l" "m" "n" "o" "p" "q" "r" "s"
#> [20] "t" "u" "v" "w" "x" "y" "z"
bench::mark(arg_match0("c", letters), iterations = 10000)
#> # A tibble: 1 x 6
#>   expression                    min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>               <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 arg_match0("c", letters)   1.27µs   1.47µs   553756.    47.3KB        0
```

<sup>Created on 2020-06-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


## `is_*()`

``` r
library(rlang)
letters
#>  [1] "a" "b" "c" "d" "e" "f" "g" "h" "i" "j" "k" "l" "m" "n" "o" "p" "q" "r" "s"
#> [20] "t" "u" "v" "w" "x" "y" "z"
bench::mark(arg_match0("c", letters), iterations = 10000)
#> # A tibble: 1 x 6
#>   expression                    min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>               <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 arg_match0("c", letters)   2.15µs   2.43µs   316031.    52.1KB        0
```

<sup>Created on 2020-06-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lionel-:
Agreed, once we're reaching for micro-optimisations at the R level I think it's time to switch to C. (This is the original reason I kept `is_null()` as a closure before we aliased it to the primitive.)

krlmlr:
Now implemented in C, includes #1011. Let's decide on that before reviewing here.

R/arg.R
```diff
@@ -35,9 +35,8 @@ arg_match <- function(arg, values = NULL) {
     fn <- caller_fn()
     values <- fn_fmls(fn)[[arg_nm]]
     values <- eval_bare(values, get_env(fn))
-  }
-  if (!is_character(values)) {
-    abort("Internal error: `values` must be a character vector.")
+  } else {
+    signal_soft_deprecated("The `values` argument to `arg_match()` is deprecated as of rlang 0.4.7. Use `arg_match0()` instead.")
```
lionel-, line 39: Should be on two lines.

R/arg.R
```diff
@@ -47,30 +46,42 @@ arg_match <- function(arg, values = NULL) {
...
+#' It requires a string as `arg` and explicit `values`.
+#' @param arg_nm The label to be used for `arg` in error messages.
+#' @rdname arg_match
+#' @export
+#' @examples
+#'
+#' # Use the bare-bones version with explicit values for speed:
+#' arg_match0("bar", c("foo", "bar", "baz"))
+arg_match0 <- function(arg, values, arg_nm = as_label(substitute(arg))) {
+  .External2(rlang_ext2_arg_match0, arg, values, arg_nm)
```
lionel-, line 63: Should use `.External()` and pass the environment, even if slightly slower. `arg_nm` doesn't need to be passed down since it's constant.

src/internal/arg.c
```diff
@@ -84,3 +84,98 @@ sexp* rlang_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
 
   return r_lgl(r_eval(r_x_sym, env) == r_missing_sym);
 }
+
+static sexp* stop_arg_match_call = NULL;
+void arg_match0_abort(const char* msg, sexp* arg_nm_promise, sexp* env);
+sexp* rlang_ext2_arg_match0(sexp* _call, sexp* _op, sexp* args, sexp* env) {
+  args = r_node_cdr(args);
+
+  sexp* arg =            KEEP(r_eval(r_node_car(args), env)); args = r_node_cdr(args);
```
lionel-, line 93: We try to avoid column indenting.

src/internal/arg.c
```diff
@@ -84,3 +84,98 @@ sexp* rlang_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
...
+  if (r_typeof(values) != r_type_character) {
+    r_abort("`values` must be a character vector.");
+  }
+
+  r_ssize len = r_length(arg);
+  r_ssize len_values = r_length(values);
+  if (len != 1 && len != len_values) {
+    //arg_match0_abort("`%s` must be a string or have the same length as `values`.",  arg_nm_promise, env);
+
+    arg = Rf_ScalarString(STRING_ELT(arg, 0));
```
lionel-, line 110: From memory there should be something like `r_str_as_character(r_chr_get(arg, 0))`

src/internal/arg.c
```diff
@@ -84,3 +84,98 @@ sexp* rlang_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
...
+    r_eval_with_xyz(stop_arg_match_call, r_base_env, arg, values, arg_nm_promise);
+
+    while (1); // No return
```
lionel-, line 113: There's a safer `never_reached()` function, can you use that please?

src/internal/arg.c
lionel-: Same for the other usages of STRING_ELT and ScalarString below.

src/internal/arg.c
```diff
@@ -84,3 +84,98 @@ sexp* rlang_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
...
+      if (arg_char == STRING_ELT(values, i)) {
+        FREE(2);
+        return(Rf_ScalarString(arg_char));
+      }
+    }
+
+    r_eval_with_xyz(stop_arg_match_call, r_base_env, arg, values, arg_nm_promise);
+
+    while (1); // No return
+  } else {
```
lionel-, line 128: Can you move this block one level up since the first block never returns.

src/internal/arg.c
```diff
@@ -84,3 +84,98 @@ sexp* rlang_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
...
+    arg_match0_abort("`%s` must be a character vector.", arg_nm_promise, env);
+  }
+  if (r_typeof(values) != r_type_character) {
+    r_abort("`values` must be a character vector.");
+  }
+
+  r_ssize len = r_length(arg);
+  r_ssize len_values = r_length(values);
+  if (len != 1 && len != len_values) {
+    //arg_match0_abort("`%s` must be a string or have the same length as `values`.",  arg_nm_promise, env);
```
lionel-, line 108: WIP?

src/internal/arg.c
```diff
@@ -84,3 +84,98 @@ sexp* rlang_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
...
+      }
+    }
+
+    // Elements are in order, return first
+    if (!need_match) {
+      FREE(2);
+      return(Rf_ScalarString(STRING_ELT(values, 0)));
+    }
+
+    for (R_xlen_t i = ii; i < len; ++i) {
```
lionel-, line 144:
```suggestion
    for (r_ssize i = ii; i < len; ++i) {
```

src/internal/arg.c
```diff
@@ -84,3 +84,98 @@ sexp* rlang_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
...
+  if (len != 1 && len != len_values) {
+    //arg_match0_abort("`%s` must be a string or have the same length as `values`.",  arg_nm_promise, env);
+
+    arg = Rf_ScalarString(STRING_ELT(arg, 0));
+    r_eval_with_xyz(stop_arg_match_call, r_base_env, arg, values, arg_nm_promise);
+
+    while (1); // No return
+  }
+
+  if (len == 1) {
```
lionel-, line 116: It would be helpful to add a comment that in the case of a single string we just need to check it's one of the values.

src/internal/arg.c
```diff
@@ -84,3 +84,98 @@ sexp* rlang_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
...
+    // Elements are in order, return first
+    if (!need_match) {
+      FREE(2);
+      return(Rf_ScalarString(STRING_ELT(values, 0)));
+    }
+
+    for (R_xlen_t i = ii; i < len; ++i) {
+      bool matched = false;
+      for (r_ssize j = ii; j < len; ++j) {
+        if (STRING_ELT(arg, i) == STRING_ELT(values, j)) {
```
lionel-, line 147: Is this correct? I think we need to use something equivalent to `all(arg[-1] %in% values)`? This can probably be moved to R code if needed, just add a `sexp* call = r_parse(*code*)` in the init section and use `r_eval_with_xy()`. Alternatively, I think we can call `match()` in a loop with the C API.

R/arg.R
krlmlr: Can you propose a change please?

I can benchmark before-after to assess impact, unless you have already?

src/internal/arg.c
krlmlr: This was cargo-culted from existing implementation :zany_face:

src/internal/arg.c
krlmlr: I think we need `identical(sort(arg), sort(values))`. Will revisit.

lionel-: Good point!

R/arg.R
lionel-: I did once, the impact was minimal. Even if not minimal, I still would like to minimise the use of .External2 because it's not a public entry point and because it's tricky to use. Any call into R that changes the value of `R_Visibility` may cause unexpected visibility return.

You can leave this as is and I'll push a change to .External once the PR is done.

src/internal/arg.c
krlmlr: Not sure what the better error message in this situation is. Happy to resolve either way.

src/internal/arg.c
krlmlr: Done.

krlmlr:
The main branch is failing checks.

What's a good way to avoid forcing `arg_nm`? This costs a lot.

## main branch

``` r
library(rlang)

foo <- function(x = c("arg1", "arg2")) {
  arg_match(x)
}

bench::mark(foo("arg1"))
#> # A tibble: 1 x 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 foo("arg1")    162µs    183µs     4382.     622KB     16.4
```

<sup>Created on 2020-06-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

## This PR

``` r
library(rlang)

foo <- function(x = c("arg1", "arg2")) {
  arg_match(x)
}

bar <- function(x) {
  arg_match0(x, c("arg1", "arg2"))
}

baz <- function(x) {
  arg_match0(x, c("arg1", "arg2"), "x")
}

bench::mark(foo("arg1"), bar("arg1"), baz("arg1"))
#> # A tibble: 3 x 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 foo("arg1") 159.22µs 171.46µs     4973.     582KB     21.9
#> 2 bar("arg1")   9.12µs  10.18µs    86717.     124KB     17.3
#> 3 baz("arg1")   1.07µs   1.32µs   609153.        0B      0
```

<sup>Created on 2020-06-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lionel-:
> What's a good way to avoid forcing arg_nm?

Pass the execution environment to the throwing function and it can evaluate `arg_nm` from there.

krlmlr:
Much better, still whopping ~3k cycles:

``` r
library(rlang)

bar <- function(x) {
  arg_match0(x, c("arg1", "arg2"))
}

baz <- function(x) {
  arg_match0(x, c("arg1", "arg2"), "x")
}

bench::mark(bar("arg1"), baz("arg1"), iterations = 1000000)
#> # A tibble: 2 x 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 bar("arg1") 984.06ns   1.32µs   565358.     6.8KB     71.2
#> 2 baz("arg1")   1.01µs   1.26µs   632004.        0B     22.8
```

<sup>Created on 2020-06-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

krlmlr:
Tests still fail, I just rebased off of master. Locally on Ubuntu I'm getting weird diffs when I run `devtools::test()` .

krlmlr:
Long live CI! Looks like I wrote into memory I wasn't allowed to write.

R/arg.R
```diff
@@ -35,9 +35,10 @@ arg_match <- function(arg, values = NULL) {
     fn <- caller_fn()
     values <- fn_fmls(fn)[[arg_nm]]
     values <- eval_bare(values, get_env(fn))
-  }
-  if (!is_character(values)) {
-    abort("Internal error: `values` must be a character vector.")
+  } else {
+    signal_soft_deprecated(
+      "The `values` argument to `arg_match()` is deprecated as of rlang 0.4.7. Use `arg_match0()` instead."
```
DavisVaughan, line 40: I feel slightly uncomfortable about deprecating `values`, but I'll defer to @lionel-

R/arg.R
```diff
@@ -47,30 +48,47 @@ arg_match <- function(arg, values = NULL) {
...
+  arg_match0(arg, values, arg_nm)
+}
 
-  if (is_na(i)) {
-    msg <- arg_match_invalid_msg(arg_nm, values)
+#' @description
+#' `arg_match0()` is a bare-bones version if performance is at a premium.
+#' It requires a string as `arg` and explicit `values`.
+#' (For convenience, setting `arg` to `values`, perhaps permutated, is also
+#' supported.)
```
DavisVaughan, line 58: For convenience, `arg` may also be a character vector containing every element of `values`, possibly permuted.

src/internal/arg.c
```diff
@@ -84,3 +85,107 @@ sexp* rlang_ext2_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
...
 }
+
+static sexp* stop_arg_match_call = NULL;
+static sexp* arg_nm_sym = NULL;
+
+void arg_match0_abort(const char* msg, sexp* env);
+sexp* rlang_ext2_arg_match0(sexp* _call, sexp* _op, sexp* args, sexp* env) {
+  args = r_node_cdr(args);
+
+  sexp* arg =    r_node_car(args); args = r_node_cdr(args);
```
DavisVaughan, line 96:
```suggestion
  sexp* arg = r_node_car(args); args = r_node_cdr(args);
```

src/internal/arg.c
```diff
@@ -84,3 +85,107 @@ sexp* rlang_ext2_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
...
+  if (r_typeof(values) != r_type_character) {
+    r_abort("`values` must be a character vector.");
+  }
+
+  r_ssize len = r_length(arg);
+  r_ssize len_values = r_length(values);
+  if (len != 1 && len != len_values) {
+    arg = KEEP(r_str_as_character(r_chr_get(arg, 0)));
+    sexp* arg_nm = KEEP(r_eval(arg_nm_sym, env));
+    r_eval_with_xyz(stop_arg_match_call, r_base_env, arg, values, arg_nm);
```
DavisVaughan, line 111: Eval in the base env? Maybe instead you could eval in rlang's package namespace? Then `stop_arg_match_call` wouldn't have to be `rlang:::` I think?

src/internal/arg.c
```diff
@@ -84,3 +85,107 @@ sexp* rlang_ext2_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
...
+
+    never_reached("rlang_ext2_arg_match0");
+  }
+
+  // Simple case: one argument, we check if it's one of the values.
+  if (len == 1) {
+    sexp* arg_char = r_chr_get(arg, 0);
+    for (r_ssize i = 0; i < len_values; ++i) {
+      if (arg_char == r_chr_get(values, i)) {
+        return(r_str_as_character(arg_char));
```
DavisVaughan, line 121: Can you just return `arg`?

src/internal/arg.c
```diff
@@ -84,3 +85,107 @@ sexp* rlang_ext2_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
...
+      }
+    }
+
+    sexp* arg_nm = KEEP(r_eval(arg_nm_sym, env));
+    r_eval_with_xyz(stop_arg_match_call, r_base_env, arg, values, arg_nm);
```
DavisVaughan, line 126: Same as above `r_base_env` comment

src/internal/arg.c
krlmlr: This pattern is used across rlang, I don't find `r_rlang_env` or similar symbols in the code.

lionel-: You can use `rlang_ns_env`:

```
src/internal/attr.c:152:  return r_eval_with_xy(as_function_call, rlang_ns_env, x, env);
src/internal/dots.c:12:  return r_eval_with_x(as_label_call, rlang_ns_env, x);
src/internal/dots.c:208:      return r_eval_with_x(rlang_as_list_call, rlang_ns_env, x);
src/internal/dots.c:214:      return r_eval_with_x(rlang_as_list_call, rlang_ns_env, x);
src/internal/dots.c:219:    return r_eval_with_x(rlang_as_list_call, rlang_ns_env, x);
src/internal/eval-tidy.c:346:  sexp* out = r_eval_with_x(call, rlang_ns_env, mask);
src/lib/env.c:141:  r_eval_with_xyz(poke_lazy_call, rlang_ns_env, name, env, eval_env);
src/lib/cnd.c:98:  r_eval_with_x(call, r_ns_env("rlang"), msg);
```

R/arg.R
lionel-: It's rarely used, only soft-deprecated, and the deprecation cycles in rlang are very long. For these reasons I think it's ok and this way we encourage using the non-NSE function when performance is important.

src/internal/arg.c
```diff
@@ -84,3 +85,107 @@ sexp* rlang_ext2_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
...
+
+    never_reached("rlang_ext2_arg_match0");
+  }
+
+  // Same-length vector: must be identical, we allow changed order.
+  r_ssize i = 0;
+  for (; i < len; ++i) {
+    if (r_chr_get(arg, i) != r_chr_get(values, i)) {
```
lionel-, line 134: `r_chr_deref_const(arg)`

src/internal/arg.c
```diff
@@ -84,3 +85,107 @@ sexp* rlang_ext2_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
...
+
+  sexp* arg = r_node_car(args); args = r_node_cdr(args);
+  sexp* values = r_node_car(args);
+
+  if (r_typeof(arg) != r_type_character) {
+    arg_match0_abort("`%s` must be a character vector.", env);
+  }
+  if (r_typeof(values) != r_type_character) {
+    r_abort("`values` must be a character vector.");
+  }
```
lionel-, line 104: Also need to check for length zero cases with unit test coverage.

src/internal/arg.c
```diff
@@ -84,3 +85,107 @@ sexp* rlang_ext2_is_missing(sexp* _call, sexp* _op, sexp* args, sexp* env) {
...
+    sexp* current_arg = r_chr_get(arg, i);
+    if (current_arg == r_chr_get(my_values, i)) {
+      continue;
+    }
+
+    bool matched = false;
+    for (r_ssize j = i + 1; j < len; ++j) {
+      if (current_arg == r_chr_get(my_values, j)) {
+        matched = true;
+        r_chr_poke(my_values, j, r_chr_get(my_values, i));
```
lionel-, line 157: Comment:

> Replace matched value by the element that failed to match at this iteration

R/arg.R
```diff
@@ -47,30 +48,55 @@ arg_match <- function(arg, values = NULL) {
...
+#'
+#' # For convenience:
+#' fn1 <- function(x = c("bar", "baz", "foo")) fn3(x)
+#' fn2 <- function(x = c("baz", "bar", "foo")) fn3(x)
+#' fn3 <- function(x) arg_match0(x, c("foo", "bar", "baz"))
+#' fn1()
+#' fn2("bar")
+#' try(fn3("zoo"))
+arg_match0 <- function(arg, values, arg_nm = as_label(substitute(arg))) {
+  .External2(rlang_ext2_arg_match0, arg, values)
```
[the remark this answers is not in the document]
krlmlr: I don't want to pass `arg_nm` because it's very expensive to evaluate. Do I still get access to the evaluation environment with `.External()` ?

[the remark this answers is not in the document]
lionel-: yes with `environment()`.

lionel-:
Thanks @krlmlr!

PR #999, opened 2020-06-15
Run exit expressions in LIFO order

### LINKED ISSUE #980: local_options() restores options in wrong order
Can this even be fixed easily for R < 3.6?

``` r
library(rlang)

options(foo = 2)

local({
  local_options(foo = 0)
  local_options(foo = 1)
})

getOption("foo")
#> [1] 0
```

<sup>Created on 2020-05-20 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lionel-:
Closes #980

Implements `after = ` logic with a custom stack of exit expressions so it works on old R versions.

### DIFF
--- a/R/events.R
+++ b/R/events.R
@@ -1,6 +1,7 @@
 
 local_exit <- function(expr, frame = caller_env()) {
   expr <- enexpr(expr)
+  out <- expr
 
   # We are at top-level when only one frame refers to the global environment
   if (is_reference(frame, global_env())) {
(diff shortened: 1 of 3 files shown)


lionel-:
Do you think `defer()` could live in rlang Jim? This way we'd merge implementations for e.g. `rlang::with_bindings()` and `withr::with_options()`.

jimhester:
My hesitation on doing that is that currently withr has no hard dependencies on any non-default packages, so depending on rlang would be a large change.

I do think it would be nice to use the same implementations somehow though.

lionel-:
Would it make sense to move `withr::defer()` to a drop-in compat file published in withr, which we would copy in rlang? In that case I'll rewrite this PR to use base R.

jimhester:
Yeah, that sounds like a good solution to me.

lionel-:
oh now I see that withr already implements its own stack of exits. I'll just go ahead and merge this now and we can switch to defer in rlang later on.

PR #1000, opened 2020-06-16
Use `withr::defer()` for restoring state

lionel-:
Follow up to #999.


lionel-:
Merging now to fix the CI builds on old R versions.

PR #1002, opened 2020-06-17
Use R_*Value in r_lgl()

krlmlr:
to shave off a few cycles perhaps? Impact not measured.

AFAICT R checks consistency of these values in `gc()`, they should be safe? Are we allowed to use them anyway?


lionel-:
You're reimplementing ScalarLogical, I think we can close this.

krlmlr:
:zany_face:

PR #1011, opened 2020-06-24
Use rlang_ext2_* prefix consistently

krlmlr:
for functions called via `.External2()` .

(Where is that documented? WRE doesn't mention it.)


lionel-:
I now think `is_call()` should be using `.External()`.

`.External2()` is not part of the public API so it's not documented. It's the only way to propagate invisibility after an `Rf_eval()`.

PR #1015, opened 2020-06-30
Namespace prefix all rlang functions in `compat-lifecycle.R`

DavisVaughan:
I noticed that this didn't namespace prefix everything that comes from rlang


lionel-:
Thanks!

PR #1034, opened 2020-07-23
Update the GHA workflows that now need git config

jennybc:


jennybc:
I did this mostly because I needed a legit PR to exercise some dev in `usethis::pr_*()` functions! Enjoy.

PR #1054, opened 2020-10-20
Initialise data mask with expected number of elements

### LINKED ISSUE #1048: Allocate environment hash table of correct size in `as_data_mask()`

lionel-:
Closes #1048. Performance is improved but not dramatically.

```r
data <- as.list(set_names(1:1e5))
bench::mark(before = as_data_mask(data))[1:8]
```

Before

```r
#>   expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 before        220ms    223ms      4.47    5.44MB        0     3     0
```

After

```r
#>   expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 after         169ms    170ms      5.88    2.76MB        0     3     0
```

For 1e6 elements this goes from 2.35s to 1.85s.

### DIFF
--- a/src/internal/eval-tidy.c
+++ b/src/internal/eval-tidy.c
@@ -160,14 +160,14 @@ sexp* rlang_new_data_mask(sexp* bottom, sexp* top) {
   sexp* data_mask;
 
   if (bottom == r_null) {
-    bottom = KEEP(r_new_environment(r_empty_env, 0));
+    bottom = KEEP(r_new_environment(r_empty_env, 100));
     data_mask = bottom;
   } else {
     check_data_mask_input(bottom, "bottom");
     // Create a child because we don't know what might be in `bottom`
     // and we need to clear its contents without deleting any object
     // created in the data mask environment
-    data_mask = KEEP(r_new_environment(bottom, 0));
+    data_mask = KEEP(r_new_environment(bottom, 100));
   }
 
   if (top == r_null) {
@@ -271,6 +271,7 @@ static void warn_env_as_mask_once() {
 }
 
 static sexp* data_pronoun_sym = NULL;
+static r_ssize mask_length(r_ssize n);
 
 sexp* rlang_as_data_mask(sexp* data) {
   if (mask_info(data).type == RLANG_MASK_DATA) {
@@ -304,7 +305,9 @@ sexp* rlang_as_data_mask(sexp* data) {
     check_unique_names(data);
 
     sexp* names = r_names(data);
-    bottom = KEEP_N(r_new_environment(r_empty_env, 0), n_protect);
+
+    r_ssize n_mask = mask_length(r_length(data));
+    bottom = KEEP_N(r_new_environment(r_empty_env, n_mask), n_protect);
 
     if (names != r_null) {
       r_ssize n = r_length(data);
@@ -333,6 +336,12 @@ sexp* rlang_as_data_mask(sexp* data) {
 
   FREE(n_protect);
   return data_mask;
+}
+
+static
+r_ssize mask_length(r_ssize n) {
+  r_ssize n_grown = r_double_as_ssize(r_double_mult(r_ssize_as_double(n), 1.05));
+  return r_ssize_max(n_grown, r_ssize_add(n, 20));
 }
 
 // For compatibility of the exported C callable
--- a/src/lib/c-utils.h
+++ b/src/lib/c-utils.h
@@ -87,8 +87,6 @@ double r_ssize_as_double(r_ssize x) {
 
 static inline
 r_ssize r_double_as_ssize(double x) {
-  x = round(x);
-
   if (x > R_SSIZE_MAX || x < R_SSIZE_MIN) {
     r_stop_internal("r_ssize_as_double", "Result can't be represented as `r_ssize`.");
   }
--- a/src/lib/c-utils.h
+++ b/src/lib/c-utils.h
@@ -101,7 +101,7 @@ double r_double_mult(double x, double y) {
   double out = x * y;
 
   if (!isfinite(out)) {
-    r_stop_internal("r_double_mult", "Can't multiple double values.");
+    r_stop_internal("r_double_mult", "Can't multiply double values.");
   }
 
   return out;
--- a/src/lib/c-utils.h
+++ b/src/lib/c-utils.h
@@ -1,6 +1,7 @@
 #ifndef RLANG_C_UTILS_H
 #define RLANG_C_UTILS_H
 
+#include <float.h>
 #include "cnd.h"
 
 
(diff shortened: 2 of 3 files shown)


src/internal/eval-tidy.c
```diff
@@ -160,14 +160,14 @@ sexp* rlang_new_data_mask(sexp* bottom, sexp* top) {
...
-    bottom = KEEP(r_new_environment(r_empty_env, 0));
+    bottom = KEEP(r_new_environment(r_empty_env, 100));
     data_mask = bottom;
   } else {
     check_data_mask_input(bottom, "bottom");
     // Create a child because we don't know what might be in `bottom`
     // and we need to clear its contents without deleting any object
     // created in the data mask environment
-    data_mask = KEEP(r_new_environment(bottom, 0));
+    data_mask = KEEP(r_new_environment(bottom, 100));
```
DavisVaughan, line 170: The increase to 100 is a bit unexpected, I think.

So previously using `0` would result in the internal default of `29` being used. I thought that this was a little high, honestly.

Isn't it somewhat rare to add additional bindings to a data mask? With the exception of: `mask$.data <- new_data_pronoun(mask)`

So I would have set the size to `4 + buffer`, where `4` comes from the 4 items you poke into the environment below, and `buffer` is maybe 3 or 4 in case the user wants to add `$.data <- ` or a small number of other things to the mask

I only bring this up because using a much smaller default hash table size can improve performance when you have to create a large number of environments. This approximates the difference:

``` r
bench::mark(
  slow = {
    for (i in seq_len(1e5L)) {
      vector("list", 100L) # large hash table
    }
    
  },
  fast = {
    for (i in seq_len(1e5L)) {
      vector("list", 7L) # small hash table
    }
  }
)
#> Warning: Some expressions had a GC in every iteration; so filtering is disabled.
#> # A tibble: 2 x 6
#> 1 slow         75.5ms   84.8ms      11.8    84.8MB     35.5
#> 2 fast         17.9ms   21.7ms      47.9    10.2KB     24.0
```

<sup>Created on 2020-10-20 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

DavisVaughan: I was also thinking the environment created in `rlang_new_ctxt_pronoun()` could be much smaller. Maybe with a size of 2-3, since I doubt that ever gets modified? (at least that is what im guessing) https://github.com/r-lib/rlang/blob/4227d04a325763611564202a1ce95d52391e3da9/src/internal/eval-tidy.c#L54

src/internal/eval-tidy.c
```diff
@@ -304,7 +305,9 @@ sexp* rlang_as_data_mask(sexp* data) {
     check_unique_names(data);
 
     sexp* names = r_names(data);
-    bottom = KEEP_N(r_new_environment(r_empty_env, 0), n_protect);
+
+    r_ssize n_mask = mask_length(r_length(data));
+    bottom = KEEP_N(r_new_environment(r_empty_env, n_mask), n_protect);
```
DavisVaughan, line 310: 👍

src/lib/c-utils.h
```diff
@@ -69,9 +70,41 @@ static inline
...
+  if (x > DBL_MAX || x < -DBL_MAX) {
+    r_stop_internal("r_ssize_as_double", "Result can't be represented as `double`.");
+  }
+
+  return (double) x;
+}
+
+static inline
+r_ssize r_double_as_ssize(double x) {
+  x = round(x);
```
DavisVaughan, line 90: `round()` or `floor()`? i.e. should `0.6` round up to `1`? i think base R uses `(R_xlen_t) x` which would truncate to `0`

src/lib/c-utils.h
```diff
@@ -69,9 +70,41 @@ static inline
...
+
+  if (x > R_SSIZE_MAX || x < R_SSIZE_MIN) {
```
DavisVaughan, line 92: do you want `x < 0` rather than `< R_SSIZE_MIN ` maybe? i don't know if you want to error on negative sizes here, but maybe:

```c
static inline
r_ssize r_double_as_ssize(double x) {
  if (x > R_SSIZE_MAX || x < 0) {
    r_stop_internal();
  }

  x = floor(x);

  return (r_ssize) x;
}
```

in which case the `floor()` might even be repetitive because of the cast to `(r_ssize)`

src/lib/c-utils.h
```diff
@@ -69,9 +70,41 @@ static inline
...
+
+  return (r_ssize) x;
+}
+
+static inline
+double r_double_mult(double x, double y) {
+  double out = x * y;
+
+  if (!isfinite(out)) {
+    r_stop_internal("r_double_mult", "Can't multiple double values.");
```
DavisVaughan, line 104:
```suggestion
    r_stop_internal("r_double_mult", "Can't multiply double values.");
```

src/internal/eval-tidy.c
lionel-: I can't reproduce your results:

```r
bench::mark(
  `0` = vector("list", 0),
  `30` = vector("list", 30),
  `100` = vector("list", 100),
  check = FALSE,
  iterations = 1e5
)[1:8]
#> # A tibble: 3 x 8
#>   expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 0             546ns    660ns  1376479.        0B    13.8  99999     1
#> 2 30            621ns    775ns  1084134.      288B    10.8  99999     1
#> 3 100           624ns    882ns   861974.      848B     8.62 99999     1
```

Isn't it weird that your slow run has 84MB of allocation while your fast run has 10KB?

In general, the difference between allocating a small and large vector is not that important. It's more about avoiding many different allocations.

src/lib/c-utils.h
lionel-: I wondered about this and couldn't decide what is best.

src/lib/c-utils.h
lionel-: Since this is specifically a signed type, I think we should allow negative integers. They might be error codes.

src/lib/c-utils.h
lionel-: When this matters, it's probably best for the caller to pre-trim the double.

src/internal/eval-tidy.c
lionel-: > Isn't it somewhat rare to add additional bindings to a data mask?

`mutate()` (which however uses as-data-mask rather than new-data-mask) would add bindings for new columns example, so it's probably not that rare. If you're using `new_data_mask()` you probably want to fill the mask with some bindings? In general there is no reason to use `eval_tidy()` unless you're masking data.

Anyway, given the small 100ns overhead that's probably a good enough default.

I do wonder where comes the difference between our benchmarks though, it seems like there is a bug somewhere.

lionel-:
Thanks for the review! Merging now but let's revisit the default values once we get to the bottom of the benchmark diffs

src/internal/eval-tidy.c
DavisVaughan: Two things I thought about regarding `mutate()`.

First, Romain's new PR actually was going to change to use `new_data_mask()` before you updated this one to initialize the mask with the expected size. I think he was essentially initializing the mask to the correct size himself.
https://github.com/tidyverse/dplyr/pull/5520/files#diff-d9709c495664bec3d8718ad24bf9db7391406f7ddac818694254efe79b2fe2acR74-R79

Second, regarding:

> `mutate()` would add bindings for new columns

I thought about this as well, and almost left a comment about it, but then I went and looked at the new implementation. _Technically_ it adds bindings into the enclosing environment of the data mask.

https://github.com/tidyverse/dplyr/pull/5520/files#diff-75499699eb24243ad9de81f1ea71d00d2cb1be968b57459fde2653a5006f5391R65

Whether this is correct or not, I'm not sure, but I think what Romain was trying to do is to install the new columns from `mutate()` into his original mask env that he created with the correct size. Since `new_data_mask()` creates a child env from `bottom`, he was inserting back into that original parent env.

PR #1058, opened 2020-11-10
Add POC simple hashing function with MD5 backend

DavisVaughan:
This PR adds a POC for `hash()` - a hashing function for arbitrary R objects backed by MD5.

It uses an implementation of MD5 placed in the open domain with absolutely no warranty or copyright.
https://openwall.info/wiki/people/solar/software/public-domain-source-code/md5

Supposedly this impl is used in PHP and Freetype, so I feel pretty good about it.

It passes the small test suite proposed in this RFC standard for MD5. Note that they also provide an implementation, which I tried using first, but it assumed a 32 bit system and was returning incorrect results. The implementation used here explicitly states that it overcomes those limitations.
https://tools.ietf.org/html/rfc1321

I've kept the copied files in `src/internal/md5/`. Then `hash.c` includes the `md5.c` file at the top. This seemed reasonable based on how you include other `.c` files (like in the top level `internal.c`), but let me know if you'd like it set up differently.

---

I have absolutely no idea what the "best" arguments to `serialize()` are, but `version = 2` seems reasonable to be reproducible on older versions of R.

The `hash_string()` function is mainly for testing.


DavisVaughan:
Hmm it seems like `serialize()` returns different things on different platforms.

On my Mac

```r
> serialize(c(1, 2, 3), NULL, version = 2)
 [1] 58 0a 00 00 00 02 00 04 00 00 00 02 03 00 00 00 00 0e 00 00 00 03 3f f0 00 00 00 00 00 00 40 00 00 00 00 00 00 00 40 08 00 00 00 00
[45] 00 00
```

On rstudio cloud linux:

```r
> serialize(c(1, 2, 3), NULL, version = 2)
 [1] 58 0a 00 00 00 02 00 04 00 03 00 02 03 00 00 00 00 0e 00 00 00 03 3f f0 00
[26] 00 00 00 00 00 40 00 00 00 00 00 00 00 40 08 00 00 00 00 00 00
```

Almost the same, except for element 10, which is `00` on Mac and `03` on Linux

DavisVaughan:
Another update. Using `data.frame(x=1)` as a test case for `hash()` might not have been the best idea. It seems that something changed with how R serializes objects somewhere around R 3.4. Possibly this only has to do with objects that have attributes, since the simpler test of `c(1, 2, 3)` was passing on all R versions.

Here is an example, equivalent calls to `serialize()` on R versions 3.3.3 and 4.0.0 on `data.frame(x=1)` return very different results (even keeping the serialization header in mind)

``` r
df_4.0.0 <- serialize(
  data.frame(x = 1), 
  ascii = FALSE, 
  version = 2L, 
  connection = NULL
)

# From equivalent serialize() call on R 3.3.3
df_3.3.3 <- as.raw(c(0x58, 0x0a, 0x00, 0x00, 0x00, 0x02, 0x00, 0x03, 0x03, 
                     0x03, 0x00, 0x02, 0x03, 0x00, 0x00, 0x00, 0x03, 0x13, 0x00, 0x00, 
                     0x00, 0x01, 0x00, 0x00, 0x00, 0x0e, 0x00, 0x00, 0x00, 0x01, 0x3f, 
                     0xf0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x04, 0x02, 
                     0x00, 0x00, 0x00, 0x01, 0x00, 0x04, 0x00, 0x09, 0x00, 0x00, 0x00, 
                     0x05, 0x6e, 0x61, 0x6d, 0x65, 0x73, 0x00, 0x00, 0x00, 0x10, 0x00, 
                     0x00, 0x00, 0x01, 0x00, 0x04, 0x00, 0x09, 0x00, 0x00, 0x00, 0x01, 
                     0x78, 0x00, 0x00, 0x04, 0x02, 0x00, 0x00, 0x00, 0x01, 0x00, 0x04, 
                     0x00, 0x09, 0x00, 0x00, 0x00, 0x09, 0x72, 0x6f, 0x77, 0x2e, 0x6e, 
                     0x61, 0x6d, 0x65, 0x73, 0x00, 0x00, 0x00, 0x0d, 0x00, 0x00, 0x00, 
                     0x02, 0x80, 0x00, 0x00, 0x00, 0xff, 0xff, 0xff, 0xff, 0x00, 0x00, 
                     0x04, 0x02, 0x00, 0x00, 0x00, 0x01, 0x00, 0x04, 0x00, 0x09, 0x00, 
                     0x00, 0x00, 0x05, 0x63, 0x6c, 0x61, 0x73, 0x73, 0x00, 0x00, 0x00, 
                     0x10, 0x00, 0x00, 0x00, 0x01, 0x00, 0x04, 0x00, 0x09, 0x00, 0x00, 
                     0x00, 0x0a, 0x64, 0x61, 0x74, 0x61, 0x2e, 0x66, 0x72, 0x61, 0x6d, 
                     0x65, 0x00, 0x00, 0x00, 0xfe))

# Different in many places, not just the serialization header
which(df_4.0.0 != df_3.3.3)
#>  [1]   8   9  10  92  93  94  95  96  97  98  99 100 101 105 107 109 110 113 114
#> [20] 115 116 117 118 119 120 121 122 123 125 126 127 129 131 133 134 135 136 137
#> [39] 138 139 140 141 142 143 144 145 146 147 148 150 152 154 155 156 157 158 159
#> [58] 160 161 162 163 164
```

<sup>Created on 2020-11-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

It is worth keeping this in mind if we are going to use `hash()` in other packages and expect it to be reproducible all the way back to R 3.3

src/internal/md5/md5.c
```diff
@@ -0,0 +1,291 @@
+/*
+ * This is an OpenSSL-compatible implementation of the RSA Data Security, Inc.
+ * MD5 Message-Digest Algorithm (RFC 1321).
+ *
+ * Homepage:
+ * http://openwall.info/wiki/people/solar/software/public-domain-source-code/md5
+ *
+ * Author:
+ * Alexander Peslyak, better known as Solar Designer <solar at openwall.com>
```
hadley, line 9: I think we should still add him as an author in DESCRIPTION.

wch:
If this will be used to hash large objects, you may want to look into how the digest package does streaming hashing with spookyhash (https://github.com/eddelbuettel/digest/blob/master/src/spooky_serialize.cpp). I believe it doesn't require the entire object to be serialized and stored in memory before passing it to the hashing algorithm; it hashes as the serialization occurs.

It appears to be adapted from [fastdigest](https://cran.r-project.org/web/packages/fastdigest/index.html). Notably:

> Speed is high and scales linearly by object size; memory usage is constant and negligible.

hadley:
There's also https://github.com/coolbutuseless/xxhashlite, which is MIT licensed.

DavisVaughan:
I've added xxhash as a streaming hash, following the implementation by @coolbutuseless in https://github.com/coolbutuseless/xxhashlite

It is _much_ faster than MD5. I've used the 128-bit version, which is the same number of bits as MD5.
I'll remove MD5 and clean up if we decide that this is what we'd rather use.

We'll need to add @coolbutuseless to the Authors list before merging if we do go this way.

For now:
- We use `version = 2` when serializing. Again, this materializes ALTREP objects.
- We skip the serialization header.

xxhashlite wasn't doing either of these so the hashes are going to be different.

It might be a little tricky to use `version = 3` to avoid materializing altrep. The serialization header in version 3 serializes a string specifying the native encoding, and we don't have access to `R_NativeEncoding()` to figure that out. I _think_ there is a way around it, but it might be fragile. We'd also have to switch to version 2 on older R versions somehow.

hadley:
I think we could just promise hash uniqueness within R version and native encoding? I suspect that's all we need in most cases.

DavisVaughan:
@hadley, promising hash uniqueness within the R (minor) version makes sense to me.

Do we want reproducibility across OSs within the same R version? Isn't the native encoding different between Windows and Mac? If we don't skip the native encoding bytes that R appends to the serialization result, then `hash(1)` would probably produce different results on a Windows vs Mac, just because the native encoding is different between them.

As currently implemented, we do skip those bytes. So currently it is reproducible across OSs within the same R version.

wch:
Does anything special need to be done to note the license of xxhash?

DavisVaughan:
I'm not sure. xxHash is BSD-2, which I've been told is compatible with MIT (which rlang will eventually become). The `xxhash.h` file itself has the BSD-2 license in it, which I've kept. The only other thing I can think of that we might do is add a top level `LICENSE.note` that includes this bit of the top level xxHash license file - noting that everything under `src/internal/xxhash/` falls under this additional license. I'll defer to @hadley here. https://github.com/Cyan4973/xxHash/blob/4c881f796d6af27ef7d9c48f87817da0d3d75dc1/LICENSE#L1-L26

hadley:
We should add Yann Collet to authors, and add a bullet point to LICENSE.note as @DavisVaughan suggests.

DavisVaughan:
@lionel- this is ready for a review now. A summary of the final implementation:

`hash()` generates a 128-bit hash for an arbitrary R object using xxHash. It is intended to be reproducible across platforms, but not across R versions. That said, extremely simple objects often do generate the same hash across R versions, i.e. the ones in the tests. Once you add attributes or ALTREP objects, the hashes start to differ.

The hash is "streaming", allowing us to hook directly into `R_Serialize()` to compute the hash progressively as it serializes. This is much faster than serializing + running the result through an xxHash function, and uses way less memory.

On R >= 3.5.0, we use `version = 3` to serialize, otherwise we use `version = 2`. Version 3 has better support for ALTREP. I have accomplished this in the code at compile time through the usage of macros that enable / disable parts of the code based on the R version.

`R_Serialize()` will always append some information to the front of the serialized object. We always skip this information. This information is different between version 2 and version 3, and again macros are used to generate the correct skipping code at compile time. The reason we skip this information is to be reproducible across platforms. With version 3, a string representing the native encoding is part of the information that gets appended, and this varies between platforms, which is why it is important to skip it. Information about the R version is also appended, but this isn't as important to skip, since we only guarantee reproducibility within the same R version.

src/internal/hash.c
```diff
@@ -0,0 +1,225 @@
...
+  if (err == XXH_ERROR) {
+    XXH3_freeState(p_xx_state);
+    r_abort("Couldn't update hash state.");
+  }
+}
+
+static inline
+void hash_char(R_outpstream_t stream, int input) {
+  static int n = 1;
+  void* p_input = (void*) (&input);
```
easyaspi314, line 172: Implementation-defined behavior, violates strict aliasing and relies on the target being little endian.

See my comment in [Cyan4973/xxHash#482](https://github.com/Cyan4973/xxHash/issues/482)

src/internal/hash.c
```diff
@@ -0,0 +1,226 @@
...
+
+  if (err == XXH_ERROR) {
+    XXH3_freeState(p_xx_state);
+    r_abort("Couldn't update hash state.");
+  }
+}
+
+static inline
+void hash_char(R_outpstream_t stream, int input) {
+  static int n = 1;
```
lionel-, line 171: Why is this a `static` variable? I would have expected just inlining `1` in the call to `hash_bytes()`.

src/internal/hash.c
```diff
@@ -0,0 +1,226 @@
...
+    return;
+  }
+
+  p_state->n_skipped += n;
+
+  // Haven't reached native encoding size yet,
+  // there is no way we have skipped everything
+  if (p_state->n_native_enc == 0) {
+    return;
+  }
```
lionel-, line 198: This might be easier to follow if this were the first branch with `p_state->n_skipped < N_BYTES_SERIALIZATION_INFO` as condition? Then the order in the code corresponds to the skip progress: skip until we reach the info's number of bytes, then record the native encoding, then skip the encoding string.

src/internal/hash.c
```diff
@@ -0,0 +1,226 @@
...
+  if (err == XXH_ERROR) {
+    XXH3_freeState(p_xx_state);
+    r_abort("Couldn't update hash state.");
+  }
+}
+
+static inline
+void hash_char(R_outpstream_t stream, int input) {
+  static int n = 1;
+  unsigned char byte = (unsigned char) input;
```
lionel-, line 172: Can you add a comment that despite the confusing signature of this method, `input` is always a `char` and so this conversion is safe?

lionel- (approved):
Looks great! I really like the simplicity of this. Just needs a bit more work to protect against jumps during serialisation, e.g. with Ctrl-c.

Also I'm seeing this with `clang -Wall -pendantic`:

```
   In file included from internal.c:12:
   In file included from ./internal/hash.c:16:
   ./internal/xxhash/xxhash.h:4766:7: warning: no newline at end of file [-Wnewline-eof]
   #endif
         ^
   1 warning generated.
```

src/internal/hash.c
```diff
@@ -0,0 +1,226 @@
...
+  XXH3_freeState(p_xx_state);
+
+  // R assumes C99, so these are always defined as `uint64_t` in xxhash.h
+  XXH64_hash_t high = hash.high64;
+  XXH64_hash_t low = hash.low64;
+
+  // 32 for hash, 1 for terminating null added by `sprintf()`
+  char out[32 + 1];
+
+  sprintf(out, "%" PRIx64 "%" PRIx64, high, low);
```
easyaspi314, line 123: `"%016" PRIx64`.

src/internal/hash.c
```diff
@@ -0,0 +1,226 @@
...
+
+  XXH128_hash_t hash = XXH3_128bits_digest(p_xx_state);
+  XXH3_freeState(p_xx_state);
+
+  // R assumes C99, so these are always defined as `uint64_t` in xxhash.h
+  XXH64_hash_t high = hash.high64;
+  XXH64_hash_t low = hash.low64;
+
+  // 32 for hash, 1 for terminating null added by `sprintf()`
+  char out[32 + 1];
```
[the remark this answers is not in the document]
easyaspi314: Overallocating has no benefit here. `PRIx64` is guaranteed to print a 64-bit integer exactly, which is guaranteed to be up to 16 hex digits. So that `sprintf` is guaranteed to write a maximum of 33 bytes.

And stack allocation is definitely faster and better for cache, especially when position independent code is involved.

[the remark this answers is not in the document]
lionel-: Stack allocation definitely has no measurable impact on performance here because this memory is used once per call.

I know the size is correct and it should be mentioned in the comment. I'd still overallocate as a matter of principle since that reduces risks on average, this is not about this particular case.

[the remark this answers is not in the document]
easyaspi314: If you are so worried, just use `snprintf`.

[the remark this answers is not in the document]
easyaspi314: If you're that worried, use `snprintf`.

easyaspi314 (changes requested):
Forgot to request changes for the sprintf.

src/internal/hash.c
DavisVaughan: Thanks!

src/internal/hash.c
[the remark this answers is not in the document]
DavisVaughan: We are going to leave it as is

lionel-:
Thanks everyone!

PR #1071, opened 2020-12-11
advocate use of 'rlang::entrace' directly in docs

kevinushey:
IIUC it's safe to just set

```
options(error = rlang::entrace)
```

and other parts of the codebase recommend + use this, so advocate using that directly here as well.


lionel-:
Thanks!

Even for developers it should be safe as well with https://github.com/r-lib/pkgload/pull/133

PR #1074, opened 2020-12-13
Deparse double and triple colon on the same line

### LINKED ISSUE #1072: expr_deparse() introduces line breaks in inappropriate places
``` r
x <- quote(1 %>%
  base::identity() %>%
  base::identity() %>%
  base::identity() %>%
  base::identity() %>%
  base::identity() %>%
  base::identity() %>%
  base::identity())

rlang::expr_deparse(x)
#> [1] "1 %>% base::identity() %>% base::identity() %>% base::identity() %>% base::"
#> [2] "  identity() %>% base::identity() %>% base::identity() %>% base::identity()"
```

<sup>Created on 2020-12-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

Initially seen in https://github.com/r-lib/testthat/issues/1282

krlmlr:
to avoid parser error when reparsing.

It's using a hammer, let me know if we can do better here.

Closes #1072.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 
 # rlang (development version)
 
+* `expr_deparse()` now correctly wraps code using `::` and `:::` (#1072, @krlmlr).
+
 
 # rlang 0.4.9
 
--- a/R/deparse.R
+++ b/R/deparse.R
@@ -358,6 +358,13 @@ spaced_op_deparse <- function(x, lines = new_lines()) {
 unspaced_op_deparse <- function(x, lines = new_lines()) {
   binary_op_deparse(x, lines, space = "")
 }
+tight_op_deparse <- function(x, lines = new_lines()) {
+  sub_lines <- new_lines(width = Inf)
+  out <- binary_op_deparse(x, sub_lines, space = "")
+
+  lines$push(out)
+  lines$get_lines()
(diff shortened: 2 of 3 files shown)


lionel-:
Could you please see if it's possible to:

- Take a `sticky_rhs` parameter in `binary_op_deparse()`
- Use `lines$make_next_sticky()` somewhere appropriate

krlmlr:
I tried briefly, in this case the test case `id_fun <- ...` is deparsed on one line, violating width constraints. Will wrapping occur after content has been added via `lines$push_sticky()`?

krlmlr:
Am I missing a problem/test case with the implementation?

krlmlr:
My attempts are in 785f87b5ed61c2eb67f98f507a33cbe0c2047b23.

lionel-:
It's alright if `<-` ends up on the same line. Probably the engine should avoid eager deparsing since we don't know the width of things in advance.

I would prefer the implementation of `::` and `:::` to be consistent with the other methods because that makes the code simpler and that might simplify future improvements to the engine.

krlmlr:
Done. Test failures were false positives, triggered a new run.

lionel-:
Thanks!

PR #1086, opened 2021-01-04
Ensure hashes are independent of endianness

DavisVaughan:
Closes #1084 

@QuLogic do the tests pass without modification on big-endian systems now?

@easyaspi314 if you get a moment, could you please comment on whether this looks appropriate as a way to ensure that the resulting hash is independent of endianness? Thanks!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # rlang (development version)
 
+* `hash()` results are now independent of endianness (#1086).
+
 * Fixed a gcc11 warning related to `hash()` (#1088).
 
 * `XXH3_64bits()` from the XXHash library is now exposed as C callable
--- a/src/internal/hash.c
+++ b/src/internal/hash.c
@@ -31,7 +31,7 @@
  * Before any R object data is serialized, `R_Serialize()` will first write out:
  *
  * Serialization info:
- * - 2 bytes for `"X\n"` to declare "binary" serialization (i.e. not "ascii")
+ * - 2 bytes for `"X\n"` to declare "xdr" serialization (i.e. not "ascii" or "binary")
  * - An `int` representing the serialization version
  * - An `int` representing `R_VERSION`
  * - An `int` representing the minimum R version where this serialization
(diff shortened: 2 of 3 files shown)


easyaspi314:
Yes, that looks correct.

QuLogic:
Sorry, unfortunately that is not enough. The problem is that the serialization itself is not endian-independent. From the comment, I assumed that was intentional:
https://github.com/r-lib/rlang/blob/316ee6860bd41bb23d40af430aa396168bedfe46/src/internal/hash.c#L113-L116

DavisVaughan:
So, it is somewhat complicated to implement endian independent hashes.

For serialization version 3, we need to decode the `int` corresponding to the native encoding size. This would require the use of `R_XDRDecodeInteger()`, which I started to use but then realized it is a hidden function that shouldn't be used. To work around this, I took a look at its implementation and saw that it is really just a call to `ntohl()` to do a byte swap when required, so I've re-implemented that here. I think this is a standard function on most OSes (maybe?), but I'm using the version that R uses.

The good news is that, on my little-endian Mac, we now get the hash values that were reported for the big endian system in https://github.com/r-lib/rlang/pull/1084/files

The bad news is that it does have a pretty big impact on performance.

``` r
library(rlang)

x <- 1:1e6 + 0L
y <- as.list(x)

bench::mark(
  hash(x),
  hash(y),
  check = FALSE,
  iterations = 50
)

# Before
#> # A tibble: 2 x 6
#> 1 hash(x)     198.6µs  202.7µs    4699.    14.95KB        0
#> 2 hash(y)      47.3ms   49.7ms      20.0    8.63KB        0

# After
#> # A tibble: 2 x 6
#> 1 hash(x)      6.24ms   6.52ms     153.    14.95KB        0
#> 2 hash(y)     82.41ms  86.48ms      11.5    8.63KB        0
```

The other bad news is that this obviously adds some complexity.

We could merge this, or we could state that the hashes are guaranteed to be reproducible within the same endianness.

lionel-:
I think you're right, ensuring stability across endianness is not worth the added complexity and the performance penalty.

lionel-:
Though maybe check the approach in https://commandcenter.blogspot.com/2012/04/byte-order-fallacy.html

DavisVaughan:
That actually worked great!

So now the main question is one of performance, since the implementation is much easier to explain now.

I am also using `R_XDR_INTEGER_SIZE`, a define that is defined in `Rinternals.h` and is always set to `4` since they use an `int32_t` when doing XDR encoding/decoding. It is right next to a group of functions that have all been "hidden" on Windows (like the `R_XDRDecodeInteger()` helper I was going to use). Is it safe to use this define or should I just replace it with a 4 here?
https://github.com/wch/r-source/blob/a1425adea54bcc98eef86081522b5dbb3e149cdc/src/include/Rinternals.h#L1358-L1364

PR #1088, opened 2021-01-05
Use `memcpy()` to copy over the native encoding size

DavisVaughan:
Hopefully fixes this gcc11 warning:
https://www.stats.ox.ac.uk/pub/bdr/gcc11/rlang.out

I think the problem comes with casting the `void*` to a `int*` to extract out the `int` representing the native encoding size that we know is there. The `void*` is probably really a `unsigned char*` under the hood, and gcc11 is probably picking up that it looks like we are trying to extract an `int` from an `unsigned char*`?

The fix here copies what R does in `R_Unserialize()` to get at that native encoding size. They use a `memcpy()` in `InBytes()` to just copy over the bytes directly into an `int`.
https://github.com/wch/r-source/blob/d22ee2fc0dc8142b23eed9f46edf76ea9d3ca69a/src/main/serialize.c#L421
https://github.com/wch/r-source/blob/d22ee2fc0dc8142b23eed9f46edf76ea9d3ca69a/src/main/serialize.c#L2767

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # rlang (development version)
 
+* Fixed a gcc11 warning related to `hash()` (#1088).
+
 * `XXH3_64bits()` from the XXHash library is now exposed as C callable
   under the name `rlang_xxh3_64bits()`.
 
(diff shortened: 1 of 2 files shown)


lionel-:
> The void* is probably really a unsigned char* under the hood, and gcc11 is probably picking up that it looks like we are trying to extract an int from an unsigned char*?

The void* is really an int though, right? Otherwise it's not correct to interpret that memory range as an int.

DavisVaughan:
Yea it is this integer here:
https://github.com/wch/r-source/blob/d22ee2fc0dc8142b23eed9f46edf76ea9d3ca69a/src/main/serialize.c#L1397

lionel- (approved):

PR #1096, opened 2021-01-21
Preserve objects in a namespace-bound dictionary

lionel-:
* `r_preserve()` now uses a dictionary to preserve objects. This should guarantee constant time insertion and deletion.

* The dictionary is protected from the embedding namespace. This way, unloading the namespace automatically unprotects all preserved objects.

* `r_init_library()` now takes a namespace as argument. This should be the embedding namespace not the rlang namespace. The namespace must be unsealed at the time of calling because a new binding is added to store the dictionary of precious objects.

* Add `dict_del()` method to remove objects from the dictionary. This is used in `r_unpreserve()`.

### DIFF
--- a/R/c-lib.R
+++ b/R/c-lib.R
@@ -151,23 +151,21 @@ detect_rlang_lib_usage <- function(src_path) {
 new_dict <- function(size, prevent_resize = FALSE) {
   .Call(rlang_new_dict, size, prevent_resize)
 }
-
 dict_size <- function(dict) {
   length(dict[[2]][[1]])
 }
-
 dict_resize <- function(dict, size) {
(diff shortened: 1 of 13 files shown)


DavisVaughan (approved):
Very cool! I did a full review and it all looks good to me. `dict_del()` is also nice

lionel-:
The logic was not totally right. If `r_preserve()` and `r_unpreserve()` were called on the same object in different contexts, the object would be removed from the dictionary too early. This is fixed in #1097 which makes the precious dict a dictionary of protection stacks.

lionel-:
oops and of course the input needed to be protected https://github.com/r-lib/rlang/commit/a999e43.

PR #1099, opened 2021-01-25
Add data frame ctors to the C library

lionel-:


src/rlang/df.c
```diff
@@ -0,0 +1,70 @@
...
+#include "decl/df-decl.h"
+
+
+sexp* r_alloc_data_frame(r_ssize n_rows,
+                         sexp* names,
+                         enum r_type* v_types, r_ssize types_size) {
+  sexp* df = KEEP(r_new_list(types_size));
+
+  if (r_typeof(names) != r_type_character) {
+    r_abort("`names` must be a character vector.");
```
DavisVaughan, line 15: Not sure if you care about `NA_character_` names here or not

src/rlang/df.c
```diff
@@ -0,0 +1,70 @@
...
+  FREE(1);
+}
+
+static
+sexp* new_compact_rownames(r_ssize n_rows) {
+  if (n_rows <= 0) {
+    return r_ints_empty;
+  }
+
+  sexp* out = r_new_integer(2);
```
DavisVaughan, line 55: I guess not protecting this is safe, but it still caught my eye

DavisVaughan (approved):
Is there also room for a `r_new_data_frame(data, n_rows)` that lets you bring your own named list of `data`? Maybe the pattern that you have is more useful, since generally you'd use `r_alloc_data_frame()`  _instead_ of allocating a named list and passing it to `r_new_data_frame()`.

Plus, with `r_new_data_frame()` you'd have to check the length of each element to ensure the list is rectangular, which could get messy if you have to do a callback to R's `length()`

tests/testthat/test-c-api.R
```diff
@@ -578,3 +578,19 @@ test_that("can preserve and unpreserve repeatedly", {
...
+  df <- alloc_data_frame(2L, c("a", "b", "c"), c(13L, 14L, 16L))
+
+  expect_equal(nrow(df), 2)
+  expect_equal(ncol(df), 3)
+  expect_equal(class(df), "data.frame")
+  expect_equal(names(df), c("a", "b", "c"))
+  expect_equal(lapply(df, typeof), list(a = "integer", b = "double", c = "character"))
+  expect_equal(lapply(df, length), list(a = 2, b = 2, c = 2))
+
+  df <- alloc_data_frame(0L, chr(), int())
```
DavisVaughan, line 592: You can allocate a data frame with `>0` rows but `0` cols with this right?

src/rlang/df.c
lionel-: hmm since the names are provided by a developer rather than a user I'd prefer to avoid doing extra checks.

tests/testthat/test-c-api.R
lionel-: yup, added a test

lionel-:
> Is there also room for a r_new_data_frame(data, n_rows) that lets you bring your own named list of data?

In that case you'd use `r_init_data_frame()` on your named list. The caller may clone the named list if needed.

Merging now but let me know if I'm missing something. Thanks for reviewing!

PR #1107, opened 2021-02-03
Dynamic arrays

lionel-:
This adds an API for dynamic arrays to the C library. Currently that's only for C-level arrays but we could add similar APIs for integers, doubles, etc. Example usage:

```c
struct snapshot_data {
  r_ssize retained_size;
  r_ssize retained_count;
};

struct r_dyn_array* p_data_arr = r_new_dyn_array(DATA_INIT_SIZE, sizeof(struct snapshot_data));
KEEP(p_data_arr->shelter);

// Push a zero-initialised element
r_arr_push_back(p_data_arr, 0);

// Push actual data
struct snapshot_data data = { .retained_size = 1, .retained_count = 2 };
r_arr_push_back(p_data_arr, &data)
```

- Genericity is achieved by taking void pointers and sizes.
- The growable array is allocated in a separate raw vector. This way it can be grown independently and pointers to `struct r_dyn_array` stay valid at all times.
- Both raw vectors are stored in a list shelter. Only this list needs to be protected by callers.


lionel-:
Merging now but please feel free to comment on the API or implementation @DavisVaughan :)

lionel-:
We might want to rename the prefix from `arr` to `dyn_arr` when we add the other dynamic vectors. I imagine those will have prefixes like `dyn_dbl` or `dyn_lgl` to disambiguate. E.g. `r_lgl_resize()` resizes an actual logical vector, whereas `r_dyn_lgl_resize()` would resize a dynamic logical vector.

src/rlang/dyn-array.c
```diff
@@ -0,0 +1,71 @@
...
+
+  if (p_elt) {
+    memcpy(r_arr_ptr_back(p_arr), p_elt, p_arr->elt_byte_size);
+  } else {
+    memset(r_arr_ptr_back(p_arr), 0, p_arr->elt_byte_size);
+  }
+}
+
+void r_arr_grow(struct r_dyn_array* p_arr,
+                r_ssize capacity) {
```
DavisVaughan, line 46: Is this supposed to be `capacity`? Or `count`? I guess you don't even really need to pass this in since you can get it from `p_arr->count`, but that doesnt matter too much

src/rlang/dyn-array.c
```diff
@@ -0,0 +1,71 @@
...
+  p_arr->v_data = r_raw_deref(arr_raw);
+  p_arr->elt_byte_size = elt_byte_size;
+
+  FREE(1);
+  return p_arr;
+}
+
+void r_arr_push_back(struct r_dyn_array* p_arr, void* p_elt) {
+  r_ssize count = ++p_arr->count;
+  r_arr_grow(p_arr, count);
```
DavisVaughan, line 36: There may be some performance advantage to inlining the condition of the most common case to avoid a function call:

```c
if (capacity <= p_arr->capacity) {
    return;
}
```

It made a difference when working on vec-order, which is the only reason I bring it up

src/rlang/dyn-array.c
```diff
@@ -0,0 +1,71 @@
...
+  r_arr_resize(p_arr, new_capacity);
+}
+
+void r_arr_resize(struct r_dyn_array* p_arr,
+                  r_ssize capacity) {
+  sexp* arr_raw = r_raw_resize(r_list_get(p_arr->shelter, 1),
+                               p_arr->elt_byte_size * capacity);
+  r_list_poke(p_arr->shelter, 1, arr_raw);
+
+  p_arr->count = r_ssize_min(p_arr->count, capacity);
```
DavisVaughan, line 62: I guess this min matters because you could, in theory, shrink the capacity?

src/rlang/dyn-array.h
```diff
@@ -0,0 +1,50 @@
...
+static inline
+void* r_arr_ptr_front(struct r_dyn_array* p_arr) {
+  return p_arr->v_data;
+}
+static inline
+void* r_arr_ptr_back(struct r_dyn_array* p_arr) {
+  return r_arr_ptr(p_arr, p_arr->count - 1);
+}
+static inline
+void* r_arr_ptr_end(struct r_dyn_array* p_arr) {
```
DavisVaughan, line 44: When is this useful?

tests/testthat/test-c-api.R
```diff
@@ -634,3 +634,91 @@ test_that("can shrink vectors", {
...
+
+  arr_pop_back(arr)
+  expect_equal(
+    arr_info(arr)[1:2],
+    list(
+      count = 5,
+      capacity = 6
+    )
+  )
+  expect_equal(arr[[2]], exp)
```
DavisVaughan, line 694: It is neat that pop-back doesn't have to clear out the old info, it just shifts the positioning backwards by 1

DavisVaughan (commented):
Looks useful!

src/rlang/dyn-array.c
lionel-: yup

src/rlang/dyn-array.h
lionel-: I added this in case someone wants to use C++-style iteration bounds. Also it helps making a semantic difference between "front" and "end". It's not an important part of the API and I don't expect I'll use it myself though.

src/rlang/dyn-array.c
lionel-: Given this and your other comment, I've now inlined `r_arr_grow()` inside `r_arr_push_back()` since it doesn't seem that useful: https://github.com/r-lib/rlang/commit/2aab3e12e2982d687208e312c018afbf44908852

lionel-:
Thanks for the review!

PR #1110, opened 2021-02-04
Support atomic and barrier vectors in dynamic arrays API

lionel-:
Turns out it was easy to adapt the dynamic arrays implemented in #1107 to support all sorts of vectors.

### DIFF
--- a/src/rlang/vec.h
+++ b/src/rlang/vec.h
@@ -78,6 +78,58 @@ sexp* const * r_list_deref_const(sexp* x) {
 }
 
 static inline
+void* r_vec_deref0(enum r_type type, sexp* x) {
+  switch (type) {
+  case r_type_logical: return r_lgl_deref(x);
+  case r_type_integer: return r_int_deref(x);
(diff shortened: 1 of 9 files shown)


src/rlang/vec.h
```diff
@@ -77,6 +77,58 @@ sexp* const * r_list_deref_const(sexp* x) {
...
+  default: r_stop_unimplemented_type("r_vec_deref_const", type);
+  }
+}
+static inline
+const void* r_vec_deref_const(sexp* x) {
+  return r_vec_deref_const0(r_typeof(x), x);
+}
+
+static inline
+int r_vec_elt_sizeof0(enum r_type type) {
```
DavisVaughan, line 115: `r_type_sizeof()`?

lionel-: oh yes

lionel-: Though that only makes sense for vectors and maybe it's best to keep the name symmetry.

PR #1131, opened 2021-03-14
Add tick as a bullet type in format_error_bullets()

rossellhayes:
Packages generally prefix messages with one of four bullets: a star, an info "i", a cross, and a tick.
For example, `usethis` offers `ui_todo()`, `ui_info()`, `ui_oops()` and `ui_done()`.
Currently, `format_error_bullets()` only offers a star, info symbol and cross.

This PR adds a green tick bullet when the element name is `"v"`, matching current usage where a cross is used for elements named `"x"` and an info symbol for elements named `"i"`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # rlang (development version)
 
+* `format_error_bullets()` now adds green tick bullets for elements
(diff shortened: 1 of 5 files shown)


lionel-:
Thanks!

Issue #1090, opened 2021-01-11
Support glue in `abort()`


lionel-:

lionel-:
@hadley I think the main problem is how do we find the correct environment when `abort()` is wrapped? I guess we would need to use `glue()` explicitly at each wrapper level:

```r
my_function <- function(bar) {
  abort("foo {bar}")
}

my_function <- function(bar) {
  stop_bar("foo {bar}")
}
stop_bar <- function(msg) {
  abort(glue(msg, envir = caller_env())
}
```

tzakharko:
@lionel- What you could do (although not sure how robust it is) is chase down the argument promise until you get a character constant. A quick and dirty non-optimal implementation: 

```c
SEXP glue_constant_arg(SEXP call, SEXP op, SEXP args, SEXP rho) { 
  // the sole argument is the argument name
  SEXP name = CADR(args);
  if(TYPEOF(name) != SYMSXP) error("name must be a symbol");

  SEXP maybe_promise = Rf_findVarInFrame(rho, name);
  // chase the unevaluated promise down the symbol chain until we get to a value
  while(
    TYPEOF(maybe_promise) == PROMSXP && 
    PRENV(maybe_promise) != R_NilValue &&
    TYPEOF(PREXPR(maybe_promise)) == SYMSXP
  ) {
    maybe_promise = Rf_findVarInFrame(PRENV(maybe_promise), PREXPR(maybe_promise));  
  }

  // bail out unless it's an unevaluated promise that holds a character const
  if( 
    TYPEOF(maybe_promise) != PROMSXP ||
    PRENV(maybe_promise) == R_NilValue ||
    TYPEOF(PREXPR(maybe_promise)) != STRSXP
  ) {
    // evaluate it in case it's a promise
    return Rf_eval(maybe_promise, R_EmptyEnv);
  }

  // we have found our character constant, pass it to glue
  SEXP glue_call = PROTECT(
    lang3(install("glue"), 
    maybe_promise, 
    PRENV(maybe_promise))
  );
  SET_TAG(CDDR(glue_call), install(".envir")); 
  UNPROTECT(1);

  return Rf_eval(glue_call, rho);
} 
```

and then

```R
f <- function(msg) {
  .External2("glue_constant_arg", sym("msg"))
}

g <- function(x) {
  f(x)
}

local({
  x <- 15
  g("hello {x}")
})
```

Few problems with this approach:
[...]

PR #1144, opened 2021-03-30
Add option to break line in formatted bullets

### LINKED ISSUE #1130: Provide way to opt out of bullet
i.e. some convention that displays it as a regular unindented line

lionel-:
Closes #1130


NEWS.md
```diff
@@ -1,7 +1,7 @@
 # rlang (development version)
 
 * `format_error_bullets()` now adds green tick bullets for elements
-  named `"v"` (@rossellhayes).
+  named `"v"` (@rossellhayes) and line breaks for elements named `" "` (#1130).
```
hadley, line 4: Start with an empty line?

I think we also need some way to display a unindented line.

lionel-: You can include a `\n` character in the bullet to start on an empty line. I'll add examples to showcase the difference.

jennybc:
It will be great to be able to print a "non-bullet" bullet (which is how I think of this feature request)!

Made with this PR:

``` r
x <- c(
        "alfa",
        "bravo",
  v   = "charlie",
  i   = "delta",
  x   = "echo",
  ` ` = "foxtrot",
        "golf"
)

rlang::abort(x)
```

Screen Shot 2021-03-31 at 4 58 50 PM

I wish things were like so:

* A lack of a name (truly no name or name is empty string) corresponds to a lack of bullet symbol. Or, said another way, the bullet symbol is a single space.
* Small set of allowed names map to specific bullet symbols:

| name            | bullet               |
|-----------------|----------------------|
| i               | ℹ                    |
| x               | x                    |
| v               | ✓                    |
| *               | ●                    |
| `""` or no name | &lt;single space&gt; |

I realize I am challenging rlang's current behaviour, where "no name" means the bullet symbol is `*`. But this scheme just feels more logical to me. It may be a nonstarter, but I can't pass up the chance to suggest it. (BTW why not use the `cli::symbol$bullet` symbol when possible, similar to all the others?)

My goal is to try to persuade you and @gaborcsardi to enough convergent evolution that the same `message` vector provided to `rlang::abort()` (and friends) and to `cli::cli_XXX()` (whatever new cli function eventually comes from https://github.com/r-lib/cli/pull/234) are formatted the same way. I doubt we can get 100% there, but I'd love to get as far as we can.

lionel-:
> BTW why not use the cli::symbol$bullet

Good point.

> I realize I am challenging rlang's current behaviour, where "no name" means the bullet symbol is 

One possible way forward is to deprecate `format_error_bullets()` and create a new `format_bullets()` function with a more general name. But that still causes a big breaking change for `abort()` which wraps `format_error_bullets()`.. `" "` is less practical and logical but it is backwards compatible.

On the other hand, bullet lists are the least common case and getting bullet-less lists until updating the format is not exactly a critical bug.

lionel-:
The other thing to take into account is frequency of bullet type. I would expect break lines to be much rarer than bullets, so making the latter the default might make more sense. Especially since bullets often appear in bunch (e.g. list of errors) whereas break lines are inserted independently.

hadley:
What if a complete unnamed character vector produced bullets, but as soon as you named one bullet, you had to name them all? 

```R
c("a", "b")
#> * a
#> * b

c("*" = "a", "b")
#> * a
#> b
```

I think this probably wouldn't affect too many existing errors, since the first element doesn't get a bullet anyway.

lionel-:
Good idea.

jennybc:
I like @hadley's proposal!  But, if I'm successful with my campaign that no name or empty-string-is-name, it would actually unfold like this:

``` r
c("a", "b")
#> * a
#> * b

c("*" = "a", "b")
#> * a
#>   b
```

Would anyone ever really want the `b` against the left margin (?).

Or I guess the single-space-as-name could be required, in which case we get:

``` r
c("*" = "a", "b", ` ` = "c")
#> * a
#> b
#>   c
```

lionel-:
> Would anyone ever really want the b against the left margin (?).

Oh I think this should be handled with a `"\n"` character in the bullet as showcased in the examples. So an empty name would cause an indented line break.

jennybc:
> I think this should be handled with a `"\n"`

Part of what I like about where both `abort()` and much of cli are going is this idea of a "semantic interface", to use cli's terminology. I love getting away from dealing with explicit `"\n"`s, relying more on the structure of the `message` to convey formatting. But I can also see that `abort()` does not (can not) embrace this mentality to the extent that cli does.

hadley:
I'm with @jennybc and really like the idea what you (generally) supply these functions with a vector, where each element represents one line.

jennybc:
I just came across this live example of a message that would be nicer to format with changes proposed in this PR:

``` r
msg <- paste(
  "Attempt to directly access a credential that can only be used within tidyverse packages.",
  "This error may mean that you need to:",
  "  * Create a new project on Google Cloud Platform",
  "  * Enable relevant APIs for your project",
  "  * Create an API key and/or an OAuth client ID",
  "  * Configure your requests to use your API key and OAuth client ID",
  sep = "\n"
)
rlang::abort(msg)
#> Error: Attempt to directly access a credential that can only be used within tidyverse packages.
#> This error may mean that you need to:
#>   * Create a new project on Google Cloud Platform
#>   * Enable relevant APIs for your project
#>   * Create an API key and/or an OAuth client ID
#>   * Configure your requests to use your API key and OAuth client ID
```

<sup>Created on 2021-04-01 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0.9002)</sup>

It's actually a good example of wanting one line tight against the left margin, then bullets. So I guess I see a use for a line with no bullet and no indentation!

hadley:
So we're suggesting it would instead look like this?

```R
rlang::abort(c(
  "Attempt to directly access a credential that can only be used within tidyverse packages.",
  "This error may mean that you need to:",
  "*" = "Create a new project on Google Cloud Platform",
  "*" = "Enable relevant APIs for your project",
  "*" = "Create an API key and/or an OAuth client ID",
  "*" = "Configure your requests to use your API key and OAuth client ID",
))
```

jennybc:
Yeah.

But I've also discovered a decent way to get this now, given the `abort()` wrapper that I seem to be creating in every package (`abort()` + calling `glue()` on each element of `message`).

``` r
msg <- c(
  "Attempt to directly access a credential that can only be used within \\
   tidyverse packages.
   This error may mean that you need to:",
  "Create a new project on Google Cloud Platform",
  "Enable relevant APIs for your project",
  "Create an API key and/or an OAuth client ID",
  "Configure your requests to use your API key and OAuth client ID"
)
gargle:::gargle_abort(msg)
#> Error: Attempt to directly access a credential that can only be used within tidyverse packages.
#> This error may mean that you need to:
#> * Create a new project on Google Cloud Platform
#> * Enable relevant APIs for your project
#> * Create an API key and/or an OAuth client ID
#> * Configure your requests to use your API key and OAuth client ID

# here's what gargle_abort() looks like:
# gargle_abort <- function(message, ..., class = NULL, .envir = parent.frame()) {
#   g <- function(line) glue(line, .envir = .envir)
#   msg <- map_chr(message, g)
#   abort(msg, class = c(class, "gargle_error"), ...)
# }
```

<sup>Created on 2021-04-01 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0.9002)</sup>

hadley:
You're not proposing that as a syntax, right? Just showing what you're doing now.

jennybc:
Correct. I'm just taking advantage of something my `glue()` usage allows.

Although I think such a `PKG_abort()` wrapper might be rather common, in order to add a class, and then it's a nice place to give yourself glue (or cli) support for all error message bits.

hadley:
I thought we discussed the idea of classing all error from a package, and came to the conclusion that it wasn't usually worth it? So hopefully we can make it uncommon for packages to need their own `abort()` wrappers.

jennybc:
> hopefully we can make it uncommon for packages to need their own `abort()` wrappers

Hmmm, I use `glue()` on such a high proportion of my error message lines that a wrapper feels worth it.

hadley:
@jennybc I think we want either rlang or cli to provide a glue + abort wrapper for us.

jennybc:
> I think we want either rlang or cli to provide a glue + abort wrapper for us.

Yes, but I assume that's out of scope here (in this particular PR).

lionel-:
Adding `glue()` support is planned for this release #1090.

Thanks for the useful discussion. I'll change this PR so that `""` is an unindented line break and `" "` an indented one.

jennybc:
Bringing up something that got buried above:

What do you think of adding `"!"` as a supported bullet? To align with `cli::cli_alert_info()`.

lionel-:
> What do you think of adding "!" as a supported bullet?

That makes sense. I think we should add all the commonly used bullets and rename `format_error_bullets()` to `format_bullets()`.

lionel-:
Do we really want to make it easier to write this:

```
* First bullet.
Line break.
* Another bullet.
```

Instead of this?

```
* First bullet.
  Line break.
* Another bullet.
```

I would think the first form is bad typographical form and should not be encouraged. It can still be achieved with an explicit `"\n"`.

On the other hand we need to allow a first unindented line. The first unnamed elements could be special-cased this way:

```r
abort(c(
  "Header line",
  "Header line break",
  "*" = "Bullet",
  "Line break"
))
#> Error: Header line.
#> Header line break.
#> * Bullet.
#>   Line break.
```

The idea is that opening a bullet list introduces an indentation level. Does this make sense?

hadley:
I think the problem is that sometimes we want multiple (well two, probably) bulleted lists:

```
#> Error: Main header
#> Header1
#> * Bullet.
#> * Bullet.
#>
#> Header 2
#> * Another bullet
```

lionel-:
hmm in that case I think you could call `format_bullets()` twice. The code would be clearer as well.

hadley:
Yeah, that's true. In that case, your syntax is rather appealing.

lionel-:
I guess it's nice to inline everything in the `abort()` call:

```r
abort(c(
  "Main header.",
  "Header 1",
  x = "Bullet 1",
  x = "Bullet 2",
  " ",
  "Header 2",
  x = "Bullet 3",
  x = "Bullet 4"
))
```

What if the `" "` line created a separation between bullet lists? The indentation level would go back to zero and leading unnamed elements would be unindented.

This way `format_bullets()` would be in charge of formatting the separation (do we want an empty line or not).

lionel-:
Though with `paste_line()` inlining multiple `format_bullets()` calls is fine I think:

```r
abort(paste_line(
  "Main header.",
  format_bullets(
    "Header 1",
    x = "Bullet 1",
    x = "Bullet 2"
  ),
  format_bullets(
    "Header 2",
    x = "Bullet 3",
    x = "Bullet 4"
  )
))
```

lionel-:
It looks like a combination of `format_bullets()` implemented as in https://github.com/r-lib/rlang/pull/1144#issuecomment-817788970 and `paste_line()` should cover all these cases, minus the two-column alignment of `:` lists which is probably out of scope.

lionel-:
In the end I went for the simpler syntax of `cli_memo()`:

- No name means a line break with no indentation
- `" "` means an indented line break.

I changed my mind because between

```r
abort(c(
  "Header line",
  "Header line break",
  "*" = "Bullet",
  "Line break"
))
```
and
```r
abort(c(
  "Header line",
  "Header line break",
  "*" = "Bullet",
  " " = "Line break"
))
```

The second form looks more like the final output:

```r
#> Error: Header line.
#> Header line break.
#> * Bullet.
#>   Line break.
```

I also added support for `"!"` and renamed `format_error_bullets()` to the more general name `format_bullets()`.

PR #1148, opened 2021-04-01
Mention supplied argument in `arg_match()`

### LINKED ISSUE #1113: arg_match should include actual value in error
```R
arg_match0("x", c("y", "z"), "arg")
#> Error: `arg` must be one of "y" or "z".
#->
#> Error: `arg` must be one of "y" or "z"; not "x"
```

lionel-:
Closes #1113

### DIFF
--- a/tests/testthat/_snaps/arg.md
+++ b/tests/testthat/_snaps/arg.md
@@ -1,35 +1,35 @@
 # `arg_match()` has informative error messages
 
     Code
-      (expect_error(arg_match0("continuuos", c("discrete", "continuous"))))
+      (expect_error(arg_match0("continuuos", c("discrete", "continuous"), "my_arg")))
     Output
       <error/rlang_error>
-      `"continuuos"` must be one of "discrete" or "continuous".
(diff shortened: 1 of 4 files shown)


tests/testthat/_snaps/arg.md
```diff
@@ -5,25 +5,29 @@
     Output
       <error/rlang_error>
       `"continuuos"` must be one of "discrete" or "continuous".
-      Did you mean "continuous"?
+      x Input is "continuuos".
```
hadley, line 8: In this case, I think we could add it to the end of the first sentence:

`"continuuos"` must be one of "discrete" or "continuous"; not "continuuos"

And maybe rework the tests so they always use a variable? (Since that will be the case you actually see in real life):

`x` must be one of "discrete" or "continuous"; not "continuuos"

hadley: @batpigandme might have thoughts on the punctuation here.

lionel-: Both done. A google search about suggests using a comma rather than a semi-colon is the correct punctuation.

PR #1162, opened 2021-04-15
compat-purrr.R tweaks

### LINKED ISSUE #1157: Use `as_function()` in compat-purrr

### LINKED ISSUE #1159: Remove pluck_* from compat-purrr
Since it's a weird old version

hadley:
* Remove pluck functions (fixes #1159)
* Remove `map_cpl()` and `map2_cpl()`
* Use `as_function()` (fixes #1157)

I didn't namespace because all the options I tried looked ugly.

### DIFF
--- a/R/compat-purrr.R
+++ b/R/compat-purrr.R
@@ -1,10 +1,16 @@
-# nocov start - compat-purrr (last updated: rlang 0.3.2.9000)
+# nocov start - compat-purrr.R
+# Latest version: https://github.com/r-lib/rlang/blob/master/R/compat-purrr.R
 
-# This file serves as a reference for compatibility functions for
-# purrr. They are not drop-in replacements but allow a similar style
-# of programming. This is useful in cases where purrr is too heavy a
-# package to depend on. Please find the most recent version in rlang's
-# repository.
+# This file provides a minimal shim to provide a purrr-like API on top of
+# base R functions. They are not drop-in replacements but allow a similar style
+# of programming.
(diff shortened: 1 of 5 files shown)


R/compat-purrr.R
```diff
@@ -7,9 +7,16 @@
...
+  .f <- as_function(.f)
   lapply(.x, .f, ...)
 }
+walk <- function(.x, .f, ...) {
+  map(.x, .f, ...)
+  invisible(.x)
+}
+
 map_mold <- function(.x, .f, .mold, ...) {
+  .f <- as_function(.f)
```
lionel-, line 19: Should either remove string support or pass an environment to `as_function()`. I would remove string support since I regret that feature and threading the environment is more verbose (need to call `as_function()` at each entry point). In this case we need to pass the `[[` function instead of its name below.

hadley: I think we're expecting these functions to be used internally, so the default lookup should be ok?

lionel-: I'd rather not encourage function lookup via strings because of the scoping issues involved.

hadley: The problem is that not allowing strings is going to make the API rather different to both base R and purrr. I think we're less likely to have problems allowing strings with slightly wrong lookup than disallowing them.

hadley: That said, I don't mind if you want to implement it; but I'm not going to 😄

PR #1164, opened 2021-04-16
Add `call_match()` and support for matching missing arguments

### LINKED ISSUE #875: Function for call_standarize to also add formals
The [`gradethis`](https://github.com/rstudio-education/gradethis) package needed a way to standarise call but also include the default function arguments so code written as `mean(c(1:3)` and `mean(1:3, na.rm = TRUE)` are marked as being the same.

We made a PR into `gradethis` that implements this behavior (https://github.com/rstudio-education/gradethis/pull/74) along with unit tests. Would it make sense to submit this function as a PR into `rlang`?

I've reproduced the function below for convenience.

```r
call_standardise_formals <- function(code, env = rlang::current_env()) {
  # try to catch invalid function, i.e., corrupt language object
  tryCatch({
    fxn <- rlang::call_fn(code, env = env)
  }, error = function(e) {
    return(code)
  })
  if (!exists("fxn")) {return(code)} ## some reason the above tryCatch doesn't go to the error part
  if(class(fxn) != "function") {return(code)}
  
  # standarise, but dont bother trying to fill out default formals
  # for primitives like mean, unable to distinguish between mean and mean.default
  if (is_infix(code) || is.primitive(fxn)) {
    return(rlang::call_standardise(code))
  }
  
  forms <- rlang::fn_fmls(fxn)
  default_params <- forms[!vapply(forms, is.symbol, logical(1), USE.NAMES = FALSE)]
  
  code_std <- rlang::call_standardise(code, env = env) # order and label existing params
  
  code_params <- rlang::call_args(code_std) # get arguments passed from user
  code_missing_default_args <- default_params[!names(default_params) %in% names(code_params)]
  if (length(code_missing_default_args) == 0) {
    return(code_std)
  }
  return(
    rlang::call_standardise(rlang::call_modify(code_std, !!!code_missing_default_args),
                            env = env)
  )
}
```

lionel-:
- Add `call_match()` as a bare bones version of `call_standardise()`. It supports matching missing arguments to their defaults in the function definition. It is also a little stricter than `match.call()`: `fn` and `dots_env` are only inferred from the call stack when `expr` is.

- Add `dots_env` and `defaults` arguments to `call_standardise()`.

Default matching is disabled by default for consistency with `match.call()` and backward compatibility with `call_standardise()`'s historical behaviour.

Closes #875.


R/call.R
```diff
@@ -651,12 +661,129 @@ call_standardise <- function(call, env = caller_env()) {
...
+    defaults = defaults,
+    dots_env = dots_env
+  )
+  set_expr(call, matched)
+}
+
+#' Match supplied arguments to function definition
+#'
+#' @description
+#' `call_match()` is like [match.call()] with these differences:
```
hadley, line 676: Would it make sense to document `call_match()` and `call_standardise()` together? I don't quite understand what the difference is, and I'm kind of surprised that `call_standardise()` calls `call_match()`, rather than the other way around

lionel-: I thought it'd be confusing to document them together since they are so close. I think the difference is that `call_standardise()` is supposed to be a little more user friendly because it's intended to work with a single argument (a call) rather than two or three (call, definition, dots environment).

Personally I find its UI a bit dangerous regarding the scoping, it's easy to make logical mistakes when not careful. It's inherited from lazyeval where the assumption was that most calls would be wrapped in formulas, thus carrying an environment. With rlang it seemed like the same principle would apply with quosures. In this context, supplying a function definition is superfluous because it can be inferred from the environment of the formula or quosure. However time has shown that we rarely work with wrapped expressions but with bare expressions, and then not requiring an `env` or `fn` argument is not ideal.

That's why `call_match()` takes a different approach.

- Unlike `match.call()`, it either infers everything from the call stack or nothing.
- Unlike `call_standardise()`, it forces you to supply a function definition.

Maybe we should supersede `call_standardise()`?

hadley: Ah ok, that makes sense to me. You can make the call on whether or not to supersede `call_standardise()`, but if we don't we need to make sure to cross-link the functions.

PR #1166, opened 2021-04-18
Update `arg_match()` to do case-insensitive match.

### LINKED ISSUE #1152: Feature request: enhance typo detection in `arg_match()`
Couple of suggestions how `arg_match()` and `arg_match0()` can be even more useful when detecting typos:

- Offer an appropriate suggestion when typo is base on characters with different cases:

``` r
# This should suggest "ABC"
rlang::arg_match0("abc", "ABC")
#> Error: `"abc"` must be one of "ABC".

# This should suggest "A"
rlang::arg_match0("a", c("A", "B"))
#> Error: `"a"` must be one of "A" or "B".
```

- Always offer suggestion. For example, remove `if` clause [here](https://github.com/r-lib/rlang/blob/160d371e0c6b59c0685d928f82caba84628701e5/R/arg.R#L88-L90) or add `else` which suggested first element in `set`. I understand this might be questionable, but I can see people prefer this.

- If execution speed is a concern here, maybe add argument `suggest = TRUE` to `arg_match()` and `arg_match0()`?

echasnovski:
This closes #1152. I added case-insensitive match only after a failed case-sensitive one as this seems to suggest better candidates overall.

### DIFF
--- a/R/arg.R
+++ b/R/arg.R
@@ -78,6 +78,7 @@ arg_match0 <- function(arg, values, arg_nm = as_label(substitute(arg))) {
 stop_arg_match <- function(arg, values, arg_nm) {
   msg <- arg_match_invalid_msg(arg, values, arg_nm)
 
+  # Try suggest the most probable and helpful candidate value
   candidate <- NULL
   i_partial <- pmatch(arg, values)
   if (!is_na(i_partial)) {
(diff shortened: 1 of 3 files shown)


lionel-:
Thanks!

PR #1167, opened 2021-04-19
Export sexp_address alias and roxygenize

hsbadr:
Fixes https://github.com/r-lib/lifecycle/issues/112


lionel-:
Thanks but the thing is that I don't want to export `sexp_address()` :)

PR #1171, opened 2021-04-20
Bump dev version

schloerke:
In reference to function name change: 
https://github.com/r-lib/rlang/commit/33f12f3b97af38a9b26ef76e46228b58ff15ebb4

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: rlang
-Version: 0.4.10.9000
+Version: 0.4.10.9001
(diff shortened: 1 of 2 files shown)


lionel-:
Thanks! Which package has been importing `sexp_adress()`? We had puzzling reports about importing errors.

schloerke:
The dev version of rstudio/htmltools is using it within `tagQuery()`

lionel-:
An intervening patch release just got on CRAN so you might need to bump your dependency requirement to 0.4.11.9000 @schloerke.

Issue #1178, opened 2021-04-25
rlang fails tests because of missing suggested packages


dmurdoch:
`rlang` is currently failing CRAN tests on the winUCRT platform with these test failures:

```
> f("{nm}" := "value")
Error: Can't use `{` symbols in LHS of `:=` if glue is not installed.
```

and 

```
  > library("testthat")
  Error in library("testthat") : there is no package called 'testthat'
  Execution halted
```

Both `glue` and `testthat` are suggested packages.

These tests should be made conditional on the presence of the suggested packages so that `rlang` reverse dependencies can still be built and tested on CRAN.  If I can get a test platform set up, I'll submit a PR.

dmurdoch:
PR #1179 fixes the issue with glue. 

The issue with testthat is harder:  there doesn't seem to be a way to warn users that testthat is missing other than failing the test, which gives an ERROR and makes all downstream packages from rlang fail:  including testthat.  Ignoring the missing testthat means that other possible errors in rlang will not be detected.

lionel-:
Closing this now until we can figure out a general solution for the hard dependencies of scripts in `tests/`. Happy to revisit once we have one. In the meantime I have fixed another issue when rmarkdown isn't installed (see https://github.com/r-lib/rlang/commit/48fa63a3). Thanks for looking into this @dmurdoch!

PR #1179, opened 2021-04-25
If `glue` is not installed, `rlang` should still be installable

dmurdoch:
Checks should give only NOTEs about missing suggested packages.  

 - Since `roxygen2` needs `glue`, we need to update docs before getting rid of `glue`.
 - Since `testthat` needs `glue`, we also have to assume it isn't installed.  
 - Since `devtools` needs `glue`, we need to do our checks using standard `R CMD check`.

This PR fixes issue #1178 in that `rlang` can pass CRAN checks without ERRORs or WARNINGs if `glue` is missing, but I don't know if it will still install on winUCRT, because I don't have a test platform set up for that yet.

### DIFF
--- a/R/dots.R
+++ b/R/dots.R
@@ -54,8 +54,10 @@
 #' # Unquote a name, showing both the `!!` bang bang and `{}` glue style
 #' nm <- "key"
 #' f(!!nm := "value")
-#' f("{nm}" := "value")
-#' f("prefix_{nm}" := "value")
+#' if (requireNamespace("glue")) {
+#'   f("{nm}" := "value")
+#'   f("prefix_{nm}" := "value")
+#' }
 #'
 #' # Tolerate a trailing comma
 #' f(this = "that", )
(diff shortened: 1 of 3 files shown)


tests/testthat.R
```diff
@@ -1,7 +1,9 @@
 # Workaround for loadNamespace() failure on R 3.2
 requireNamespace("rlang")
 
-library("testthat")
-library("rlang")
+if (require(testthat)) {
```
lionel-, line 4: Can you please remove this and open another PR?

R/dots.R
```diff
@@ -54,8 +54,10 @@
 #' # Unquote a name, showing both the `!!` bang bang and `{}` glue style
 #' nm <- "key"
 #' f(!!nm := "value")
-#' f("{nm}" := "value")
-#' f("prefix_{nm}" := "value")
+#' if (requireNamespace("glue")) {
```
lionel-, line 57: Can you use `is_installed()` please.

dmurdoch: Done

lionel-:
Thanks!

PR #1180, opened 2021-04-26
Implement `hash_file()`

### LINKED ISSUE #1134: `hash_file()`

DavisVaughan:
Closes #1134 

This PR adds `hash_file(path)`.

- It borrows the `fopen()` wrapper from brio to ensure that Windows file paths are opened correctly. I've renamed that `r_fopen()` and used rlang `r_abort()` internally.
- The file is always opened in `"rb"` mode, at @hadley's advice
- The file is read in chunks, with a chunk size of `512 * 1024`. This seems to work well, and is also what Jeroen uses in openssl. I noticed that there was around 1.5x performance improvement in going from a small stack allocated buffer of `1024` to a heap allocated one like this. With minimal testing, it didn't seem to make as much of a difference if the buffer was any larger.

@hadley I do think you can probably do `hash(map_chr(paths, hash_file))` for combining the hashes of multiple files, as @lionel- mentioned. That would probably be simpler than using the hasher API below.

---

I've also implemented a separate API that could be used for R level streaming hashes, if desired. It is composed of:
- `hasher_init()`: Returns an external pointer to a xxhash hash state object
- `hasher_update(x, data)`: Takes a hasher state, and a raw vector of new data to add to the hash
- `hasher_value(x)`: Converts the current state of the hasher into an actual string hash value

I originally thought we'd need this to implement `hash_file()`, but since we have limited it to file paths rather than connection objects, we can implement `hash_file()` completely at the C level.

I'm happy to ditch the hasher API if we don't feel it has any usage.


R/hash.R
```diff
@@ -1,26 +1,39 @@
...
+#' - For `hash_file()`, a character vector containing one hash per file.
+#'
 #' @export
 #' @examples
 #' hash(c(1, 2, 3))
 #' hash(mtcars)
+#'
+#' authors <- file.path(R.home("doc"), "AUTHORS")
+#' copying <- file.path(R.home("doc"), "COPYING")
+#' hash_file(c(authors, copying))
```
lionel-, line 36: Maybe show `hash(hash_file))`?

src/internal/hash.c
```diff
@@ -246,3 +259,149 @@ void hash_skip(struct hash_state_t* p_state, void* p_input, int n) {
...
+
+#define CHUNK_SIZE 512 * 1024
+
+static
+r_obj* hash_file_impl(void* p_data) {
+  struct exec_data* p_exec_data = (struct exec_data*) p_data;
+  r_obj* path = p_exec_data->x;
+  XXH3_state_t* p_xx_state = p_exec_data->p_xx_state;
+
+  if (!r_is_character(path, -1)) {
```
lionel-, line 284: `r_is_character()` is no longer part of the C lib and mainly remains because it is used to implement `purrr::is_character()`. It should probably be renamed with a different prefix.

We normally use `r_typeof(path) != R_TYPE_character` for checking types.

src/internal/file.h
```diff
@@ -0,0 +1,8 @@
+#ifndef RLANG_INTERNAL_FILE_H
+#define RLANG_INTERNAL_FILE_H
+
+#include <stdio.h>
+
+FILE* r_fopen(r_obj* path, const char* mode);
```
lionel-, line 6: It would be helpful to mention that the caller needs to manage the implicit protection stack (not needed here but may be in future usages).

lionel- (approved):

src/internal/file.h
DavisVaughan: Decided to manage the stack for the user with `vmaxget()` and `vmaxset()`

DavisVaughan:
Keeping the `hasher_*()` API around (internally) for now. Might be useful for prototyping later

PR #1187, opened 2021-05-04
Implement `r_normalise_encoding()`

DavisVaughan:
Pulled over from https://github.com/r-lib/vctrs/blob/master/src/translate.c

I've left in the maybe-referenced copy semantics, as that seems to work best with vctrs.

Otherwise, I've done minor tweaks to use more of the rlang lib, but nothing major.

I've pulled in all the relevant tests from vctrs, with a few testthat utilities. There are quite a few tests, but I went through each one and none of them feel redundant. I think they are required to ensure we hit all the code paths, but we could move them out of `test-c-api.R` if you feel it clutters it too much.

### DIFF
--- a/src/Makevars
+++ b/src/Makevars
@@ -44,6 +44,7 @@ internal-files = \
         internal/call.c \
         internal/dots.c \
         internal/dots-ellipsis.c \
+        internal/encoding.c \
         internal/env.c \
         internal/env-binding.c \
         internal/eval.c \
--- a/src/internal/decl/encoding-decl.h
+++ b/src/internal/decl/encoding-decl.h
@@ -1,0 +1,23 @@
+static
+r_obj* chr_normalise_encoding(r_obj* x);
+
+static inline
+r_ssize chr_find_normalise_start(r_obj* x, r_ssize size);
+
+static
+r_obj* list_normalise_encoding(r_obj* x);
+
+static
+r_obj* r_attrib_normalise_encoding(r_obj* x, r_obj* attrib);
+
+static
+r_obj* attrib_normalise_encoding(r_obj* x);
+
+static inline
+r_obj* str_normalise(r_obj* x);
+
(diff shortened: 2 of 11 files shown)


lionel-:
Thanks!

I added it to the C library under the name `r_obj_fix_encoding()`. Do you like it? If so I'll change the C callable as well.

Merging now so I can depend on this in the unique branch.

DavisVaughan:
I like the `obj_` prefix, but I'd prefer something a bit more specific than `fix`. Normalise was nice because I could document what a "normalised" CHARSXP was (NA, ASCII, or UTF-8 marked)

Based on the naming scheme we used in `arg_match()`, I could see:

- `r_normalise_encoding()` -> `obj_normalise_encoding()`
- `r_attrib_normalise_encoding()` -> `obj_attrib_normalise_encoding()`
- `rlang_normalise_encoding()` -> `rlang_obj_normalise_encoding()`
- `r_obj_fix_encoding()` -> `r_obj_normalise_encoding()`

lionel-:
I try to avoid "normalise" and "standardise" because they are long names and have the "s" vs "z" spelling issue. I think "fixing encoding" could still be documented as producing "an object with standard encoding, meaning that ...".

DavisVaughan:
Alternatively I considered naming this `obj_encode_utf8()` since that is the intention of the function. Fully ASCII strings are left with an `"unknown"` encoding (as seen by `Encoding()`), but they are always valid UTF-8 so that's probably not too confusing

lionel-:
Right, the "unknown" thing is just an implementation detail of UTF-8 strings in R so that shouldn't have an impact on the naming scheme or documentation. I think I like `r_obj_encode_utf8()`.

lionel-:
Initial vctrs PR: https://github.com/r-lib/vctrs/pull/565

PR #1195, opened 2021-05-06
Add bytes class

### LINKED ISSUE #1117: Bytes class
e.g. for memtools

lionel-:
From https://github.com/r-lib/bench/blob/master/R/bytes.R

- The class name is `"rlib_bytes"`.
- Variadic user-facing ctor is called `bytes()` (but see note below).
- Programmatic ctor is `new_bytes()`.
- Conversion function is `as_bytes()`.
- Parsing function is `parse_bytes()`.

There is already an experimental function named `bytes()`. It's only used in vctrs tests so I'll wait for the next version of vctrs to get on CRAN to replace the existing `bytes()` function by a variadic ctor. In the meantime the ctor is called `bytes2()`.

The print method generally follows Jim's practice of no header. However it sometimes includes a header to disambiguate edge cases (empty and unspecified vectors). If that seems reasonable, I think it'd be good to do the same with other classes to prevent confusion.

Closes #1117.


DESCRIPTION
```diff
@@ -23,6 +23,7 @@ Biarch: true
 Depends:
     R (>= 3.3.0)
 Imports:
+    methods,
```
jimhester, line 26: Perhaps we could do the oldclass registration with a onLoad hook, so we wouldn't have to import methods? We would have to verify it works, but I think it should...

On the other hand methods is already a default package, so most people will already have it loaded anyway.

On the other other hand :) just loading methods adds a _lot_ of symbols to the symbol table, which significantly affects the time it takes to perform a gc. e.g.

```
R_DEFAULT_PACKAGES=base,compiler,graphics,utils,grDevices,stats,datasets R --quiet --vanilla -e 'system.time(for (i in 1:100) gc())'
> system.time(for (i in 1:100) gc())
   user  system elapsed
  0.447   0.002   0.449
R_DEFAULT_PACKAGES=base,compiler,graphics,utils,grDevices,stats,datasets,methods R --quiet --vanilla -e 'system.time(for (i in 1:100) gc())'
> system.time(for (i in 1:100) gc())
   user  system elapsed
  1.283   0.003   1.287
```

jimhester (approved):
I think this conditional printing of the header is a good compromise.

But I am not dead set against unconditionally having a header, if you really think it is better to always show it I think that would be fine as well.

Otherwise LGTM!

lionel-:
> But I am not dead set against unconditionally having a header, if you really think it is better to always show it I think that would be fine as well.

I like no header as well when it's clear from the output what type a vector is, which is the case for bytes objects. I wonder if it'd be handy or annoying to have headers for base `NA` vectors?

PR #1198, opened 2021-05-07
Prompt user for outdated downstream deps

lionel-:
User is now prompted to update outdated dependencies when they are loaded in the session:

```r
As of rlang 0.4.11.9000, pkgFoo must be at least version 1.0.0.
x pkgFoo 0.1.0 is too old for rlang 0.4.11.9000.
! Would you like to update the package pkgFoo now?

1: Yes
2: No

Selection: 
```

If they accept, the package is installed with pak if available. If they decline, they are informed about a global option to turn off the checks. This should be useful for users who are not able to update right away:

```r
> Selection: 2
Set `options(rlib_downstream_check = FALSE)` to disable this prompt.
```

In non-interactive sessions, the usual warning is displayed:

```r
Warning message:
As of rlang 0.4.11.9000, pkgFoo must be at least version 1.0.0.
x pkgFoo 0.1.0 is too old for rlang 0.4.11.9000.
i Please update pkgFoo with `install.packages("pkgFoo")` and restart R. 
```

Added ellipsis 0.3.2 as a requirement to avoid import warnings for `check_dots_empty()` etc. This uses the new UI based on the DESCRIPTION syntax (implemented outside this PR):

```r
on_load({
  check_downstream("0.5.0", "ellipsis (>= 0.3.2)")
})
```

### DIFF
--- a/R/compat-downstream-dep.R
+++ b/R/compat-downstream-dep.R
@@ -10,6 +10,13 @@
 
 # Changelog:
 #
+# 2020-05-12:
+#
+# * All packages are now updated at once. The user is not prompted
+#   again after accepting or declining to update the packages, even
+#   when one of the packages is loaded later on.
+#
+#
 # 2020-05-07:
 #
 # * In interactive sessions, user is now prompted to update outdated
@@ -46,15 +53,13 @@ check_downstream <- function(ver, ..., with_rlang = requireNamespace("rlang")) {
     }
   }
 
-  # Must be `lapply()` instead of a `for` loop so that each `dep` is
-  # bound to its own closure env
-  lapply(deps, function(dep) {
-    force(dep)
+  checked <- FALSE
+  for (dep in deps) {
     on_package_load(
       dep[["pkg"]],
-      .rlang_downstream_check(pkg, ver, dep, with_rlang = with_rlang)
+      .rlang_downstream_check(pkg, ver, deps, with_rlang = with_rlang)
     )
-  })
+  }
 }
 
 .rlang_downstream_parse_deps <- function(deps) {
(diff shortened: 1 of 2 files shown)


R/compat-downstream-dep.R
```diff
@@ -82,20 +102,39 @@ check_downstream <- function(ver, ..., with_rlang = requireNamespace("rlang")) {
...
+
+  prompt <- c("!" = sprintf("Would you like to update the package %s now?", dep_pkg))
+  inform(c(header, info_pkg, prompt))
+
+  if (utils::menu(c("Yes", "No")) != 1) {
+    inform("Set `options(rlib_downstream_check = FALSE)` to disable this prompt.")
+    return()
+  }
+
+  if (is_installed("pak")) {
```
hadley, line 134: Is this the same code path as `check_installed()` or different?

R/compat-downstream-dep.R
```diff
@@ -82,20 +102,39 @@ check_downstream <- function(ver, ..., with_rlang = requireNamespace("rlang")) {
...
+  if (!is_interactive()) {
+    warn(c(header, info_pkg, info_install))
+    return()
+  }
+
+  prompt <- c("!" = sprintf("Would you like to update the package %s now?", dep_pkg))
+  inform(c(header, info_pkg, prompt))
+
+  if (utils::menu(c("Yes", "No")) != 1) {
+    inform("Set `options(rlib_downstream_check = FALSE)` to disable this prompt.")
```
hadley, line 130: What will happen if they choose no? Is the package loaded or not?

R/compat-downstream-dep.R
lionel-: This is the same code but not the same code path.

lionel-:
>  is the hook run early enough that you can actually install a new version and it be used instead of the old?

I think the usual caveat about installing loaded packages applies. If user chooses to update the package they will have to reload to be safe. I guess we should make that clear in the prompt.

hadley:
Yeah, I think the message will need work then. I'd suggest working on it a bit and then sharing in the tidyverse channel for further discussion.

lionel-:
We now install all packages at once when multiple checks are supplied.

Following #tidyverse feedback on Slack, we also inform about the consequences of not updating:

```
The packages `ellipsis` (>= 0.3.2) and `vctrs` (>= 0.3.8) are required as of rlang 0.5.0.
! Would you like to update them now?
  You will likely need to restart R if you update now.
i Not updating now is completely safe and will only cause import warnings.

1: Yes
2: No
```

This is supported by a new `info` argument that is treated as an `"i"` bullet by default.

PR #1199, opened 2021-05-10
Add restart for package not found errors

lionel-:
`check_installed()` signals error conditions of class `rlib_error_package_not_found`. The error includes `pkg` and `version` fields. They are vectorised and may include several packages.

The error is signalled with a `rlib_restart_package_not_found` restart on the stack to allow handlers to install the required packages. To do so, add a calling handler for `rlib_error_package_not_found`, install the required packages, and invoke the restart without arguments. This restarts the check from scratch.

The condition is not signalled in non-interactive sessions, in the restarting case, or if the `rlib_restart_package_not_found` user option is set to `FALSE`.

For now the ctor and signal functions for the package-not-found errors are not exported because the error class and signalling protocol are rather intertwined.


gaborcsardi (approved):
Cool! I'll add support for this in pak, and we should probably add links to the docs, both ways.

gaborcsardi:
@lionel- Actually, wait a sec before merging.

gaborcsardi:
I just checked that `requireNamespace()` in `is_installed()` does not throw a `packageNotFoundError` condition currently, only `loadNamespace()` does. So that is good, because pak will also catch `packageNotFoundError`. 

However R-core is considering throwing this condition from `find.package()`, in which case `is_installed()` will also throw `packageNotFoundError`, possibly, depending on how they implement this. Hopefully they'll do it in a way that we can still use `requireNamespace()` to check for installed packages.

I _think_ we don't need to do anything special about this, just wanted to note this here. 

So I think this PR is good to merge. :)

PR #1213, opened 2021-05-20
Add message formatting to `compat-cli.R`

lionel-:
Branched from #1212.

* Add compats for `format_error()`, `format_warning()`, and `format_message()` to the compat file. These call the corresponding cli functions if available. Otherwise, the fallback bullets formatting is used.

* Rename `style_emph()` etc to `format_emph()`. These stylers format eagerly and are compatible with `cat()`.

* Add new `style_emph()` etc API. Unlike the `format_` variants, these stylers create `{.span input}` markup when cli is available. They must be paired with a function like `format_error()` that will process the markup.


lionel-:
hmm maybe the `style_` functions should be renamed to `mark_` to make it clearer that they create markup. Then we can use the `style_` prefix more consistently with the cli API, e.g. `style_bold()` instead of `ansi_bold()`.

PR #1216, opened 2021-05-21
WIP: Overwrite deprecated with lifecycle::deprecated

krlmlr:
as soon as it becomes available.


lionel-:
Thanks! Locking and unlocking should only be used in development tools. Also we probably want a more general approach, maybe we can fix this in pkgload.

PR #1217, opened 2021-05-21
compat-purrr.R: fix typo in `imap()`

mgirlich:
use `.x` instead of `x`

### DIFF
--- a/R/compat-purrr.R
+++ b/R/compat-purrr.R
@@ -62,7 +62,7 @@ map2_chr <- function(.x, .y, .f, ...) {
   as.vector(map2(.x, .y, .f, ...), "character")
 }
 imap <- function(.x, .f, ...) {
-  map2(.x, names(x) %||% seq_along(x), .f, ...)
+  map2(.x, names(.x) %||% seq_along(.x), .f, ...)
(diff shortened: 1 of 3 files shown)


lionel-:
Thanks! Can you please add a unit test and add a mention in the changelog at the top of the file (please tag yourself in the bullet).

mgirlich:
Interestingly, there was already a definition of `imap()` with swapped order of `.x` and names. But it seems they weren't used anywhere in the code.
By the way I have lots of test failures that seem to be related to changes in `abort()` and friends, e.g.

```
Failure (test-dots-ellipsis.R:53:3): error if if dots not empty
Snapshot of code has changed:
old[3:9] vs new[3:10]
  "Output"
  "  <error/rlib_error_dots_nonempty>"
  "  `...` is not empty."
- "  i These dots only exist to allow future extensions and should be empty."
+ "  i These dots only exist to allow future extensions and should be"
+ "    empty."
  "  x We detected these problematic arguments:"
  "  * `xy`"
  "  i Did you misspecify an argument?"
```

lionel-:
Thanks!

PR #1235, opened 2021-07-06
Handle `call` fields in errors

lionel-:
To support https://github.com/tidymodels/recipes/pull/734 (cc @DavisVaughan).

1. Display the whole message ourselves including the `Error: ` prefix. This removes the limitation on the length of error messages (the limit can easily be reached because of ANSI markup by cli). This also makes it possible to format calls according to our standards.

2. Include `call` fields in error prefixes if present. Complex calls containing inlined objects are ignored and the arguments of the call are stripped to keep it simple. The call is formatted as code using cli if available. Unlike `stop(call. = TRUE)`, there is no space after `:`.

```r
abort("msg", call = quote(foo(bar, baz)))
#> Error in `foo()`: msg
```

Rethrowing with a context added in `call` field:

```r
f <- function() g()
g <- function() h()
h <- function() abort("msg")

a <- function() b()
b <- function() c()
c <- function() {
  withCallingHandlers(f(), error = function(c) {
    c$call <- quote(foo())
    cnd_signal(c)
  })
}

a()
#> Error in `foo()`: msg
#> Run `rlang::last_error()` to see where the error occurred.

last_error()
#> <error/rlang_error>
#> msg
#> Context: `foo()`          <---
#> Backtrace:
#>  1. global::a()
#>  2. global::b()
#>  3. global::c()
#>  5. global::f()
#>  6. global::g()
#>  7. global::h()
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: rlang
-Version: 0.4.11.9000
+Version: 0.4.11.9001
 Title: Functions for Base Types and Core R and 'Tidyverse' Features
 Description: A toolbox for working with base types, core R features
   like the condition system, and core 'Tidyverse' features like tidy
--- a/NEWS.md
+++ b/NEWS.md
@@ -29,6 +29,12 @@
 
 ## Features and bugfixes
 
+* `abort()` now displays `call` field in error messages. The call is
+  only displayed if it is a simple expression (e.g. no inlined
+  function) and the arguments are not displayed to avoid distracting
+  from the error message. The message is formatted with the tidyverse
+  style (`code` formatting by the cli package if available).
+
 * The `.ignore_empty` argument of `enexprs()` and `enquos()` no longer
   treats named arguments supplied through `...` as empty, consistently
   with `exprs()` and `quos()` (#1229).
--- a/R/cnd-abort.R
+++ b/R/cnd-abort.R
@@ -203,22 +203,20 @@ signal_abort <- function(cnd) {
   # Save the unhandled error for `rlang::last_error()`.
   last_error_env$cnd <- cnd
 
-  if (!is_interactive()) {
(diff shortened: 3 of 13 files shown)


gaborcsardi (approved):
Looks good!

hadley (approved):
Looks great. Does this also make it easier to override the default red colour in RStudio?

And I assume there will be a future change to `abort()` to make the call easier to capture?

lionel-:
> Does this also make it easier to override the default red colour in RStudio?

Yes we could use the same approach as in `inform()` to stream to stdout in interactive sessions. Is this something we want to do?

> And I assume there will be a future change to abort() to make the call easier to capture?

yup then it's just a matter of capturing the grand-parent call by default. But how does this interact with displaying the tidymodels (and tidyverse?) step/context in the error message? When both the call site and the larger context are available, which one should have priority? Or do we want to somehow display both? In that case we'd need another condition field such as `ctxt`.

hadley:
Yeah, we should copy `inform()` here, maybe just making "Error" bold by default (but we'll need to discuss the default styling more broadly). 

Not sure how we'll balance the different context needs. Not directly related, but there's some interesting in https://blog.rust-lang.org/inside-rust/2021/07/01/What-the-error-handling-project-group-is-working-towards.html.

lionel-:
> Yeah, we should copy inform() here, maybe just making "Error" bold by default (but we'll need to discuss the default styling more broadly).

Done in https://github.com/r-lib/rlang/commit/84f46a31


> Not sure how we'll balance the different context needs. Not directly related, but there's some interesting in https://blog.rust-lang.org/inside-rust/2021/07/01/What-the-error-handling-project-group-is-working-towards.html.

Good point, a better display of chained errors might be key. We have access to the `call` fields of all errors in a chain.

PR #1241, opened 2021-07-15
better arg_match description

topepo:
While the docs say

> This is equivalent to `base::match.arg()`

it is not obvious that `rlang::arg_match()` should only be used when the argument takes character values.


lionel-:
Thanks!

PR #1242, opened 2021-07-19
Inferring erroring call from the stack

lionel-:
#### Call inference

- `abort()` automatically picks up the `call` argument from the immediate caller if not supplied. If the global option `"rlang:::restrict_default_error_call"` is set to `TRUE`, only calls for exported functions are picked up.

- If `call` is an environment, the behaviour depends on the presence and value of a `.__error_call__.` binding.
    - If there is no flag, the relevant `sys.call()` for that environment is taken.
    - If the flag is `NULL` or a call, this is taken as call.
    - If the flag is an environment (or equivalently the string `"caller"` for performance-critical code), the logic recurses.

   Setting the flag is useful to avoid passing a call argument through many layers in input checking functions, i.e. in vctrs.

#### Helpers

- `error_arg()` is a variant of `substitute()` that returns a string. It should be used as default `arg` (or `error_arg`) argument in input checkers. The result can be included in error messages and formatted as argument using cli or compat-cli. Here is the signature of `arg_match()`, a function that takes both `arg` and `call` arguments:

    ```r
    arg_match <- function(arg,
                          values = NULL,
                          ...,
                          error_arg = caller_arg(arg),
                          error_call = caller_env())
    ```

- `format_error_call()` is a helper for generating the "in" part of error messages, i.e. in testthat for displaying error messages. This is also used by `abort()`. It uses cli if available.

- The `args_error_context` documentation topic can be inherited from to document `arg` and `call` arguments consistently.

#### Naming proposal

Use `arg` and `call` if possible. Otherwise disambiguate with a prefix (e.g. `x_arg` and `y_arg` in `vec_ptype2()`). Or disambiguate with the `error_` namespace.

#### Ports

- `arg_match()` gains `error_call` and `error_arg` arguments.

- Ellipsis helpers gain `call` arguments.

- `dots_list()` forwards the error call when checkinq argument constraints such as homonyms.


tests/testthat/_snaps/arg.md
```diff
@@ -1,88 +1,129 @@
...
+    i Did you mean "bar"?
+
+# gives error with different than rearranged arg vs value
+
+    `myarg` must be one of "fun" or "bar", not "foo".
+
+---
+
+    `x` must be one of "foo" or "bar", not "foo".
+    i Did you mean "foo"?
```
hadley, line 17: This is a bit confusing. Is foo allowed or not? (I assume I'm probably reading this out of context)

R/arg.R
```diff
@@ -137,33 +160,20 @@ arg_match_invalid_msg <- function(arg, values, arg_nm) {
 #' # Succeeds
 #' f(NULL)
 #' @export
-arg_require <- function(arg) {
+arg_require <- function(arg,
+                        error_arg = caller_arg(arg),
```
hadley, line 164: Per your naming scheme, should this be `arg_arg` and `arg_call`?

R/cnd-abort.R
```diff
@@ -170,7 +168,7 @@
 abort <- function(message = NULL,
                   class = NULL,
                   ...,
-                  call = NULL,
+                  call = missing_arg(),
```
hadley, line 171: Why the explicit `missing_arg()`?

R/cnd-abort.R
```diff
@@ -248,6 +290,325 @@ signal_abort <- function(cnd, file = NULL) {
...
+#' - `call` parameters should be included in error conditions in a
+#'   field named `call`. An easy way to do this is by passing a `call`
+#'   argument to [abort()]. See also [local_error_call()].
+#'
+#' @param arg,error_arg An argument name as a string. This argument
+#'   will be mentioned in error messages as the input that is at the
+#'   origin of a problem.
+#' @param call,error_call An expression (as returned by e.g.
+#'   `sys.call()`) representing the context in which the error
+#'   occurred. If non-null, the call is stripped from its arguments to
```
hadley, line 468:
```suggestion
#'   occurred. If non-null, the call is stripped of its arguments to
```

R/cnd-abort.R
```diff
@@ -248,6 +290,325 @@ signal_abort <- function(cnd, file = NULL) {
...
+NULL
+
+#' Find the caller argument for error messages
+#'
+#' @description
+#'
+#' `caller_arg()` is a variant of `substitute()` or [ensym()] that
+#' returns an argument name as a string.
+#'
+#' - The resulting name should be formatted as argument, e.g. using
```
hadley, line 487: I don't quite get what you mean by "should" here.

R/cnd-abort.R
```diff
@@ -248,6 +290,325 @@ signal_abort <- function(cnd, file = NULL) {
...
+#'
+#' @name args_error_context
+NULL
+
+#' Find the caller argument for error messages
+#'
+#' @description
+#'
+#' `caller_arg()` is a variant of `substitute()` or [ensym()] that
+#' returns an argument name as a string.
```
hadley, line 485: I think you need a sentence explaining how it's different from `substitute()`.

R/cnd-abort.R
```diff
@@ -248,6 +290,325 @@ signal_abort <- function(cnd, file = NULL) {
...
+#' @param call,error_call An expression (as returned by e.g.
+#'   `sys.call()`) representing the context in which the error
+#'   occurred. If non-null, the call is stripped from its arguments to
+#'   keep it simple.
+#'
+#'   Can also be an execution environment of a currently running
+#'   function (as returned by e.g. `parent.frame()`). The
+#'   corresponding call is then retrieved.
+#'
+#'   See also [rlang::local_error_call()].
```
hadley, line 549: Maybe it's worth documenting these together? I think it might be useful to have a couple of examples here to illustrate the basic ideas of call look up.

R/cnd-abort.R
```diff
@@ -248,6 +290,325 @@ signal_abort <- function(cnd, file = NULL) {
...
+    break
+  }
+
+  # This happens when a function using its caller's call is called in
+  # testthat snapshots or within `local()`.
+  if (is_call(call, c("eval", "evalq"))) {
+    return(NULL)
+  }
+
+  if (is_call(call) && is_expression(call)) {
```
hadley, line 584: What does being a call and an expression mean?

R/cnd-abort.R
```diff
@@ -554,3 +915,26 @@ last_trace <- function() {
...
 last_error_env <- new.env(parent = emptyenv())
 last_error_env$cnd <- NULL
+
+
+#' Development notes - `cnd-abort.R`
+#'
+#' @section Interaction between error calls and NSE:
+#'
+#' Functions that forward their error context to their caller
+#' shouldn't generally be called via NSE but there are expections,
```
hadley, line 925:
```suggestion
#' shouldn't generally be called via NSE but there are exceptions,
```

R/cnd-abort.R
```diff
@@ -554,3 +915,26 @@ last_trace <- function() {
 # This is where we save errors for `last_error()`
 last_error_env <- new.env(parent = emptyenv())
 last_error_env$cnd <- NULL
+
+
+#' Development notes - `cnd-abort.R`
```
hadley, line 920: I think this might be a bit too distant to the code it affects and it'd be easy to miss when updating the code. Maybe include in the docs for `error_call()`, but use `#` instead of `#'` so it doesn't become user facing?  Or maybe just inline as comments in the appropriate bit of the code?

tests/testthat/_snaps/arg.md
```diff
@@ -1,88 +1,129 @@
...
 
     Code
-      (expect_error(arg_match0("continuuos", c("discrete", "continuous"), "my_arg")))
+      (expect_error(arg_match_wrapper("continuuos", c("discrete", "continuous"),
+      "my_arg")))
     Output
       <error/rlang_error>
       `my_arg` must be one of "discrete" or "continuous", not "continuuos".
       i Did you mean "continuous"?
+      Context: `arg_match0_wrapper()`
```
hadley, line 28: What's the deal with `Context:`?

tests/testthat/helper-rlang.R
```diff
@@ -90,3 +90,10 @@ skip_if_not_windows <- function() {
   system <- tolower(Sys.info()[["sysname"]])
   skip_if_not(is_string(system, "windows"), "Not on Windows")
 }
+
+arg_match_wrapper <- function(arg, ...) {
+  arg_match(arg, ...)
+}
+arg_match0_wrapper <- function(arg, values, arg_nm = "arg", ...) {
+  arg_match0(arg, values, arg_nm = arg_nm, ...)
```
hadley, line 98: Why do we need these wrappers?

hadley (commented):
First set of questions 😄

tests/testthat/_snaps/arg.md
lionel-: This is now:

```r
#> Error in `arg_match()`: `arg` must be length 1 or a permutation of `values`.
```

I haven't taken the time to include an enumeration of values because this is a rare error message. Normally `arg_match()` requires a length 1 vector or a length n vector in case the user hasn't supplied a value and `arg_match()` gets the argument default. In some cases, that default vector is not identical to the vector of values because it is generated in another package. A permutation of the values is then allowed. This unit test shows what happens when a length > 1 `arg` is not a permutation of `values`.

R/arg.R
lionel-: I would prefix with the argument name only when there are two arguments to disambiguate as in `vec_ptype2()`. Normally I'd simply use `arg` and `call`. Otherwise I prefix with `error_` when there are existing arguments with these names as is the case for `arg_require()` and `arg_match()`. I believe this naming scheme should be flexible and intuitive enough for most cases.

R/cnd-abort.R
lionel-: I think I found it weird to not have a default for an argument on the RHS of `...`. I don't have any strong opinion about this though.

R/cnd-abort.R
lionel-: Is this clearer?

```
#' - When included in an error message, the resulting name should
#'   generally be formatted as argument, for instance using the `.arg`
#'   in the cli package.
```

R/cnd-abort.R
lionel-: How about:

```r
#' `caller_arg()` is a variant of `substitute()` or [ensym()] for
#' arguments that reference other arguments. Unlike `substitute()`
#' which returns any expression, the referenced argument should be a
#' symbol. That symbol is then returned as a string ready to be
#' included in error messages.
```

R/cnd-abort.R
lionel-: The narrow definition of "expression" in rlang is any R object that can be generated by the parser. Being a call and an expression in the narrow sense means that the call does not contain any complex object such as a function definition or inlined vector of length > 1. This check is meant to avoid displaying complex calls created by NSE.

lionel-: I guess to be completely correct we should also check that the CAR is not a call though.

tests/testthat/_snaps/arg.md
lionel-: This shows the `call` condition field. Would you prefer

```
Call: `arg_match0_wrapper()`
```

tests/testthat/helper-rlang.R
lionel-: This is just to test that the wrapper call appears in error messages.

R/cnd-abort.R
lionel-: I've just tried this, it doesn't work very well because `local_error_call()` and `error_call()`/`format_error_call()` have different audiences. The former is for end users of `abort()` whereas the latter two are for environments that display error messages from error conditions, such as testthat. So instead I added another link to `local_error_call()` in the description.

R/cnd-abort.R
lionel-: The check is now simply:

```r
is_call(call) && is_symbol(call[[1]])
```

R/arg.R
hadley: Ok, makes sense

R/cnd-abort.R
hadley: I think it looks slightly better to me to leave it as `call`, but I don't feel particularly strongly about it.

tests/testthat/_snaps/arg.md
hadley: Yeah, call would be better. But why is it displayed here at all?

R/cnd-abort.R
```diff
@@ -248,6 +290,345 @@ signal_abort <- function(cnd, file = NULL) {
...
+#' structured into many different functions. In this case, use
+#' `local_error_call()` to set the call locally or instruct `abort()`
+#' to climb the call stack one level to find the relevant call. In the
+#' following example, the complexity is not so important that sparing
+#' the argument passing makes a big difference. However this
+#' illustrates the pattern:
+#'
+#' ```
+#' arg_check <- function(arg,
+#'                       error_arg = as_string(substitute(arg)),
```
hadley, line 382: Should this use `caller_arg()`?

R/cnd-abort.R
```diff
@@ -248,6 +290,345 @@ signal_abort <- function(cnd, file = NULL) {
...
+#' @name args_error_context
+NULL
+
+#' Find the caller argument for error messages
+#'
+#' @description
+#'
+#' `caller_arg()` is a variant of `substitute()` or [ensym()] for
+#' arguments that reference other arguments. Unlike `substitute()`
+#' which returns any expression, the referenced argument should be a
```
hadley, line 486:
```suggestion
#' which returns any expression, the referenced argument must be a
```

tests/testthat/_snaps/arg.md
lionel-: For end users calling `last_error()` we display it for the same reason we display the "in" part in error messages, because it is informative. In unit tests, it's useful to snapshot the "in" part of the messages.

R/cnd-abort.R
```diff
@@ -248,6 +290,345 @@ signal_abort <- function(cnd, file = NULL) {
...
+#'   [caller_arg()].
+#'
+#' - `call` parameters should be included in error conditions in a
+#'   field named `call`. An easy way to do this is by passing a `call`
+#'   argument to [abort()]. See also [local_error_call()].
+#'
+#' @param arg,error_arg An argument name as a string. This argument
+#'   will be mentioned in error messages as the input that is at the
+#'   origin of a problem.
+#' @param call,error_call An expression (as returned by e.g.
```
hadley, line 466: Should it usually be an environment? In most checker functions it'll be `caller_env()`, right? And then supplying a call is just a special case?

R/cnd-abort.R
hadley: And mention that it errors if not?

tests/testthat/_snaps/arg.md
hadley: Maybe it'd feel more natural with slightly different formatting?

```R
<error/rlang_error>
`my_arg` must be one of "discrete" or "continuous", not "continuuos".
i Did you mean "continuous"?

$call: arg_match0_wrapper()
```

That would make it more obvious that you're displaying a component of the error object, and would be straightforward to generalise in the future.

lionel-: Then should it be `$trace:` as well? The current formatting is meant to be consistent with `Backtrace:`.

R/cnd-abort.R
lionel-: Definitely, done.

R/cnd-abort.R
lionel-: I removed the explicit default.

hadley (approved):
BTW I think this is fine to merge as is, assuming that we'll keep iterating on it for a bit with smaller PRs.

PR #1249, opened 2021-08-18
Allow `names_inform_repair()` to be silenced by a global option

DavisVaughan:
This also adds the `"rlib_message_name_repair"` class to the message signaled by `names_inform_repair()`, giving us yet another method to handle it.

I did not add a NEWS bullet because there isn't one for `names_inform_repair()` being a new function.

_Side note:_ Should this function be named `inform_name_repair()`? We normally use `stop_*()` and `warn_*()` elsewhere for our error/warning naming conventions, and that form matches the global option / class name better.

### DIFF
--- a/R/names.R
+++ b/R/names.R
@@ -3,7 +3,24 @@ names_as_unique <- function(names, ..., quiet = FALSE) {
   .Call(ffi_names_as_unique, names, quiet)
 }
 
-#' Inform about names repair
+#' Inform about name repair
+#'
+#' @section Muffling and silencing messages:
+#'
+#' Name repair messages are signaled with [inform()] and are given the class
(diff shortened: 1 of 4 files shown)


lionel-:
> Should this function be named inform_name_repair()? We normally use stop_*() and warn_*() elsewhere for our error/warning naming conventions, and that form matches the global option / class name better.

Hard to say. The idea here is to have a `names_` namespace for completion. It does create an inconsistency with the `stop_`, `warn_` and `inform_` namespaces. What do you think @jennybc @hadley?

lionel- (approved):
Looks good!

hadley:
I don't have any strong sense of which prefix should win here. I think you could argue it in either direction.

PR #1253, opened 2021-08-23
Use error style guide in `env_get()` error for missing binding

lionel-:
Previously:

```r
env_get(nm = "foobar")
#> Error in env_get(nm = "foobar") : 
#>   argument "default" is missing, with no default
```

Now:

```r
env_get(nm = "foobar")
#> Error: Can't find `foobar` in environment.
```


hadley (approved):
I could see an argument for making it `` Can't find %s in `env`. ``, but I guess this error message is usually seen downstream and referring to the argument name might not be that useful?

lionel-:
@hadley That was my thought as well.

PR #1264, opened 2021-08-31
Display source locations in error messages

### LINKED ISSUE #1260: Display srcref stored in `$call` fields

lionel-:
- Add a line break between the error prefix (the "Error:" part) and the error message when the former gets too long. A long prefix is currently defined as longer than half the screen width or when a source location is included.

- Add srcref information in the error prefix (closes #1260). This is not displayed when testthat is running to avoid making snapshots brittle. testthat provides source locations for unexpected errors and warnings via other means.

- Add an "Error prefix:" section in `?abort`.

- Include srcrefs on their own lines in simplified backtraces.

- Use "at" to introduce srcrefs in backtraces, consistently with the error prefix.

```r
rlang:::eval_parse("{
    f <- function() g()
    g <- function() h()
    h <- function() abort('Foo.')
  }")

src_file <- h %@% srcref %@% srcfile
src_file$filename <- "/foo/bar/baz/myfile.R"

f()
#> Error in `h()` at bar/baz/myfile.R:3:9:
#> Foo.
#> Run `rlang::last_error()` to see where the error occurred.

last_error()
#> <error/rlang_error>
#> Error in `h()` at bar/baz/myfile.R:3:9:
#> Foo.
#> Backtrace:
#>  1. global::f()
#>  2. global::g()
#>     at bar/baz/myfile.R:2:9
#>  3. global::h()
#>     bar/baz/myfile.R:3:9
```

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -386,9 +386,5 @@ df_print <- function(x, ...) {
 }
 
 is_testing <- function() {
-  if (is_bool(opt <- peek_option("rlang:::is_testing"))) {
-    return(opt)
-  }
-
-  is_installed("testthat") && testthat::is_testing()
+  identical(Sys.getenv("TESTTHAT"), "true")
 }
--- a/R/utils.R
+++ b/R/utils.R
@@ -384,3 +384,11 @@ df_print <- function(x, ...) {
   print(x, ...)
   invisible(x)
 }
+
+is_testing <- function() {
+  if (is_bool(opt <- peek_option("rlang:::is_testing"))) {
+    return(opt)
+  }
+
+  is_installed("testthat") && testthat::is_testing()
+}
--- a/tests/testthat/_snaps/trace.md
+++ b/tests/testthat/_snaps/trace.md
@@ -4,10 +4,10 @@
       print(trace, dir = dir)
     Output
           x
-       1. \-rlang:::i() test-trace.R:25:2
-       2.   \-rlang:::j(i) test-trace.R:18:7
-       3.     \-rlang:::k(i) test-trace.R:19:21
-       4.       \-rlang:::l(i) test-trace.R:22:4
+       1. \-rlang:::i() at test-trace.R:25:2
+       2.   \-rlang:::j(i) at test-trace.R:18:7
+       3.     \-rlang:::k(i) at test-trace.R:19:21
+       4.       \-rlang:::l(i) at test-trace.R:22:4
     Code
       cat("\n")
     Output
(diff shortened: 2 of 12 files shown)


tests/testthat/_snaps/trace.md
```diff
@@ -48,19 +48,20 @@
...
-       16.         +-[ base::tryCatch(...) ] with 1 more call test-trace.R:52:20
+       16.         +-[ base::tryCatch(...) ] with 1 more call at test-trace.R:52:20
        18.         \-rlang::trace_back(e, bottom = 0)
     Code
       # Branch
       print(trace, simplify = "branch", dir = dir, srcrefs = srcrefs)
     Output
         1. rlang:::f()
-        2. rlang:::g() test-trace.R:49:20
+        2. rlang:::g()
```
hadley, line 63: Why is this spread over two lines?

R/utils.R
```diff
@@ -384,3 +384,11 @@ df_print <- function(x, ...) {
   print(x, ...)
   invisible(x)
 }
+
+is_testing <- function() {
+  if (is_bool(opt <- peek_option("rlang:::is_testing"))) {
+    return(opt)
+  }
+
+  is_installed("testthat") && testthat::is_testing()
```
hadley, line 393: We normally just inline `testthat::is_testing()`; and then use `local_envvar()` to override in tests.

hadley (approved):
Once all this work is done we should discuss in the tidyverse weekly meeting to get everyone's eyes on it one last time.

tests/testthat/_snaps/trace.md
lionel-: Because calls and srcrefs tend to be long, this way it's easier to read. The srcref line is also dimmed (styled in grey).

R/utils.R
lionel-: Do you mean instead of the peek-option? The is-installed is needed right?

hadley: Yeah, we normally just do:

```R
is_testing <- function() {
    identical(Sys.getenv("TESTTHAT"), "true")
}
```

and then override the `TESTTHAT` var.  Or are you saying that doesn't work because it breaks testthat itself?

tests/testthat/_snaps/trace.md
hadley: Ok. I just asked because it looks different to the full trace above.

R/utils.R
lionel-: oh ok, got it! I'll try that.

PR #1266, opened 2021-08-31
Print parent errors at throw time

lionel-:
And indent error messages when printing a chain of errors.

The display is unchanged in the normal case:

```r
f <- function() g()
g <- function() h()
h <- function() rlang::abort("Low-level.")

f()
#> Error in `h()`: Low-level.

last_error()
#> <error/rlang_error>
#> Error in `h()`: Low-level.
#> Backtrace:
#>  1. global::f()
#>  2. global::g()
#>  3. global::h()
```

If the error is part of a chain, all parent causes are displayed with indentation:

```r
a <- function() b()
b <- function() c()
c <- function() {
  withCallingHandlers(
    error = function(cnd) {
      rlang::abort(
        "High-level",
        parent = cnd,
        call = env_parent()
      )
    },
    f()
  )
}

a()
#> Error in `c()`:
#>   High-level
#> Caused by error in `h()`:
#>   Low-level.

last_error()
#> <error/rlang_error>
#> Error in `c()`:
#>   High-level
#> Caused by error in `h()`:
#>   Low-level.
#> Backtrace:
#>  1. global::a()
#>  2. global::b()
#>  3. global::c()
#>  5. global::f()
#>  6. global::g()
#>  7. global::h()
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -73,6 +73,10 @@
     break is introduced to print the error message on the line below.
     If a source location is displayed in the prefix, a line break is
     always included.
+    
+* The display of chained errors created with the `parent` argument of
+  `abort()` has been improved. Chains of errors are now displayed at
+  throw time with the error prefix "Caused by error:".
 
 * The `.ignore_empty` argument of `enexprs()` and `enquos()` no longer
   treats named arguments supplied through `...` as empty, consistently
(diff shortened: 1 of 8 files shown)


hadley (approved):
Looks good — I think the only thing I'd add is an example to `abort()` that shows how to rethrow errors. It took me a second to understand why `call = env_parent()` was needed.  (Or maybe that should be the default when `parent` is supplied?)

lionel-:
I don't think we can make `env_parent()` the default without causing weird behaviour in various cases. I'll add an example in the doc though. We probably need a whole section about rethrowing and error chains.

Issue #956, opened 2020-03-28
Abbreviating long expressions involving operators


krlmlr:
``` r
rlang::as_label(quote(
  very_long_expression[with(subsetting), -1] -
    another_very_long_expression[with(subsetting), -1]
))
#> [1] "-..."
```

<sup>Created on 2020-03-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #1277, opened 2021-09-14
Improve infix operator labelling

lionel-:
Closest #956.

Also exports `is_call_infix()`. I didn't end up using it though, because we're technically after the print type rather than the parse type here.

cc @hadley


hadley:
Can you explain why? i.e. what function it makes a difference for?

lionel-:
It doesn't make a big difference, only in artificial cases:

```r
is_call_infix(quote(`+`(1, 2)))
#> [1] TRUE
is_call_infix(quote(`+`(1, 2, 3)))
#> [1] TRUE

call_print_fine_type(quote(`+`(1, 2)))
#> [1] "infix"
call_print_fine_type(quote(`+`(1, 2, 3)))
#> [1] "call"
```

Still doesn't feel right though. Binary operators with multiple arguments are not really "infix". Maybe I should just unexport `is_call_infix()`.

lionel-:
Or change it to:

```r
is_call_infix(quote(`+`(1)))
#> [1] FALSE
is_call_infix(quote(`+`(1, 2)))
#> [1] TRUE
is_call_infix(quote(`+`(1, 2, 3)))
#> [1] FALSE    # <- Used to be TRUE
```

hadley:
Yeah, I like the last option — I think an infix operator must have two arguments. I'm also not 100% sure about the inclusion of special operators.

PR #1278, opened 2021-09-14
Zap inline objects from backtraces

### LINKED ISSUE #1069: Trim literal objects from backtraces
To avoid heavy serialisation https://github.com/r-lib/testthat/issues/1223#issuecomment-723650610

lionel-:
Closes #1069.
Closes r-lib/testthat#1223.

Inline objects are replaced by their pillar type sum:

```r
call <- expr(foo(!!(1:2)))
call_zap_inline(call)
#> foo(`<int>`)
```

### DIFF
--- a/R/call.R
+++ b/R/call.R
@@ -1050,3 +1050,24 @@ call_type <- function(x) {
     abort("corrupt language object")
   }
 }
+
+call_zap_inline <- function(x) {
+  if (is_call(x)) {
+    if (is_call(x, "function")) {
+      if (!is_null(x[[2]])) {
+        x[[2]] <- as.pairlist(map(x[[2]], call_zap_inline))
+      }
+      x[[3]] <- call_zap_inline(x[[3]])
+    } else {
+      x[] <- map(x, call_zap_inline)
+    }
+    return(x)
+  }
+
+  if (is_symbol(x) || is_syntactic_literal(x)) {
+    x
+  } else {
+    name <- sprintf("<%s>", rlang_type_sum(x))
+    sym(name)
+  }
+}
--- a/tests/testthat/_snaps/cnd-entrace.md
+++ b/tests/testthat/_snaps/cnd-entrace.md
@@ -40,7 +40,7 @@
        12.     \-rlang:::c()
        13.       +-base::withCallingHandlers(...)
        14.       +-rlang::with_abort(f())
-       15.       | \-base::withCallingHandlers(...)
+       15.       | \-base::withCallingHandlers(expr, error = `<fn>`)
        16.       \-rlang:::f()
        17.         \-rlang:::g()
        18.           \-rlang:::h()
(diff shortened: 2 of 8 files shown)


gaborcsardi (approved):

R/call.R
```diff
@@ -1050,3 +1050,24 @@ call_type <- function(x) {
...
+      }
+      x[[3]] <- call_zap_inline(x[[3]])
+    } else {
+      x[] <- map(x, call_zap_inline)
+    }
+    return(x)
+  }
+
+  if (is_symbol(x) || is_syntactic_literal(x)) {
+    x
```
hadley, line 1068: I find this slightly easier to understand, but it's not that important

```suggestion
    x
  } else if (is_symbol(x) || is_syntactic_literal(x)) {
    x
```

tests/testthat/test-trace.R
```diff
@@ -659,3 +659,18 @@ test_that("can bind backtraces", {
...
+  skip_if_not_installed("base", "3.5.0")
+
+  local_options(
+    rlang_trace_format_srcrefs = FALSE
+  )
+  e <- environment()
+  f <- function(...) g()
+  g <- function() h()
+  h <- function() trace_back(e)
+  trace <- inject(f(!!list()))
```
hadley, line 674: Could also use `do.call()` here, since that's the most common way that users will create inlined objects.

tests/testthat/_snaps/cnd-entrace.md
```diff
@@ -40,7 +40,7 @@
        12.     \-rlang:::c()
        13.       +-base::withCallingHandlers(...)
        14.       +-rlang::with_abort(f())
-       15.       | \-base::withCallingHandlers(...)
+       15.       | \-base::withCallingHandlers(expr, error = `<fn>`)
```
hadley, line 43: Again, not very important, but why does this change?

tests/testthat/test-trace.R
hadley: Maybe something like along these lines:

``` r
e <- environment()
f <- function() do.call("g", list(1:1e6 + 1))
g <- function(...) h()
h <- function() rlang::trace_back(e)
lobstr::obj_size(f())
#> 8,014,888 B
```

<sup>Created on 2021-09-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

(But using something other than `obj_size()`)

hadley: Maybe 

```R
trace <- f()
trace$calls[[3]]
```

?

hadley (approved):

R/call.R
lionel-: The idea here is to keep the terminating and recursive branches separate.

tests/testthat/_snaps/cnd-entrace.md
lionel-: The injected handler is now a symbol. This allows `as_label()` to keep the whole call instead of eliding.

tests/testthat/test-trace.R
lionel-: Good idea, I've added a `do.call()` and size check in the test.

PR #1279, opened 2021-09-15
Don't use namespace operators invalidly

### LINKED ISSUE #902: misleading qualified name when functions defined inside closures
``` r
options(error = rlang::entrace)
as.Date("")
#> Error in charToDate(x) : 
#>   character string is not in a standard unambiguous format
#> Backtrace:
#>  1. base::as.Date("")
#>  2. base::as.Date.character("")
#>  3. base:::charToDate(x)
```

<sup>Created on 2020-01-27 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

The issue is the third item in the backtrace, it makes it seem like you can access it with `base:::charToDate()` but `charToDate()` is actually an internal function is actually defined inside the `as.Date.character()` function. I think maybe in this case it would be better to just leave off the `base:::`?

lionel-:
@jimhester makes a good point that it's confusing to display namespaced calls in the backtrace that don't work in the console. This PR leverages the new metadata recorded in backtraces to replace namespace operators with a simple space. Closes #902.

Local functions are no longer qualified with `:::`

```r
as.Date("foo")
#> Error in charToDate(x) :
#>   character string is not in a standard unambiguous format

# Before
last_error()
#> <error/rlang_error>
#> Error: character string is not in a standard unambiguous format
#> Backtrace:
#>  1. base::as.Date(...)
#>  2. base::as.Date.character(...)
#>  3. base:::charToDate(...)

# After
last_error()
#> <error/rlang_error>
#> Error: character string is not in a standard unambiguous format
#> Backtrace:
#>  1. base::as.Date(...)
#>  2. base::as.Date.character(...)
#>  3. base charToDate(...)
```

Global functions are no longer qualified with `::`

```r
f <- function() g()
g <- function() h()
h <- function() abort("foo")
f()
#> Error in `h()`: foo

# Before
last_error()
#> <error/rlang_error>
#> Error in `h()`: foo
#> Backtrace:
#>  1. global::f()
#>  2. global::g()
#>  3. global::h()

# After
last_error()
#> <error/rlang_error>
#> Error in `h()`: foo
#> Backtrace:
#>  1. global f()
#>  2. global g()
#>  3. global h()
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -34,6 +34,11 @@
 
 ## Features and bugfixes
 
+* Functions defined in the global environments or in local execution
+  environments are now displayed with a space separator in backtraces
+  instead of `::` and `:::`. This avoids making it seem like these
+  frame calls are valid R code ready to be typed in (#902).
+
 * Backtraces no longer contain inlined objects to avoid performance
   issues in edge cases (#1069, r-lib/testthat#1223).
 
--- a/R/trace.R
+++ b/R/trace.R
@@ -179,8 +179,10 @@ call_trace_context <- function(call, fn) {
     namespace <- ns_env_name(top)
     if (ns_exports_has(top, nm)) {
       scope <- "::"
-    } else {
+    } else if (env_has(top, nm)) {
       scope <- ":::"
+    } else {
+      scope <- "local"
     }
   } else {
     namespace <- NA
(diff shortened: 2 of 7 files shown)


jimhester (approved):
LGTM!

hadley (approved):

PR #1288, opened 2021-09-22
Fix C++11 detection in `use_rlang_c_library()`

DavisVaughan:
I'm pretty sure we need to pass through the project path in `has_cpp11_sysreq()`.

Also, `grep()` doesn't like the `++` in the `"C++11"` pattern and returns this without `fixed = TRUE`:

```r
Error in grepl("C++11", desc_lines, fixed = F) : 
  invalid regular expression 'C++11', reason 'Invalid use of repetition operators'
```


lionel-:
Thanks!

PR #1290, opened 2021-09-23
Tweaks to the call API

### LINKED ISSUE #670: call_name behavior with qualified functions
I found the behavior of `call_name` on qualified functions a little surprising. Is this a bug?

``` r
library(rlang)

call_name(~ func())
#> [1] "func"
call_name(~ func)
#> Error: `call` must be a call or must wrap a call (e.g. in a quosure)

call_name(~ pkg::func())
#> [1] "func"
# I would expect an error here, not `::`
call_name(~ pkg::func)
#> [1] "::"
```

<sup>Created on 2018-10-22 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

lionel-:
- Add `is_call_named()` to compose with `call_name()` and `call_ns()`. We now have the invariant that `call_name()` always returns a string when `is_call_named()` is `TRUE`. You can also supply the `ns = TRUE` argument to produce the same invariant with `call_ns()`.

- Fail with calls of the form `foo::bar` in `call_name()` and `call_namespace()`. Closes #670.

- Deprecate `call_fn()` which performs a weird evaluation, with questionable default environment when the input is not a quosure. Silently for now as it's used in packages like hardhat and sparklyr.

- Remove support for the deprecated frame objects in `call_args()` and `call_args_names()`.

- Remove experimental badge for `call_name()`, `call_args()`, etc. The changes above make me somewhat at peace with the API and it can still be useful. I've used `call_name()` and `call_ns()` in the backtrace code recently.


NEWS.md
```diff
@@ -33,9 +33,14 @@
 
 * `with_handlers()` is deprecated.
 
+* `call_fn()` is deprecated.
+
 
 ## Features and bugfixes
 
+* `call_name()` and `call_ns()` now fail with calls of the form
```
hadley, line 41: Mention `is_call_named()`?

R/call.R
```diff
@@ -825,51 +828,21 @@ call_match <- function(call = NULL,
...
-    recursive = abort("`call` does not call a named or inlined function"),
-    inlined = node_car(expr),
-    named = ,
-    namespaced = ,
-    eval_bare(node_car(expr), env)
-  )
-}
-
-#' Extract function name or namespace of a call
+#' # Is the function named?
```
hadley, line 838: "named" is a bit confusing here because it makes you think `names()`. I wonder if something like more like `is_call_simple()` would be clearer?

R/call.R
```diff
@@ -908,29 +877,50 @@ call_name <- function(call) {
   )
 }
 #' @rdname call_name
+#' @param x An object to test.
+#' @param ns Whether call is namespaced.
```
hadley, line 881:
```suggestion
#' @param ns Whether call can be namespaced.
```

hadley (approved):

R/call.R
lionel-: oh yes I like this

R/call.R
lionel-: This is now:

```
#' @param ns Whether call is namespaced. If `NULL`, `is_call_simple()`
#'   is insensitive to namespaces. If `TRUE`, `is_call_simple()`
#'   detects namespaced calls. If `FALSE`, it detects unnamespaced
#'   calls.
```

PR #1294, opened 2021-10-04
Replace `action = signaller` arguments by `error = handler`

lionel-:
To avoid tying the ellipsis functions to a fuzzy signaller interface (e.g. does the signaller support `call` arguments?).

Also more consistent with general condition handling idioms.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -63,6 +63,10 @@
 
 * `friendly_type()` is deprecated and `as_pairlist()` is defunct.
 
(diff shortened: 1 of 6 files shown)


lionel-:
Just realised that `warning()` and `message()` can be used as handlers directly. This makes this interface very similar in feel for users, except they supply base signallers instead of rlang ones.

```r
check_dots_used(action = warn)
check_dots_used(error = warning)
```

hadley (approved):
Interesting idea. I don't have strong feelings about the interface either way.

jimhester (approved):

PR #1299, opened 2021-10-11
Less aggressive ellipsizing in as_label

### LINKED ISSUE #1295: Aggressive ellipsizing in as_label
#1277 changed the behavior of `as_label` when the call contains operators. On current release of rlang, we see:

```
> as_label(quote(nchar(chr, type = "bytes", allowNA = TRUE) == 1))
[1] "nchar(chr, type = \"bytes\", allowNA = TRUE) == 1"
```

On dev rlang, we get

```
> as_label(quote(nchar(chr, type = "bytes", allowNA = TRUE) == 1))
[1] "... == 1"
```

The [tests](https://github.com/r-lib/rlang/commit/4ee1b05869272ad76c0a5247427549a78ac06775#diff-37deb3c3c1079c1f5433d780f478e3e2c9a8fae1fe70fcf3e02cb5c85d856514R618-R623) added in #1277 show that this is intentional behavior, but I wonder if this is too aggressive in shortening the output. Judging from the [criteria](https://github.com/r-lib/rlang/commit/4ee1b05869272ad76c0a5247427549a78ac06775#diff-9bea1e8787a8eaa532a36952bbdb166d6bee31fcdb8e00063abc8a09546f8445R899-R900) for whether to insert `...`, it looks like it is trying to hold to `deparse`'s default 60-character width (28 + 29 + 3 for the operator). But we aren't exceeding that in the example above, it's just that the left side is longer than 28 chars:

```
> nchar("nchar(chr, type = \"bytes\", allowNA = TRUE) == 1")
[1] 47
```

I understand the use case that #1277 intended to fix, but I'm wondering if you'd be interested in a PR that would limit this new behavior to where the total width of the deparsed expression was greater than 60 characters?

nealrichardson:
Resolves #1295

### DIFF
--- a/R/deparse.R
+++ b/R/deparse.R
@@ -892,13 +892,15 @@ as_label <- function(x) {
       if (is_data_pronoun(x)) {
         return(data_pronoun_name(x) %||% "<unknown>")
       }
-      if (call_print_type(x) %in% c("infix", "subset")) {
+      if (call_print_type(x) %in% c("infix", "subset") &&
+          length(expr_deparse(x, width = 60)) > 1) {
+        # Shorten the expression if we're too long, starting with the left side
         left <- expr_deparse(x[[2]], width = 29)
-        right <- expr_deparse(x[[3]], width = 28)
         if (length(left) > 1) {
           x[[2]] <- quote(...)
         }
-        if (length(right) > 1) {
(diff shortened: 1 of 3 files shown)


lionel-:
Thanks! I made some changes so that the RHS gets ellipsised first.

PR #1308, opened 2021-11-02
Typo fix (and, apparently, some whitespace)

jennybc:
The new reference index looks great! I noticed this while there.


lionel-:
Thanks!

PR #1309, opened 2021-11-04
Fix comment in test following #1299

nealrichardson:
In my original attempt at #1299, I truncated the left side before checking the right, but before merging @lionel- improved that to keep the left side intact if possible. That meant that the comment I left describing one of the tests wasn't accurate anymore.


lionel-:
Thanks!

PR #1311, opened 2021-11-11
Correct pak function reference

nathaneastwood:


lionel-:
`pak()` is from the dev version of pak.

nlneas1:
fair enough, explains why the link on the rlang website doesn't work then

nathaneastwood:
that explains why the link on the rlang website doesn't work then

lionel-:
Thanks!

PR #1323, opened 2021-12-02
Add `knitr::knit_cnd_format()` methods for rlang conditions

lionel-:


lionel-:
Not sure this is still needed.

PR #1327, opened 2021-12-07
Improve ellipsis errors

### LINKED ISSUE #1319: Improve empty dots errors
Could be made more user-friendly

```
Error in `is_installed()`: `...` is not empty.
i These dots only exist to allow future extensions and should be empty.
x We detected these problematic arguments:
* `..1`
i Did you misspecify an argument?
```

- Don't mention "future extensions", just that no arguments are expected in `...`.
- Suggest user has probably forgotten to give a name to an argument.
- Show the value of the argument.
- Try to reduce text density.
- Use conventional "must be" phrasing.

lionel-:
Closes #1319.

In all ellipsis errors:

* Show the value of the arguments.
* Try to reduce text density.
* Use conventional "must be" or "can't" phrasing.

In empty dots errors:

* Don't mention "future extensions", just that no arguments are expected in `...`.
* Suggest user has probably forgotten to give a name to an argument when unnamed dots are supplied and dots should be empty.

Before:

```r
is_installed("rlang", "0.99.0.9001")
#> Error in `is_installed()`: `...` is not empty.
#> i These dots only exist to allow future extensions and should be empty.
#> x We detected these problematic arguments:
#> * `..1`
#> i Did you misspecify an argument?
```

After:

```r
is_installed("rlang", "0.99.0.9001")
#> Error in `is_installed()`:
#> `...` must be empty.
#> x Problematic argument:
#> * ..1 = "0.99.0.9001"
#> i Did you forget to name an argument?
```

### DIFF
--- a/R/dots-ellipsis.R
+++ b/R/dots-ellipsis.R
@@ -104,7 +104,7 @@ check_dots_unnamed <- function(env = caller_env(),
   action_dots(
     error = error,
     action = action,
-    message = "Arguments in `...` can't be named.",
+    message = "Arguments in `...` must be passed by position, not name.",
     dots_i = named,
     class = "rlib_error_dots_named",
     call = call,
--- a/R/dots-ellipsis.R
+++ b/R/dots-ellipsis.R
@@ -55,19 +55,16 @@ check_dots <- function(env = caller_env(), error, action, call) {
   }
 
   proms <- ellipsis_dots(env)
-  used <- map_lgl(proms, promise_forced)
+  unused <- !map_lgl(proms, promise_forced)
 
-  unused <- names(proms)[!used]
   action_dots(
     error = error,
     action = action,
-    message = c(
-      "Arguments in `...` must be used.",
-      "x" = "Problematic arguments:"
-    ),
-    dot_names = unused,
+    message = "Arguments in `...` must be used.",
+    dots_i = unused,
     class = "rlib_error_dots_unused",
-    call = call
+    call = call,
+    env = env
   )
 }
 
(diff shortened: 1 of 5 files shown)


R/dots-ellipsis.R
```diff
@@ -55,16 +55,16 @@ check_dots <- function(env = caller_env(), error, action, call) {
...
-  used <- map_lgl(proms, promise_forced)
+  unused <- !map_lgl(proms, promise_forced)
 
-  unused <- names(proms)[!used]
   action_dots(
     error = error,
     action = action,
-    message = paste0(length(unused), " arguments in `...` were not used."),
-    dot_names = unused,
+    message = "Arguments in `...` must be used.",
```
hadley, line 63: Maybe all arguments?

R/dots-ellipsis.R
```diff
@@ -90,24 +90,25 @@ check_dots_unnamed <- function(env = caller_env(),
...
   }
+  named <- !unnamed
 
-  named <- names(proms)[!unnamed]
   action_dots(
     error = error,
     action = action,
-    message = paste0(length(named), " arguments in `...` had unexpected names."),
-    dot_names = named,
+    message = "Arguments in `...` can't be named.",
```
hadley, line 107: Must not be?

tests/testthat/_snaps/dots-ellipsis.md
```diff
@@ -23,19 +22,78 @@
...
-      i Did you misspecify an argument?
+      `...` must be empty.
+      x Problematic argument:
+      * xy = 4
+
+# expression contents are mentioned
+
+    Code
+      f("foo")
+    Error <rlib_error_dots_nonempty>
```
hadley, line 41: Are you deliberately using the old testthat style?

hadley (approved):
Looks like a nice improvement!

tests/testthat/_snaps/dots-ellipsis.md
lionel-: I can't use the new style because rlang isn't in its own Imports, so I'll wait for the new style to become the default.

R/dots-ellipsis.R
hadley: Maybe it would be better to remove the negative and match the error above: `All arguments in `...` must be unnamed` ?

R/dots-ellipsis.R
lionel-: I might be wrong but to me "All arguments" implies that some but not all of the arguments were used, which isn't a claim we can make here.

PR #1331, opened 2021-12-15
Add `empty` and `missing` arguments to `is_string()` and `is_character()`

lionel-:
To help type-check inputs for special values.

These arguments follow the matching style of other predicates. If `NULL`, any value is allowed. If `TRUE`, all values must match. If `FALSE` no value can match.

```r
is_character(c("foo", ""), empty = NULL)
#> [1] TRUE

is_character(c("", ""), empty = TRUE)
#> [1] TRUE
is_character(c("foo", ""), empty = TRUE)
#> [1] FALSE

is_character(c("foo", "foo"), empty = FALSE)
#> [1] TRUE
is_character(c("foo", ""), empty = FALSE)
#> [1] FALSE
```

This makes it easy to check that actual strings are supplied as input:

```r
is_character(c("foo", "foo"), empty = FALSE, missing = FALSE)
#> [1] TRUE
```

### DIFF
--- a/R/types.R
+++ b/R/types.R
@@ -68,14 +68,14 @@ is_complex <- function(x, n = NULL, finite = NULL) {
 #' @export
 #' @rdname type-predicates
 #' @inheritParams args_dots_empty
-#' @param missing,empty A logical value indicating whether to forbid
-#'   (`FALSE`) or enforce (`TRUE`) missing values or empty `""`. If
-#'   `NULL`, these values are ignored.
-is_character <- function(x,
-                         n = NULL,
-                         ...,
-                         missing = NULL,
-                         empty = NULL) {
+is_character <- function(x, n = NULL) {
+  .Call(ffi_is_character, x, n, NULL, NULL)
+}
+is_character2 <- function(x,
+                          n = NULL,
+                          ...,
+                          missing = NULL,
+                          empty = NULL) {
   check_dots_empty0(...)
   .Call(ffi_is_character, x, n, missing, empty)
 }
@@ -163,10 +163,12 @@ is_scalar_raw <- function(x) {
 #' @export
 #' @param string A string to compare to `x`. If a character vector,
 #'   returns `TRUE` if at least one element is equal to `x`.
-#' @inheritParams type-predicates
 #' @inheritParams args_dots_empty
 #' @rdname scalar-type-predicates
-is_string <- function(x, string = NULL, ..., empty = NULL) {
+is_string <- function(x, string = NULL) {
+  .Call(ffi_is_string, x, string, NULL)
+}
+is_string2 <- function(x, string = NULL, ..., empty = NULL) {
   check_dots_empty0(...)
   .Call(ffi_is_string, x, string, empty)
 }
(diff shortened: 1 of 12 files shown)


R/types.R
```diff
@@ -67,8 +67,17 @@ is_complex <- function(x, n = NULL, finite = NULL) {
 }
 #' @export
 #' @rdname type-predicates
-is_character <- function(x, n = NULL) {
-  .Call(ffi_is_character, x, n)
+#' @inheritParams args_dots_empty
+#' @param missing,empty A logical value indicating whether to forbid
+#'   (`FALSE`) or enforce (`TRUE`) missing values or empty `""`. If
```
hadley, line 72: What does enforcing an empty string or missing value mean?

lionel-: I am trying to find a helpful formulation for the matching behaviour of type predicates in rlang.

For instance `is_character(x, n = NULL)` allows any length. `is_character(x, n = 2)` requires a length 2.

In the same way, `is_character(x, empty = NULL)` allows empty as well as non-empty values. If set to `TRUE`, all values must be empty, If set to `FALSE` no values can be empty. The `FALSE` case is most helpful but occasionally you might set special values to `TRUE`, for instance `is_logical(x, missing = TRUE)` would detect unspecified vectors.

The meaning of enforce/forbid in this context is that these arguments adds type requirements that the input must match for the predicate to return `TRUE`. If the value doesn't match, `FALSE` is returned.

lionel-: @hadley oh maybe we should reframe this in terms of include and exclude lists.

Then `is_character(x, missing = FALSE, empty = FALSE)` would be:

```r
is_character(x, exclude = c("", NA))
```

And `is_integerish(x, finite = FALSE)` would be:

```r
is_integerish(x, exclude = c(NA, NaN, Inf, -Inf)
```

This would be much more flexible and probably easier to understand. Another example, this would be a predicate version of `arg_match()`:

```r
is_character(x, include = c("foo", "bar"))
```

hadley: That feels maybe a bit too generic to me (e.g. it'll be annoying to have to remember `c(NA, NaN, Inf, -Inf))` every time. How about a flag like `na_ok = TRUE`?

lionel-: oh we already have a `finite` argument in `is_integerish()`. I was just suggesting that it could be documented by showing the equivalent restrictions.

How about calling these restriction arguments `set` and `exclude`? Then these restrictions would be equivalent:

```r
# Restrict type to finite values
is_integerish(x, finite = TRUE)
is_integerish(x, exclude = c(NA, NaN, Inf, -Inf))

# Restrict type to non-finite values
is_integerish(x, finite = FALSE)
is_integerish(x, set = c(NA, NaN, Inf, -Inf))
```

And then:

```r
# Restrict expected character type to actual strings
is_character(x, exclude = c("", NA))

# Restrict expected character type to letters
is_character(x, set = c(letters, LETTERS))
```

This seems fairly natural to me, what do you think?

hadley: It still feels to general to me and it makes the implied type system too rich (I think we want to minimise the use of values). Restricting `NA` or `""` seems ok, as is filtering doubles to finite values, but I really don't think we want to expose other custom filtering to the user.

lionel-: Ok. I'll just note that a type system can be as rich as needed with adequate customisation points for runtime checks, e.g. https://wiki.haskell.org/Smart_constructors

PR #1332, opened 2021-12-15
Improve version requirements in `is_installed()` and `check_installed()`

lionel-:
Branched from #1331.

* DESCRIPTION-style requirements are now supported, e.g. `is_installed("rlang (>= 1.0.0)"`.
* New `compare` argument to supply requirement info programmatically.
* When `<` or `<=` is used, `check_installed()` doesn't prompt for install.

This will allow support for `check_installed()` in pkgload, so that we get install prompts on missing deps.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: rlang
-Version: 0.99.0.9002
+Version: 0.99.0.9003
 Title: Functions for Base Types and Core R and 'Tidyverse' Features
 Description: A toolbox for working with base types, core R features
   like the condition system, and core 'Tidyverse' features like tidy
--- a/NEWS.md
+++ b/NEWS.md
@@ -72,6 +72,11 @@
 
 ## Features and bugfixes
 
+* `is_installed()` and `check_installed()` now support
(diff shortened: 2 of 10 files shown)


lionel-:
@hadley In pkgload, a failing version requirement only issues a warning instead of an error, this way the package can be loaded anyway. I just want to check this is still the behaviour we want to support since that might be slightly tricky to implement. (Probably we need to signal a classed condition before calling `invokeRestart("abort")`.)

hadley (approved):

hadley:
In pkgload, I think we can just protect with some `tryCatch()`s, like we need to do for namespace imports.

hadley:
Oh but the idea is to now use `check_installed()` directly in pkgload, so we'd be replacing the warning with a prompt to install the package; I think that's ok.

lionel-:
> Oh but the idea is to now use check_installed() directly in pkgload, so we'd be replacing the warning with a prompt to install the package; I think that's ok.

ok cool, that will make things simpler!

We now take an optional `action` callback that is called when the user chooses to update/install dependencies.

Issue #1106, opened 2021-02-01
Confusing behaviour of is_missing() when an argument has a default


joeroe:
The documentation for is_missing() says that it is "like base::missing() but also supports testing for missing arguments contained in other objects like lists". I therefore expected these functions to be equivalent, but they're not:

```r
base_missing <- function(x = 1) {
  missing(x)
}

rlang_missing <- function(x = 1) {
  rlang::is_missing(x)
}

base_missing()
#> [1] TRUE
rlang_missing()
#> [1] FALSE
```

Probably I'm just misunderstanding the purpose of is_missing(), but perhaps the documentation could be updated to clarify that this is another way it differs from base::missing()?

aalucaci:
Hi,
I think you don't compare the same thing in the 2 cases. If you look at the definition of `rlang::is_missing`, you'll see it calls `base::is_missing` (it does a bit more, but that's not important for this comparison). The thing is, since your example function have default arguments, the argument is not really missing... not sure how to express this. Please see my examples below: if you nest the `base::is_missing` one frame more, as `rlang::is_missing` does it, you'll see that both base and rlang give the same results:

`
base_missing <- function(x = 1) {
  message("x: ", x)
  missing(x)
}
base_missing() # TRUE

###############################
missing_0 <- function(x) {
  message("x: ", x)
  missing(x)
}
base_missing_deep <- function(x = 1) {
  message("x: ", x)
  missing_0(x)
}
base_missing_deep() # FALSE

###############################
rlang_missing <- function(x = 1) {
  rlang::is_missing(x)
}
rlang_missing() # FALSE
`
I would say this has little to do with rlang, but with the way R handles default arguments.

Just my 2 cents,
Angela

joeroe:
Thank you for the additional information @aalucaci. I do understand *why* the behaviour is different, which is why I opened this as a documentation issue rather than a bug. The `missing()` documentation warns that it does not work reliably if nested in another function. You could debate whether arguments with a default are really "missing". IMO it is a useful feature that `missing()` treats them this way, e.g. when you want to override a default value but only if the user hasn't confirmed they want it. 

But more to the point, the documentation for `missing()` makes it obvious that this is the expected behaviour: "missing can be used to test whether a value was *specified as an argument* to a function." The documentation for `is_missing()` implies that it behaves in the same was as `missing()` just with added support for missing arguments contained inside other objects. I'm suggesting that it should make it clearer that it doesn't behave the same way in this instance.

[...]

PR #1344, opened 2022-01-18
Never treat default arguments as missing in `is_missing()`

lionel-:
We have changed `is_missing()` in this release to treat default arguments the same way as `base::missing()`, see #1106.

However, seeing failures in revdep checks made me reconsider this behaviour because the rules under which  a default argument is treated as missing seem a little complicated:

```r
bare <- function(x) missing(x)
default <- function(x = 1) missing(x)

bare_bare <- function(x) bare(x)
bare_default <- function(x) default(x)

default_bare <- function(x = 1) bare(x)
default_default <- function(x = 1) default(x)

bare()
#> [1] TRUE
default()
#> [1] TRUE

bare_bare()
#> [1] TRUE
bare_default()
#> [1] TRUE

default_bare()
#> [1] FALSE
default_default()
#> [1] FALSE
```

`missing()` within a function makes a difference between default and no default. This way, a function can implement special behaviour for "true" or "missing" defaults (versus supplied default).

If a default value comes from outside that function, `missing()` doesn't make a distinction, the value is deemed non-missing. Which makes sense since that function might be from a different package (and is from a different context in any case) so there should be no difference in behaviour in that case. Making a difference would leak internal implementation details from foreign contexts.

This all sort of makes sense (I'm assuming this isn't a bug in `base::missing()`) but is complicated and hard to predict. So this PR makes the `is_missing()` behaviour simpler: A default value is never treated as missing:

```r
bare <- function(x) is_missing(x)
default <- function(x = 1) is_missing(x)

bare()
#> [1] TRUE
default()
#> [1] FALSE

bare_bare()
#> [1] TRUE
bare_default()
#> [1] FALSE

default_bare()
#> [1] FALSE
default_default()
#> [1] FALSE
```


lionel-:
>  So this PR makes the is_missing() behaviour simpler: A default value is never treated as missing

This went a little too far. If a caller calls a function with a missing argument, a default defined in the callee is not applied by the interpreter, so the argument should be considered missing.

```r
callee <- function(x = 1) x
caller <- function(x) callee(x)
caller()
#> Error in `callee()`:
#> ! argument "x" is missing, with no default
```

The last commit restores the (much simpler) implementation of CRAN rlang. We get these results, which I think were correct all along:

```r
bare()
#> [1] TRUE
bare_bare()
#> [1] TRUE
bare_default()
#> [1] TRUE

default()
#> [1] FALSE
default_bare()
#> [1] FALSE
default_default()
#> [1] FALSE
```

Also:

```r
all(c(
  bare(missing_arg()),
  bare_bare(missing_arg()),
  bare_default(missing_arg()),
  default(missing_arg()),
  default_bare(missing_arg()),
  default_default(missing_arg())
))
#> [1] TRUE
```

lionel-:
@hadley Nerging now since I've restored the original implementation of `is_missing()`, but feel free to take a look. This PR now just adds documentation and unit tests.

Issue #1347, opened 2022-01-21
Confusing error message in `as_version_info()`


bobjansen:
Related to [this issue](https://github.com/r-lib/roxygen2/issues/1275) I raised for roxygen2:

The problem, having a malformed (name, version) tuple in an item in for example `Depends:` or `Imports` is handled better when using the very latest versions of `pkgload` and `rlang`. The improved error message shows:

```R
Error in `check_installed()`:
! Can't parse version in `pkg`.
Run `rlang::last_error()` to see where the error occurred.
```

I think this message is a tiny bit confusing as the user doesn't have `pkg` anywhere in their code. However, this error is `as_version_info()` which has as argument the individual items. I think it would be more helpful to show the first offending item (or maybe all items).

lionel-:
Can you post an example please?

bobjansen:
Sorry, I messed up the link to the roxygen2 issue. Now edited. The issue is as follows:

If a string such as `'rlang (>=1.0)'` is passed to `rlang:::as_version_info()` the following error is generated:

```
> rlang:::as_version_info('rlang (>1.0)') # No space between > and 1
Error:
! Can't parse version in `pkg`.
```
When I got this after running `devtools::document()` I found it quite confusing because I didn't understand what `pkg` was or where it came from. The PR changes the error message into:

```
> as_version_info('rlang (>1.0)')
Error:
! Can't parse version in `rlang (>1.0)`.
Run `rlang::last_error()` to see where the error occurred.
```

P.S. My original issue was with packages currently on CRAN. A misspecified item in 'Imports:' threw an error like `Error in pieces[[2]] : subscript out of bounds` during `document()` which was really confusing but thankfully solved now.

lionel-:
> I think this message is a tiny bit confusing as the user doesn't have pkg anywhere in their code.

When an argument name is mentioned, it (should) always belongs to the function mentioned in the error, in this case `check_installed()`. So I don't think we can do better here.

PR #1348, opened 2022-01-21
Improve error message in as_version_info()

bobjansen:
My proposal to clear up the error message in `as_version_info()` as described in #1347


lionel-:
Thanks for giving this a go. Should be fixed by https://github.com/r-lib/rlang/commit/2b472ffb74e264974186c641585d0ffe4e0ee893.

PR #1352, opened 2022-01-28
Review dynamic vector API

lionel-:
As discussed @DavisVaughan 

- Use `dyn_` instead of `arr_`
- Add poker and getter helpers

Also add `r_as_char()`, `r_as_complex()`, and `r_as_double()` helpers for input validation.


src/rlang/vec.c
```diff
@@ -226,3 +226,41 @@ void r_vec_poke_range(r_obj* x, r_ssize offset,
                       r_obj* y, r_ssize from, r_ssize to) {
   r_vec_poke_n(x, offset, y, from, to - from + 1);
 }
+
+bool _r_is_finite(r_obj* x) {
```
DavisVaughan, line 230: Is `_r` meaning that its extremely internal?

src/rlang/vec.h
```diff
@@ -266,6 +278,34 @@ r_obj* r_shared_lgl(bool x) {
   }
 }
 
+static inline
+bool _r_has_correct_length(r_obj* x, r_ssize n) {
+  return n < 0 || r_length(x) == n;
+}
+extern
```
DavisVaughan, line 285: Why `extern`?

DavisVaughan (commented):
These look great, thanks again!

src/rlang/vec.c
lionel-: yes it means it's not part of the API. I had to pull them from `internal/`, but the interface of these functions still need a bit of review.

src/rlang/vec.h
lionel-: oh it's not necessary here since it's not a variable. Not sure why I wrote this.

PR #1353, opened 2022-01-28
Don't use `_t` suffix

lionel-:
For consistency with `r_ssize`, `r_obj`, etc., rename:

- `r_complex_t` to `r_complex`
- `r_keep_t` to `r_keep_loc`


DavisVaughan (approved):
Very much agree. See also https://stackoverflow.com/questions/231760/what-does-a-type-followed-by-t-underscore-t-represent which suggests that you don't add `_t` in general to avoid confusion.

I also like `_loc` for the protect index!

PR #1357, opened 2022-02-01
Include `<exception>` header in `rlang.hpp`

DavisVaughan:
It is used in `rcc_abort()` below. The only reason this hasn't caused issues so far is because the only place this header is included is in `vec.cpp`, which includes `<algorithm>` first and that includes the exception header, so we don't see the issue

I have also flipped the order of includes in `vec.cpp` because I believe that including your own headers first is better practice. See also https://github.com/tidyverse/readxl/commit/1059a768436df38fb7b33f8e0415c18585c11ac5

### DIFF
--- a/src/rlang/cpp/vec.cpp
+++ b/src/rlang/cpp/vec.cpp
@@ -1,5 +1,5 @@
-#include <algorithm>
 #include <rlang.hpp>
+#include <algorithm>
 
 extern "C" {
 
--- a/src/rlang/rlang.hpp
+++ b/src/rlang/rlang.hpp
@@ -1,4 +1,5 @@
 #include <cmath>
+#include <exception>
 using std::isfinite;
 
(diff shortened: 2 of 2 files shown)


lionel-:
> because I believe that including your own headers first is better practice

Could you elaborate? Is this to allow the local header to include external headers with special dispositions?

lionel- (approved):

DavisVaughan:
The top answer [here](https://stackoverflow.com/questions/2762568/c-c-include-header-file-order) describes it pretty well but the basic idea is that it avoids this very issue because it ensures that each header is self-contained. i.e. there is no way to get a hidden dependency like this where `std::exception` really did require the `<exception>` header, but it was hidden behind the fact that including `<algorithm>` in another header already included it. This makes it frustrating to use `<rlang.hpp>` anywhere else, because there was no way to know ahead of time that you needed to get `<exception>` included somehow

PR #1359, opened 2022-02-01
Rename `r_syms.class` to `r_syms.class_`

DavisVaughan:
Follow up to https://github.com/r-lib/rlang/commit/70c90a8be9987c0b1fb96a74fd9d84d839b5d7ae and https://github.com/r-lib/rlang/commit/30cae2afc181c3ddfbb59a3fd5d323f919831413

We previously added workarounds to the fact that C++ doesn't like you to use `class` as a variable name at all by redefining it to `klass` before including `<rlang.h>`. This currently works, but I am afraid it is extremely brittle and is the reason for many other hacks we have had to do to make the C and C++ code here play nicely together.

I propose moving from `r_syms.class` to `r_syms.class_`, which is better than `cls` or `klass` because it works with auto-complete. That way in the future we won't ever think that we don't have a symbol for `class` and try and add one back in.

If you look at the `rlang.hpp` file now, it has become extremely simple, with all hacks removed.

---

There are 3 distinct hacks that we had to make, which we no longer have to make if we make this change.

1. The early inclusion of `<Rinternals.h>`. We did this because in rray3 I had found that this was necessary when using my own custom top level `rlang.h` C++ header that included the rlang C library. i.e. this commit https://github.com/DavisVaughan/rray3/commit/d5a0e4d3beb050816e886993c17673e61560b373#diff-25a6634263c1b1f6fc4697a04e2b9904ea4b042a89af59dc93ec1f5d44848a26. Now that I am just supposed to include the `<rlang.hpp>` that rlang already provides, this is no longer needed. I have tested here that we can no longer reproduce the original Mac issue on GitHub actions after making this change https://github.com/DavisVaughan/rray3/pull/1

2. The need for `klass`. This one is self explanatory, but now that we use `class_` we don't need to redefine the `class` symbol to `klass`. This is a big win though, because it leads into point 3

3. No need for early includes of `<cmath>` and `using std::isfinite`. This one is a bit of a head scratcher, but I think I figured out how this is related to `klass`.
    - `rlang.hpp` does the `extern "C"` inclusion of `<rlang.h>`, the whole rlang lib
    - That includes the `c-utils.h` header, which requires including `math.h` for `isfinite()` https://github.com/r-lib/rlang/blob/9df0a25768d6c7ba32da79a3aa3dcc8da3c69e42/src/rlang/c-utils.h#L4
    - Now, `math.h` is defined in both C and C++, and the C++ one is a bit special and is the one getting included here. On my computer, it has the special behavior in the code block below. This means that it swaps back to C++ linkage temporarily to add definitions for some template functions, which involve using the `class` keyword. Now, we redefined `class` to be `klass` at this moment in time, which C++ complains about because it looks like an unknown type (see further below). The "fix" was to include `<cmath>` very early on, which ensures that this all gets included before the redefinition to `klass`. But I am worried there are other headers that do the same thing, and trying to use them in the C rlang library will end up causing issues like this in the C++ bits.

```c++
#ifdef __cplusplus

// We support including .h headers inside 'extern "C"' contexts, so switch
// back to C++ linkage before including these C++ headers.
extern "C++" {

#include <type_traits>
#include <limits>

// signbit

#ifdef signbit

template <class _A1>
_LIBCPP_INLINE_VISIBILITY
bool
__libcpp_signbit(_A1 __lcpp_x) _NOEXCEPT
{
    return signbit(__lcpp_x);
}

#undef signbit
``` 

```c++
In file included from /Library/Developer/CommandLineTools/usr/include/c++/v1/math.h:310:
   /Library/Developer/CommandLineTools/usr/include/c++/v1/limits:141:20: error: a non-type template parameter cannot have type 'class _Tp'
   template <class _Tp, bool = is_arithmetic<_Tp>::value>
                      ^
   /Library/Developer/CommandLineTools/usr/include/c++/v1/limits:142:1: error: unknown type name 'klass'; did you mean 'class'?
   class __libcpp_numeric_limits
   ^
   ./rlang/rlang.hpp:17:15: note: expanded from macro 'class'
   #define class klass
                 ^
```

### DIFF
--- a/src/internal/dots.c
+++ b/src/internal/dots.c
@@ -1065,7 +1065,7 @@ void rlang_init_dots(r_obj* ns) {
     r_preserve(splice_box_attrib);
     r_mark_shared(splice_box_attrib);
 
-    r_node_poke_tag(splice_box_attrib, r_syms.class);
+    r_node_poke_tag(splice_box_attrib, r_syms.class_);
     FREE(1);
   }
 
@@ -1086,7 +1086,7 @@ void rlang_init_dots(r_obj* ns) {
     r_preserve(quosures_attrib);
     r_mark_shared(quosures_attrib);
 
-    r_node_poke_tag(quosures_attrib, r_syms.class);
+    r_node_poke_tag(quosures_attrib, r_syms.class_);
     FREE(1);
   }
 
--- a/src/internal/eval-tidy.c
+++ b/src/internal/eval-tidy.c
@@ -45,7 +45,7 @@ static r_obj* rlang_new_data_pronoun(r_obj* mask) {
   r_obj* pronoun = KEEP(r_alloc_list(1));
 
   r_list_poke(pronoun, 0, mask);
-  r_attrib_poke(pronoun, r_syms.class, data_pronoun_class);
+  r_attrib_poke(pronoun, r_syms.class_, data_pronoun_class);
 
   FREE(1);
   return pronoun;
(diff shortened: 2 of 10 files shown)


lionel- (approved):
That makes sense. `class_` seems like a reasonable workaround.

PR #1361, opened 2022-02-02
Rethrow unchained

lionel-:
- To rethrow unchained errors, users of `abort()` must now pass `parent = NA` to let it know that an error is rethrown from a condition handler. This way we can detect the handling context to soft-truncate backtraces and use the correct default call (same as when a condition is passed as `parent`).

- When `parent` is supplied, we now loop over the callers to find the condition handler frame. The goal is to make it possible to wrap handler functions without having to pass `.frame` with `call` and `parent`. Passing `.frame` would likely be easy to forget and confusing to everyone, me included.

- We now more consistently use the caller of the condition setup frame (e.g. the caller of `withCallingHandlers()`) as default call in case of rethrow.

cc @hadley @DavisVaughan


R/cnd-abort.R
```diff
@@ -53,7 +54,14 @@
 #'
 #'   Can also be `NULL` or a [defused function call][topic-defuse] to
 #'   respectively not display any call or hard-code a code to display.
-#' @param body,footer Additional bullets.
+#' @param parent Supply `parent` when you rethrow an error from a
+#'   condition handler (e.g. with [try_fetch()]).
+#'
+#'   If `parent` is a condition object, a _chained error_ is
+#'   created. If `parent` is `NA`, it indicates an unchained rethrow.
```
DavisVaughan, line 61: It might be useful to add additional context about when these two cases are useful. Something like:

If `parent` is a condition object, a _chained error_ is created, which is useful when you want to enhance an error with more details, while still retaining the original information.

If `parent` is `NA`, it indicates an unchained rethrow, which is useful when you want to take ownership over an error and rethrow it with a custom message that better fits the surrounding context.

R/cnd-signal.R
```diff
@@ -62,7 +62,7 @@ cnd_signal <- function(cnd, ...) {
         frame <- caller_env()
       }
       if (is_null(cnd$trace)) {
-        info <- abort_context(frame, cnd$parent)
+        info <- abort_context(frame, !is_null(cnd$parent))
```
DavisVaughan, line 65:
```r
rethrowing <- !is_null(cnd$parent)
info <- abort_context(frame, rethrowing)
```

I would personally find giving this a name a little clearer

R/cnd-abort.R
lionel-: Good point and great suggestions! This param doc is now much better: https://github.com/r-lib/rlang/commit/5196cd3055d4f9d7c6c82747bc670074dbde50eb

R/cnd-signal.R
lionel-: Like `stack_info`?

DavisVaughan: No I'm literally talking about just giving this intermediate variable a name, rather than just inlining `!is_null(cnd$parent)`. It just makes it a little easier to read the code, otherwise I have to go back to the definition of `abort_context()` to see what the second argument is. This is a very minor comment

lionel-: oh this is in `cnd_signal()`, got it.

https://github.com/r-lib/rlang/commit/321840b776c58008f32d0ffe3e7121117ec7d3ac

PR #1362, opened 2022-02-08
use consistent formatting for base version skip

MichaelChirico:
Using 4.0.0 here:

https://github.com/r-lib/rlang/blob/68f768deed899ddded2ea102217531bc6e5a0b1e/tests/testthat/test-env.R#L364

There are also usages skipping based on `getRversion`, e.g.:

https://github.com/r-lib/rlang/blob/68f768deed899ddded2ea102217531bc6e5a0b1e/tests/testthat/test-compat-vctrs.R#L348

Should we unify those as well?

----

Of course it's somewhat trivial, but the test output is a bit noisy with the inconsistency. e.g. here's running from 3.6.3:

```
══ Skipped tests ═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
• Cannot set MBCS locale (1)
• !compiled_by_gcc() is TRUE (1)
• Disabled (1)
• Failing (1)
• getRversion() < "4.0" is TRUE (1)
• Installed base is version 3.6.3; but 4.0.0 is required (1)
• Installed base is version 3.6.3; but 4.0 is required (2)
• Not on Windows (1)
```

### DIFF
--- a/tests/testthat/test-cnd-entrace.R
+++ b/tests/testthat/test-cnd-entrace.R
@@ -229,7 +229,7 @@ test_that("can supply handler environment as `bottom`", {
 })
 
 test_that("can set `entrace()` as a global handler", {
-  skip_if_not_installed("base", "4.0")
+  skip_if_not_installed("base", "4.0.0")
 
   expect_snapshot_output(run('{
     suppressMessages(testthat::local_reproducible_output())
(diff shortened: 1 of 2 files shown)


lionel-:
I like the level of attention to details 😅

MichaelChirico:
just a consequence of our tests failing more than in usual environments 😂

PR #1398, opened 2022-05-01
Update docs for exprs_auto_name()

krlmlr:
The current docs are pointing to questioning functions.

Should we soft-deprecate `quos_auto_name()`? It's forwarding to `exprs_auto_name()`, with fewer arguments.


lionel-:
Thanks!

> Should we soft-deprecate quos_auto_name()? It's forwarding to exprs_auto_name(), with fewer arguments.

Probably yes.

PR #1400, opened 2022-05-03
Add hyperlinks to src_loc()

### LINKED ISSUE #1396: Add links to back trace
Using `cli::style_hyperlink()` or similar.

Could also hyperlink the advice to run call `rlang::last_trace()` to save a little typing.

hadley:
A very quick POC.

Fixes #1396

### DIFF
--- a/R/trace.R
+++ b/R/trace.R
@@ -884,13 +884,24 @@ src_loc <- function(srcref) {
   }
 
   if (is_null(peek_option("rlang:::disable_trim_srcref"))) {
-    file <- path_trim_prefix(file, 3)
+    file_trim <- path_trim_prefix(file, 3)
+  } else {
+    file_trim <- file
   }
 
   line <- srcref[[1]]
   column <- srcref[[5]] - 1L
 
-  paste0(file, ":", line, ":", column)
+  if (is_installed("cli")) {
+    cli::style_hyperlink(
+      paste0(file_trim, ":", line, ":", column),
+      paste0("file://", file),
+      params = c(line = line, col = column)
+    )
+  } else {
+    paste0(file_trim, ":", line, ":", column)
+  }
+
 }
 
 trace_root <- function() {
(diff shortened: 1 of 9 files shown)


R/trace.R
```diff
@@ -884,13 +884,24 @@ src_loc <- function(srcref) {
...
+  } else {
+    file_trim <- file
   }
 
   line <- srcref[[1]]
   column <- srcref[[5]] - 1L
 
-  paste0(file, ":", line, ":", column)
+  if (is_installed("cli")) {
+    cli::style_hyperlink(
```
DavisVaughan, line 896: Would you get weird behavior if you don't have a new enough version of cli?

hadley: Yes, hence this being a quick POC 😄

hadley:
Another idea: should clicking on a function name in the back trace display its source code?

lionel-:
Nice idea.

hadley:
Yeah, agreed. Let's leave that for now.

One other simple thing we could do in this PR would be to make `rlang::last_trace()` and `rlang::last_error()` clickable.

hadley:
I think this is probably worth a final review and then we can merge it.

R/utils.R
```diff
@@ -358,3 +358,10 @@ cli_with_whiteline_escapes <- function(x, fn) {
   x <- gsub("__NEW_LINE__", "\n", x, fixed = TRUE)
   x
 }
+
+style_rlang_run <- function(code) {
+  style_hyperlink(
+    paste0("`rlang::", code, "`"),
+    paste0("rstudio:run:rlang::", code)
```
lionel-, line 365: We were discussing with @romainfrancois about making the protocol more general:

- Check if `R_CLI_HAS_HYPERLINK_RUN` envvar is set.
- If set, hyperlink to `ide:run:`

PR #1401, opened 2022-05-11
Implement `use_compat()`

DavisVaughan:
Used like `use_compat("r-lib/rlang", "purrr")`

- Doesn't overwrite by default
- Requires `r-lib/rlang` to have a `main` branch. This is where we download the source code from.
- This is an internal function

### DIFF
--- a/R/c-lib.R
+++ b/R/c-lib.R
@@ -15,7 +15,7 @@ use_rlang_c_library <- function() {
 
   rlang_path <- Sys.getenv("RLANG_PATH")
   if (!nzchar(rlang_path)) {
-    rlang_path <- download_rlang()
+    rlang_path <- download_package("r-lib", "rlang")
     on.exit(fs::dir_delete(rlang_path))
   }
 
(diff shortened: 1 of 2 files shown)


DavisVaughan:
Decide if https://usethis.r-lib.org/reference/use_github_file.html is good enough instead

lionel-:
I think we'll still need a wrapper around `use_github_file()` so that we can build a simple dependency system into it. I'm thinking about something like this:

```r
.compat_deps <- c("compat-rlang.R", "compat-cli.R")
```

We'd import/update these deps as well recursively.

R/c-lib.R
```diff
@@ -77,16 +77,24 @@ use_rlang_c_library <- function() {
   }
 }
 
-download_rlang <- function() {
-  dest_zip <- fs::file_temp("rlang-src")
+download_package <- function(owner, repo) {
+  # Hardcoded for now
+  branch <- "main"
```
jennybc, line 82: In general, you can often skirt the `master` vs. `main` issue by using `HEAD` as the ref.

lionel-:
Closed by https://github.com/r-lib/usethis/pull/1748

PR #1404, opened 2022-05-17
Use stdout() for messages in all RStudio panes

### LINKED ISSUE #1393: Inform should use stdout in RStudio build pane
Can use cli heuristics (https://github.com/r-lib/cli/blob/main/R/rstudio-detect.R#L88-L96) or wait until IDE provides env var (https://github.com/rstudio/rstudio/issues/11024).

hadley:
This requires the latest daily for correct detection. I think it's probably not worth bundling in the cli heuristics as this is a relatively small improvement and it'll gradually resolve itself over time.

Fixes #1393


R/cnd-signal.R
```diff
@@ -215,6 +215,10 @@ default_message_file <- function() {
   }
 }
 
+is_rstudio <- function() {
+  Sys.getenv("RSTUDIO_SESSION_PID") == Sys.getpid()
```
lionel-, line 219: This made me pause until I realised `Sys.getenv()` is uniformly typed (always length 1, always non missing). Maybe use `is_string()` or `identical()` just for explicitness?

lionel- (approved):

R/cnd-signal.R
hadley: Yeah, it's a bit weird but I don't see a better way that isn't much more verbose

lionel-:
oh or maybe export it as a C callable from somewhere e.g. ps? Then we could have two code paths, depending on whether ps is available. This is probably better than a compat file.

gaborcsardi:
The C code is basically standalone, just remove `errors.h` and the `R_THROW_ERROR()` call. I think it better to include it in rlang than relying on ps and cross-package calls. 

Also, if ps is not available, then you fall back to some inferior behavior? I think it is just better to include this C code, it is pretty minimal, I don't see any downsides of using it.

The R code is already a compat file: https://github.com/r-lib/cli/blob/main/R/rstudio-detect.R (but this does not have support for the new env var yet).

lionel-:
ok if you think this piece of code is set in stone and won't require updates let's do that.

lionel-:
I've added `r_getppid()` to the rlang C library, now used in `is_rstudio()`.

@hadley I guess `abort()` should also output to stdout in these panes?

hadley:
This looks good to me. (It appears to reveal some downstream issues with roxygen2, but I think that's now a matter of deliberately suppressing cli formatting when executing Rmd code chunks)

PR #1411, opened 2022-05-25
Improve annotation of local functions in backtraces

### LINKED ISSUE #1399: could trace_back() better annotate functions "hidden" in closures?
For example, consider the following:

```
library(rlang)
trace(getOption, quote(print(rlang::trace_back())))
local({ rlang:::defer(42) })
```

I see:

```
Tracing getOption("knitr.in.progress") on entry 
     ▆
  1. ├─base::local(...)
  2. │ └─base::eval.parent(substitute(eval(quote(expr), envir)))
  3. │   └─base::eval(expr, p)
  4. │     └─base::eval(expr, p)
  5. └─base::eval(...)
  6.   └─base::eval(...)
  7.     └─rlang:::defer(42)
  8.       └─rlang add_handler(...)
  9.         └─rlang exit_frame(envir, frames, calls)
 10.           └─rlang in_knitr(envir)
 11.             └─rlang knitr_in_progress()
 12.               ├─base::isTRUE(getOption("knitr.in.progress"))
 13.               └─base::getOption("knitr.in.progress")
```

The "internal" functions like `in_knitr` and `knitr_in_progress` are marked as part of the `rlang` namespace, but IMHO it could be made more explicit that those are part of a separate internal closure somehow.

For what it's worth, this was prompted by me reading that output and thinking I would've been able to reference those functions via `rlang:::in_knitr`, but that failed since they're really in `defer()`, e.g. you could find them in `environment(rlang:::defer)$in_knitr`)

(feel free to close if you feel this is too narrow, since this pattern shouldn't be especially common)

lionel-:
Closes #1399


R/trace.R
```diff
@@ -857,7 +857,7 @@ trace_call_text <- function(call, collapsed, namespace, scope) {
   if (is_string(scope, "global")) {
     text <- paste0("global ", text)
   } else if (is_string(scope, "local") && !is_na(namespace)) {
-    text <- paste0(namespace, " ", text)
+    text <- paste0(namespace, " (local) ", text)
```
hadley, line 860: Does it still make sense to include the namespace here?

lionel-: I think so because these functions do belong to the namespace, they are just not accessible easily.

hadley: Do they? I'd say they belong to a function that belongs to the namespace.

lionel-: I agree but I see belonging as a transitive property. At a high level the goal of showing the package names is to make it clearer which packages are responsible for which calls. I would prefer not to hide this information for local functions.

hadley: It just feels to me like the namespace and the (local) are fighting each other. Would you mind seeing what it looks like if you just drop the namespace for local functions and don't add (local)?

lionel-: I see what you mean. Here is what it looks like without namespace: https://github.com/r-lib/rlang/commit/62f8a16627ff6c34349a9c3fc1374cd959f3e5cc

Here is an example where I think this is a regression:

```r
f <- function() {
  helper <- function() abort("foo")
  tryCatch(helper())
}
environment(f) <- ns_env("rlang")
env_unlock(ns_env("rlang"))
env_bind(ns_env("rlang"), f = f)

f()
#> Error in `helper()`:
#> ! foo

last_error()
#> <error/rlang_error>
#> Error in `helper()`:
#> ! foo
#> ---
#> Backtrace:
#>  1. rlang:::f()
#>  4. helper()
#> Run `rlang::last_trace()` to see the full context.

last_trace()
#> rlang::last_trace()
#> <error/rlang_error>
#> Error in `helper()`:
#> ! foo
#> ---
#> Backtrace:
#>     ▆
#>  1. └─rlang:::f()
#>  2.   ├─base::tryCatch(helper())
#>  3.   │ └─tryCatchList(expr, classes, parentenv, handlers)
#>  4.   └─helper()
#>  5.     └─rlang::abort("foo")
```

Compare to the current output:

```r
last_error()
#> <error/rlang_error>
#> Error in `helper()`:
#> ! foo
#> ---
#> Backtrace:
#>  1. rlang:::f()
#>  4. rlang helper()
#> Run `rlang::last_trace()` to see the full context.

last_trace()
#> <error/rlang_error>
#> Error in `helper()`:
#> ! foo
#> ---
#> Backtrace:
#>     ▆
#>  1. └─rlang:::f()
#>  2.   ├─base::tryCatch(helper())
#>  3.   │ └─base tryCatchList(expr, classes, parentenv, handlers)
#>  4.   └─rlang helper()
#>  5.     └─rlang::abort("foo")
```

So I suggest we just keep the current behaviour, what do you think?

PR #1412, opened 2022-05-25
Display both last_error() and last_trace()

hadley:
This saves me a click


hadley:
I can't reproduce the snapshots locally due to unrelated problems so someone else will have to finish this off 😞

PR #1415, opened 2022-05-28
Update for roxygen2

hadley:
Should give same results as released roxygen2 as long as you run it in an environment where `cli::num_ansi_colors()` is zero.


lionel-:
Thanks!

PR #1416, opened 2022-06-07
Add `env_coalesce()`

lionel-:
Implemented in the rlang C library as `r_env_coalesce()`. `env_clone()` and `r_env_clone()` are now implemented using coalesce.


NEWS.md
```diff
@@ -1,5 +1,9 @@
 # rlang (development version)
 
+* New `env_coalesce()` function to copy bindings from one environment
```
hadley, line 3: Another option would be `env_modify()` by analogy to `modifyList()`. I don't mind `coalesce()` either.

hadley (approved):

NEWS.md
lionel-: hmm good point but I think this would suggest a similar UI to `call_modify()`.

PR #1429, opened 2022-06-23
Use prettyunits to format bytes

lionel-:
It would be nice to provide a global option to format bytes with binary prefixes. This requires https://github.com/r-lib/prettyunits/issues/29.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # rlang (development version)
 
+* The `rlib_bytes` class now uses prettyunits to format bytes. The
+  bytes are now represented with decimal prefixes instead of binary
+  prefixes.
+
 * Supplying a frame environment to the `call` argument of `abort()`
   now causes the corresponding function call in the backtrace to be
   highlighted.
--- a/R/bytes.R
+++ b/R/bytes.R
@@ -106,7 +106,7 @@ parse_bytes <- function(x) {
   stopifnot(is_character(x))
 
   pos <- regexpr(
-    "^(?<size>[[:digit:].]+)\\s*(?<unit>[KMGTPEZY]?)i?[Bb]?$",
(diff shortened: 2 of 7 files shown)


gaborcsardi:
AFAIR I used `kB` because `KiB` might be unfamiliar to (non-CS) people. 

These sizes are approximate, so it does not actually matter if 1024 bytes are "1 kB" (or "1 KiB") or  "1.02 kB".

lionel-:
What do you mean by approximate? When examining memory allocations with artificial use cases (benchmarking, debugging), the sizes are not approximate right? For instance we snapshot memory allocs performed by vctrs to monitor them over time.

gaborcsardi:
The output of `pretty_bytes()` is approximate:
```
❯ prettyunits::pretty_bytes(1020)
[1] "1.02 kB"

❯ prettyunits::pretty_bytes(1024)
[1] "1.02 kB"
```

lionel-:
Got it. I was referring to artifical cases:

```r
prettyunits::pretty_bytes(1000)
#> [1] "1 kB"
```

lionel-:
There is a greater chance for memory allocations to be displayed as round numbers using binary prefixes:

```r
# Before

as_bytes(1024)
#> [1] 1KB

as_bytes(1024 * 8)
#> [1] 8KB

as_bytes(1024 * 8 + 64)
#> [1] 8.06KB


# After

as_bytes(1024)
#> [1] 1.02 kB

as_bytes(1024 * 8)
#> [1] 8.19 kB

as_bytes(1024 * 8 + 64)
#> [1] 8.26 kB
```

I guess this is why Jim decided to use that representation for packages like bench.

gaborcsardi:
I don't mind which one you use, we can add support for binary sizes in prettyunits.

lionel-:
yeah that'd be nice, no rush though.

PR #1435, opened 2022-07-02
missing is_dictionaryish from pkgdown

mmuurr:
Maybe this omission was (once) intentional (hinted-at by the undoing here: afc195846209f84015b3871ded3f9dc99af8d2ee).

If not intentional, this adds it to the pkgdown site.

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -315,7 +315,7 @@ reference:
       - names2
       - has_name
       - is_named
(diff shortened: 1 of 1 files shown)


lionel-:
It's intentionally omitted. You can see this line in the documentation of functions that we consider internal or removed from the public eye:

```r
#' @keywords internal
```

I recommend not using this function.

PR #1439, opened 2022-07-07
Fix `is_installed("base")`

### LINKED ISSUE #1434: `is_installed()` doesn't correctly detect base packages on 4.0 and earlier
i.e. `rlang::is_installed("base")` returns `FALSE`

lionel-:
Closes #1434

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # rlang (development version)
 
+* `is_installed()` and `check_installed()` now properly detect that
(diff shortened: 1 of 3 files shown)


lionel-:
Only the base package is special so I wasn't worried about the other packages. I've added tests for all default packages in https://github.com/r-lib/rlang/commit/c4874ae84054c8f0ee913b428db38c6502b6aa4d.

PR #1443, opened 2022-07-18
Increment version number to 1.0.4.9000

DavisVaughan:
Just as a reminder, not sure if you had any intermediate changes between the HEAD of `main` and CRAN actually accepting 1.0.4


lionel-:
oh that's now already done

PR #1449, opened 2022-07-26
Fix returning invisibly

JanJecmen:
Hi, I just randomly stumbled upon this. I believe the intention was to return invisibly, however, that is not what R does:
```
f = function() invisible(return(5))
f()
## [1] 5
f = function() return(invisible(5))
f()
##
```
`invisible` is a builtin that evaluates its arguments eagerly, and `return` uses a non-local jump to get out of the function, skipping the visibility change.


lionel-:
Thanks!

PR #1461, opened 2022-08-11
Turn `types-check.R` into a true compat file

DavisVaughan:
@lionel- do you have any way of managing compat dependencies? i.e. this compat file requires `compat-obj-type.R` but I don't know where to record that information


DavisVaughan:
Oh hmm `is_number()` is used in this compat file but it is not exported from rlang. It lives in `types.R`

PR #1462, opened 2022-08-12
Tweak `r_arg_as_*()` error messages

DavisVaughan:
As noted in https://github.com/r-lib/vctrs/pull/1616#discussion_r944609452

### DIFF
--- a/src/rlang/vec.h
+++ b/src/rlang/vec.h
@@ -339,7 +339,7 @@ bool r_is_string(r_obj* x) {
 static inline
 bool r_arg_as_bool(r_obj* x, const char* arg) {
   if (!r_is_bool(x)) {
-    r_abort("`%s` must be a logical value.", arg);
+    r_abort("`%s` must be `TRUE` or `FALSE`.", arg);
   }
   return r_lgl_get(x, 0);
 }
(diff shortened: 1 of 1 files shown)


lionel- (approved):
Should we export `r_obj_type_friendly()` via C callable?

src/rlang/vec.h
```diff
@@ -339,7 +339,7 @@ bool r_is_string(r_obj* x) {
 static inline
 bool r_arg_as_bool(r_obj* x, const char* arg) {
   if (!r_is_bool(x)) {
-    r_abort("`%s` must be a logical value.", arg);
+    r_abort("`%s` must be `TRUE` or `FALSE`.", arg);
```
lionel-, line 342: By the way the rlang lib exports `r_format_lazy_error_arg()`, this returns a vmax-protected C string that is equivalent to `{.arg }` formatting.

DavisVaughan:
> r_obj_type_friendly

That sounds useful! Currently our C errors are less useful than ones thrown by things like `rlang:::check_bool()` because of that

lionel-:
I've opened https://github.com/r-lib/rlang/issues/1463

PR #1479, opened 2022-09-05
Avoid extra work in the good case in `try_catch()`

krlmlr:
Should be as lean as possible.

I wonder if `try_catch()` could be implemented entirely from C via `R_tryCatch()` or `R_tryCatchError()`, or perhaps with new entry points. It seems that it's currently constructing R code of the form `tryCatch(withCallingHandlers(...))` (or just `withCallingHandlers(...)` in some cases), which is very slow.

### DIFF
--- a/R/cnd-handlers.R
+++ b/R/cnd-handlers.R
@@ -185,10 +185,10 @@ try_fetch <- function(expr, ...) {
   frame <- environment()
 
   catch <- value <- NULL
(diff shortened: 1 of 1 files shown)


lionel-:
`R_tryCatch()` is a wrapper around the R-level `tryCatch()`. Also it is buggy in older R versions so we couldn't use it. And we mostly need `withCallingHandlers()` semantics anyway.

Issue #1481, opened 2022-09-07
Consider `check_` functions


hadley:
```R
library(rlang)

# Context:
# * user created scalars
# * normally typed directly in the function call
# * complementary to vctrs style checks
# * lives in rlang compat file

check_number <- function(x, 
                         min = NULL,
                         max = NULL,
                         allow_decimal = FALSE,
                         allow_missing = FALSE,
                         allow_infinite = FALSE,
                         allow_null = FALSE
                         )
  
  # Should this be inclusive or exclusive?
  if (!is.null(min) && x < min) {
    abort("`x` must be greater than {min}, not {x}")
  }
}

check_bool <- function(x, allow_missing = FALSE, allow_null = FALSE) {
}

check_string <- function(x, 
                         allow_empty = FALSE,
                         allow_missing = FALSE,
                         allow_null = FALSE,
                         ) {
  
}

check_enum <- function(x, values = NULL) {
  
}
```

jennybc:
I just pulled `compat-types-check.R` and `combat-obj-type.R` into a gargle branch and this spurred 1 question and 1 suggestion.

Question: gargle already imports rlang and, indeed, imports the entire rlang namespace. I don't understand why I need to bring in a compat file to get rlang's checkers, for example. Why wouldn't these just be exported for other packages to use? I associate compat files with the desire to avoid a dependency.

Observation: Consider the definition of `check_string()` at the time I'm writing this:

``` r
check_string <- function(x,
                         ...,
                         what = "a single string",
                         arg = caller_arg(x),
                         call = caller_env()) {
  if (!is_string(x)) {
    stop_input_type(x, what, ..., arg = arg, call = call)
  }
}
```

If used like so:

``` r
gargle_oauth_client <- function(
    id,
    ...
) {
  check_string(id)
  ...
}
```

I see

``` r
gargle_oauth_client()
#> Error in is_string(x): argument "id" is missing, with no default
```

I.e. the error comes from the `is_string()` call in `check_string()` and we never make it to the intended call to `stop_input_type()` and as a result we also don't report the user-facing caller (`gargle_oauth_client()`).

I think I expected `check_string()` to effectively act like `if (missing(x) || !is_string(x)) { ... }`, which is presumably equivalent to @hadley's suggestion re: `allow_missing = FALSE` and yields:

``` r
gargle_oauth_client()
#> Error in `gargle_oauth_client()`:
#> ! `id` must be a single string, not absent.

#> Backtrace:
#>     ▆
[...]

PR #1493, opened 2022-09-28
Replace `what` arguments by `allow_null` and `allow_na`

lionel-:
Part of #1481.


R/compat-obj-type.R
```diff
@@ -207,27 +210,57 @@ obj_type_oo <- function(x) {
...
-    format_arg(arg),
+    cli$format_arg(arg),
     what,
     obj_type_friendly(x)
   )
 
   abort(message, ..., call = call, arg = arg)
 }
 
+chr_enumerate <- function(chr, sep = ", ", final = "or") {
```
hadley, line 245: FWIW I'd call this function `oxford_comma` because that explains precisely what it does.

R/compat-types-check.R
```diff
@@ -53,85 +113,181 @@ is_number <- function(x) {
...
+    call = call
+  )
 }
 
 
 # Vectors -----------------------------------------------------------------
 
-# TODO: Restrict missing and special values
-
 check_character <- function(x,
```
hadley, line 271: Should we remove this since it's the only function that checks for vectors?

R/compat-types-check.R
```diff
@@ -19,32 +24,87 @@
...
+    arg = arg,
+    call = call
+  )
 }
 
 check_string <- function(x,
                          ...,
-                         what = "a single string",
+                         allow_na = FALSE,
+                         allow_null = FALSE,
```
hadley, line 55: Did we want to have `allow_empty` here too?

hadley (approved):

R/compat-types-check.R
lionel-: It's there because I've used it somewhere. I meant to treat this file as append-only storage for check functions we use in our packages. Does this make sense?

hadley: Now that we're expanding the usage of this file, it might make sense to be a bit more deliberative about adding functions here. But if you think removing it (or moving to a different file if only rlang uses it), then we can keep it here.

lionel-: I think we could keep it there as a template and basis of discussion for vector checks.

R/compat-obj-type.R
```diff
@@ -207,27 +210,57 @@ obj_type_oo <- function(x) {
 stop_input_type <- function(x,
                             what,
```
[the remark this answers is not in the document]
DavisVaughan: Something about being an enumeration of allowed options

R/compat-types-check.R
```diff
@@ -19,32 +24,90 @@
...
+    allow_null = allow_null,
+    arg = arg,
+    call = call
+  )
 }
 
 check_string <- function(x,
                          ...,
-                         what = "a single string",
+                         allow_empty = TRUE,
```
DavisVaughan, line 54: I vote for a `FALSE` default!

R/session.R
```diff
@@ -306,7 +306,7 @@ check_pkg_version <- function(pkg,
 check_action <- function(action, call = caller_env()) {
   # Take `pkg`, `version`, and `compare`?
   if (!is_null(action)) {
-    check_closure(action, what = "`NULL` or a function", call = call)
+    check_closure(action, call = call, allow_null = TRUE)
```
DavisVaughan, line 309: Nice

tests/testthat/_snaps/compat-types-check.md
```diff
@@ -7,17 +7,23 @@
       Error in `checker()`:
       ! `foo` must be `TRUE` or `FALSE`, not `NA`.
     Code
-      err(checker(lgl(), check_bool))
+      err(checker(NULL, check_bool))
     Output
       <error/rlang_error>
       Error in `checker()`:
-      ! `foo` must be `TRUE` or `FALSE`, not an empty logical vector.
+      ! `foo` must be `TRUE` or `FALSE`, not NULL.
```
DavisVaughan, line 14: Weird that `NULL` isn't backticked here. Possible issue in `obj_type_friendly()`?

DavisVaughan (approved):
Looks great!

R/compat-types-check.R
```diff
@@ -19,32 +24,90 @@
...
                        ...,
-                       what = "`TRUE` or `FALSE`",
+                       allow_na = FALSE,
+                       allow_null = FALSE,
                        arg = caller_arg(x),
                        call = caller_env()) {
-  if (!is_bool(x)) {
-    stop_input_type(x, what, ..., arg = arg, call = call)
+  if (is_bool(x)) {
+    return(invisible(NULL))
```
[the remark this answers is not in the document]
lionel-: I'm not sure. It seems safer to use a more restricted output type for now?

R/compat-types-check.R
lionel-: Sounds good.

tests/testthat/_snaps/compat-types-check.md
lionel-: Good point.

PR #1494, opened 2022-09-28
Improve absent argument errors in input checkers

lionel-:
Branched from #1493.
Part of #1481.


hadley (approved):
I really like this change — it not only tells that the argument is missing, but what you should be supplying.

DavisVaughan (approved):

PR #1495, opened 2022-09-28
Add `allow_decimal` and `allow_infinite`

### LINKED ISSUE #1481: Consider `check_` functions

lionel-:
Branched from #1494.
Closes #1481.


R/compat-obj-type.R
```diff
@@ -74,7 +75,14 @@ obj_type_friendly <- function(x, value = TRUE, length = FALSE) {
         typeof(x),
         logical = if (x) "`TRUE`" else "`FALSE`",
         integer = "an integer",
-        double = "a number",
+        double =
+          if (abs(x) == Inf) {
```
hadley, line 79: `is.infinite(x)`?

R/compat-obj-type.R
```diff
@@ -74,7 +75,14 @@ obj_type_friendly <- function(x, value = TRUE, length = FALSE) {
...
+            "an infinite number"
+          } else if (is.nan(x)) {
+            "`NaN`"
```
hadley, line 82: I think you need a test for this because I'm not sure it works — `is.na(NaN)` is `TRUE` so if I'm reading the code correctly it'll be caught by the previous `if` block.

R/compat-types-check.R
```diff
@@ -85,12 +86,19 @@ check_string <- function(x,
 
 check_number <- function(x,
                          ...,
+                         allow_decimal = FALSE,
```
hadley, line 89: Should we make this true by default? I think most people would say that 10.5 is a number.

hadley (approved):

R/compat-types-check.R
lionel-: The thing is that, at least in the sort of packages we work on, it's much more common to take integer inputs than real numbers (e.g. when we take a length or size, a number of rows, an iteration number, etc). So I think we need a function with the right default for these cases. We could export another function for this but the name would be longer: `check_round_number()`.

R/compat-obj-type.R
lionel-: Good catch.

R/compat-types-check.R
hadley: `check_number_decimal()` + `check_number_whole()` might work? Then autocomplete would remind you of the options, and the longer names aren't that important since the call will usually be short.

lionel-: I've split the functions in the two variants that you suggested. `_decimal` allows infinite values by default, and `_whole` hard-codes `allow_infinite` to `FALSE`.

R/compat-types-check.R
```diff
@@ -103,9 +148,15 @@ check_number <- function(x,
     }
   }
 
+  if (allow_decimal) {
+    what <- "a number"
+  } else {
+    what <- "a round number"
```
DavisVaughan, line 154:
```suggestion
    what <- "a whole number"
```

Since the function is `check_number_whole()`. This also sounds a little better to me

R/compat-obj-type.R
```diff
@@ -74,7 +80,12 @@ obj_type_friendly <- function(x, value = TRUE, length = FALSE) {
         typeof(x),
         logical = if (x) "`TRUE`" else "`FALSE`",
         integer = "an integer",
-        double = "a number",
+        double =
+          if (is.infinite(x)) {
+            "an infinite number"
```
DavisVaughan, line 85: I actually like seeing ``"`Inf`"`` and ``"`-Inf`"`` rather than this, if that sounds good to you. Since we also like `NA` and `NaN` out explicitly

tests/testthat/test-friendly-type.R
```diff
@@ -54,3 +54,7 @@ test_that("obj_type_friendly() handles empty vectors", {
   expect_equal(obj_type_friendly(raw()), "an empty raw vector")
   expect_equal(obj_type_friendly(list()), "an empty list")
 })
+
+test_that("obj_type_friendly() handles NaN", {
+  expect_equal(obj_type_friendly(NaN), "`NaN`")
+})
```
DavisVaughan, line 60: Maybe an additional test about Inf/-Inf

DavisVaughan (approved):

PR #1501, opened 2022-10-04
Add `check_name()`

lionel-:
And allow `""` by default in `check_string()`


tests/testthat/_snaps/compat-types-check.md
```diff
@@ -88,6 +88,59 @@
...
       ! `foo` must be a single string, not a number.
 
+# `check_name()` checks
+
+    Code
+      err(checker("", check_name))
+    Output
+      <error/rlang_error>
+      Error in `checker()`:
+      ! `foo` must be a single string, not `""`.
```
hadley, line 98: Would be nice if this could be `must be a valid variable name`

hadley (approved):

tests/testthat/_snaps/compat-types-check.md
lionel-: I was worried that referring to "name" might be a bit too restricting. "variable name" feels especially too specific to me, but we could try `must be a valid name"`?

hadley: "Valid name" feels slightly too unspecified to me, but I can't think of anything better.

PR #1502, opened 2022-10-04
Add `min` and `max` to number checkers

lionel-:
- Add `min` and `max` to `check_number_whole()` and its decimal variant.

- Refactor `obj_type_friendly()` to more consistently show values. This is the largest part of this PR, with many new tests, and some functionality extracted in a separate `vec_type_friendly()` helper.

The goal of the second change is to consistently show values in error messages when a range is requested, so that we don't get messages like "must be a number greater than 0, not a number".

```r
foo <- function(x, min = -Inf, max = Inf) {
  check_number_whole(x, min = min, max = max)
  x
}

foo(1, max = 0)
#> Error in `foo()`:
#> ! `x` must be a number smaller than 0, not 1.

foo(-1, min = 0)
#> Error in `foo()`:
#> ! `x` must be a number larger than 0, not -1.

foo(10, min = 1, max = 5)
#> Error in `foo()`:
#> ! `x` must be a number between 1 and 5, not 10.
```

### DIFF
--- a/R/compat-obj-type.R
+++ b/R/compat-obj-type.R
@@ -40,12 +40,13 @@
 
 #' Return English-friendly type
 #' @param x Any R object.
-#' @param value Whether to describe the value of `x`.
+#' @param value Whether to describe the value of `x`. Special values
+#'   like `NA` or `""` are always described
 #' @param length Whether to mention the length of vectors and lists.
 #' @return A string describing the type. Starts with an indefinite
 #'   article, e.g. "an integer vector".
 #' @noRd
-obj_type_friendly <- function(x, value = TRUE, length = FALSE) {
+obj_type_friendly <- function(x, value = TRUE) {
   if (is_missing(x)) {
     return("absent")
   }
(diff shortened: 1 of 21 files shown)


tests/testthat/_snaps/trace.md
```diff
@@ -1282,7 +1282,7 @@
     Output
       <error/rlang_error>
       Error in <<CALL `h()`>>:
-      ! `x` must be a single string, not a number.
+      ! `x` must be a single string, not 1.
```
lionel-, line 1285: Maybe that should be:

> not the number 1.

R/compat-types-check.R
```diff
@@ -127,13 +131,17 @@ check_number_decimal <- function(x,
 
 check_number_whole <- function(x,
                                ...,
+                               min = -Inf,
+                               max = Inf,
```
hadley, line 135: Could you pass `allow_infinite` here too? I know that technically Inf is not an integer, but it's often use to have an interface that usually requires integers, but uses `Inf` to mean an "unbounded number".

tests/testthat/_snaps/trace.md
hadley: Yeah, I like that.

tests/testthat/_snaps/compat-types-check.md
```diff
@@ -292,7 +310,7 @@
     Output
       <error/rlang_error>
       Error in `checker()`:
-      ! `foo` must be a symbol, not a string.
+      ! `foo` must be a symbol, not "foo".
```
hadley, line 313: If we're now inlining strings, do we need to worry about very long strings?

hadley (approved):

tests/testthat/_snaps/compat-types-check.md
lionel-: hmm good point. We should also worry about strings with newlines.

hadley: Maybe truncate at 30 chars and apply `encodeString()`?

R/compat-obj-type.R
```diff
@@ -59,44 +66,62 @@ obj_type_friendly <- function(x, value = TRUE, length = FALSE) {
...
+          "`Inf`"
+        } else {
+          "`-Inf`"
+        }
+      }
+
+      if (value) {
+        if (is.numeric(x) || is.complex(x)) {
+          if (is.infinite(x)) {
+            return(show_infinites(x))
```
DavisVaughan, line 99: You might need a special path for complex because you can't compare them

``` r
complex(real = 1, imaginary = Inf) > 0
#> Error in complex(real = 1, imaginary = Inf) > 0: invalid comparison with complex values
```

R/compat-obj-type.R
```diff
@@ -34,12 +40,13 @@
 
 #' Return English-friendly type
 #' @param x Any R object.
-#' @param value Whether to describe the value of `x`.
+#' @param value Whether to describe the value of `x`. Special values
+#'   like `NA` or `""` are always described
```
DavisVaughan, line 44:
```suggestion
#'   like `NA` or `""` are always described.
```

R/compat-types-check.R
DavisVaughan: For example, I think `n` in `slice_max()` is like that - which admittedly is probably a little weird, because why are you even slicing at that point

src/internal/cnd.c
```diff
@@ -202,12 +202,12 @@ r_obj* new_condition_names(r_obj* data) {
...
 
-const char* rlang_obj_type_friendly_full(r_obj* x, bool value, bool length) {
-  r_obj* out_obj = KEEP(r_eval_with_xyz(obj_type_friendly_call,
-                                        x,
-                                        r_lgl(value),
-                                        r_lgl(length),
-                                        rlang_ns_env));
+// `length` is no longer a valid argument
+const char* rlang_obj_type_friendly_full(r_obj* x, bool value, bool _length) {
+  r_obj* out_obj = KEEP(r_eval_with_xy(obj_type_friendly_call,
```
DavisVaughan, line 207: How do we manage deprecation here? I guess all downstream users should update, then at some point in the future we can remove this?

tests/testthat/_snaps/trace.md
DavisVaughan: Similar with `not the string "foo"`? Or is that different?

tests/testthat/_snaps/compat-types-check.md
```diff
@@ -95,7 +95,7 @@
     Output
       <error/rlang_error>
       Error in `checker()`:
-      ! `foo` must be a single string, not `""`.
+      ! `foo` must be a single string, not "".
```
DavisVaughan, line 98: I guess we aren't going to backtick strings?

R/compat-types-check.R
```diff
@@ -145,21 +153,53 @@ check_number_whole <- function(x,
...
+
   if (!missing(x)) {
     is_number <- is_number(
       x,
       allow_decimal = allow_decimal,
       allow_infinite = allow_infinite
     )
+
     if (is_number) {
+      if (min > -Inf && max < Inf) {
```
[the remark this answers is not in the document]
lionel-: oh duh. This needs to fall through.

src/internal/cnd.c
lionel-: We first need to add a new C callable, but I didn't bother here. This C callable is still valid. When we add a new one, we just forget this one for a few year and then remove it if we ever come across it again.

tests/testthat/_snaps/compat-types-check.md
```diff
@@ -86,7 +86,7 @@
     Output
       <error/rlang_error>
       Error in `checker()`:
-      ! `foo` must be a single string, not a number.
+      ! `foo` must be a single string, not 1.
```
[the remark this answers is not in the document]
lionel-: I think we don't need to quote numbers in prose.

tests/testthat/_snaps/compat-types-check.md
lionel-: Double-quoting seems unnecessary.

R/compat-types-check.R
```diff
@@ -108,6 +108,8 @@ check_name <- function(x,
 
 check_number_decimal <- function(x,
                                  ...,
+                                 min = -Inf,
+                                 max = Inf,
```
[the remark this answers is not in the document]
lionel-: I think having an expressive default when possible is always better, since it makes the signature intelligible without the documentation?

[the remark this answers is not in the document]
DavisVaughan: Fair enough!

PR #1507, opened 2022-10-07
Detect non-numeric types in number checkers

lionel-:


R/compat-types-check.R
```diff
@@ -250,6 +255,9 @@ is_number <- function(x,
   if (!typeof(x) %in% c("integer", "double")) {
     return(FALSE)
   }
+  if (!is.numeric(x)) {
```
lionel-, line 258: This normally implies that the storage type is int/dbl but better double-check in case of improperly implemented methods.

hadley (approved):
I can't tell if we should be even stricter for these simple scalar checkers. Should we always require bare vectors?

DavisVaughan (approved):
I wouldn't mind being super strict here either. i.e. only bare integer/double

lionel-:
I'd prefer not being stricter here because:

- `is.numeric()` only returns `TRUE` for unproxied types (or at least it should, it's documented this way). So we don't need to make any more assumptions about the data (e.g. `.subset2()` must return a meaningful number).

- With vctrs we are making it easier to create classed vectors. Forcing users or third-party developers to worry about `x` being classed when e.g. passing `n = x[[1]]` seems to go against this objective.

hadley:
I’m not sure point two is correct because I think the scalar type system that we are creating with these functions is largely orthogonal to vctrs. And I don’t think there are many cases that you’ll create a scalar from a single element of a vector. We should probably talk this through and think about the connections to S3/R7.

PR #1536, opened 2023-01-09
Remove unused extra argument in `ffi_new_condition`

georgestagg:
In `src/internal/init.c`, the function `ffi_new_condition` is set in the `r_callables` table to take 3 arguments, but the implementation takes 4 arguments. This breaks rlang when running under WebAssembly, where function pointer signatures must be consistent.

It doesn't look like the `error_call` argument is used, so I've removed it to leave the first 3 arguments.


lionel-:
Thanks!

PR #1542, opened 2023-01-18
Optimize `r_dyn_*_push_back()` for vector types

DavisVaughan:
The push-back operation of our dynamic vector is probably one of the most commonly used ops in practice, so it would be nice if this was really fast. Currently it calls into `r_dyn_push_back()`, which is generic (which is great!) but slow because of its generality. 

I was looking into replacing our vctrs `growable` type with dynamic vectors and was surprised to see just how slow our dynamic push-back was in comparison.

Consider this simple example of doing repeated push-backs and no resizing.

```r
r_obj* ffi_test_push_back() {
  r_ssize size = 100000;

  struct r_dyn_array* p_vec = r_new_dyn_vector(R_TYPE_integer, size);
  KEEP(p_vec->shelter);

  for (r_ssize i = 0; i < size; ++i) {
    r_dyn_int_push_back(p_vec, i);
  }

  r_obj* out = r_dyn_unwrap(p_vec);

  FREE(1);
  return out;
}
```

```r
# CRAN rlang
bench::mark(.Call(ffi_test_push_back), iterations = 10000)
#> # A tibble: 1 × 6
#>   expression                     min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 .Call(ffi_test_push_back)    503µs    610µs     1609.     391KB     11.7

# This PR
bench::mark(.Call(ffi_test_push_back), iterations = 10000)
#> # A tibble: 1 × 6
#>   expression                     min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 .Call(ffi_test_push_back)    113µs    165µs     5699.     391KB     41.3
```

Providing "native" push-back operators is a big win here, so it definitely seems worth it to me.

### DIFF
--- a/src/rlang/dyn-array.h
+++ b/src/rlang/dyn-array.h
@@ -73,36 +73,6 @@ const void* r_dyn_cend(struct r_dyn_array* p_arr) {
   return r_dyn_cpointer(p_arr, p_arr->count);
 }
 
-static inline
-void* const * r_dyn_pop_back(struct r_dyn_array* p_arr) {
-  void* const * out = (void* const *) r_dyn_clast(p_arr);
-  --p_arr->count;
-  return out;
-}
-
-static inline
-void r_dyn_lgl_push_back(struct r_dyn_array* p_vec, int elt) {
-  r_dyn_push_back(p_vec, &elt);
-}
-static inline
-void r_dyn_int_push_back(struct r_dyn_array* p_vec, int elt) {
-  r_dyn_push_back(p_vec, &elt);
-}
-static inline
-void r_dyn_dbl_push_back(struct r_dyn_array* p_vec, double elt) {
-  r_dyn_push_back(p_vec, &elt);
-}
-static inline
-void r_dyn_cpl_push_back(struct r_dyn_array* p_vec, r_complex elt) {
-  r_dyn_push_back(p_vec, &elt);
-}
-static inline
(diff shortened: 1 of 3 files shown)


src/rlang/dyn-array.h
```diff
@@ -164,4 +134,51 @@ void r_dyn_list_poke(struct r_dyn_array* p_vec, r_ssize i, r_obj* value) {
...
+r_ssize r__dyn_increment(struct r_dyn_array* p_arr) {
+  r_ssize loc = p_arr->count++;
+
+  if (p_arr->count > p_arr->capacity) {
+    r_ssize new_capacity = r_ssize_mult(p_arr->capacity, p_arr->growth_factor);
+    r_dyn_resize(p_arr, new_capacity);
+  }
+
+  return loc;
+}
```
DavisVaughan, line 154: If you don't like this name, I am not tied to it. I can't think of anything better for "increment the count, resize if necessary to make sure you can actually put something there, and return the location"

lionel- (approved):
Nice!

By the way, are your benchmarks with `-O2/3`?

DavisVaughan:
`-O2` and on an Intel mac

PR #1543, opened 2023-01-20
Use snprintf()

hadley:


src/internal/hash.c
```diff
@@ -180,7 +180,7 @@ r_obj* hash_value(XXH3_state_t* p_xx_state) {
   // 32 for hash, 1 for terminating null added by `sprintf()`
```
DavisVaughan, line 180:
```suggestion
  // 32 for hash, 1 for terminating null added by `snprintf()`
```

DavisVaughan (approved):
LGTM. I verified that `snprintf()` also adds the null terminator. It also assumes that the size you give it _includes_ the null terminator, so `sizeof(out)` should be correct. https://cplusplus.com/reference/cstdio/snprintf/

PR #1548, opened 2023-01-30
Update `compat-` files to new naming scheme

hadley:


lionel- (approved):

jennybc (commented):
This renaming needs to be tweaked for consistency (remove the `import-`):

```
R/compat-defer.R → R/import-standalone-defer.R
```

hadley:
@jennybc doesn’t that file come from withr?

jennybc:
OK I just did what I thought needed to be done. See what you think.

jennybc:
The name of the file now matches what it claims its name to be.

hadley:
Isn't https://github.com/r-lib/withr/blob/main/R/compat-defer.R the source of this content? Do I misunderstand?

lionel-:
yup it is, defer is an import in rlang from withr.

jennybc:
I see a file here named `import-standalone-defer.R`. **Note the `import-` prefix.**

I don't think any other standalone file is named that way? Nor is the original file in withr named that way.
And inside that file, it claims to come from an original name `standalone-defer.R`, also without `import-` prefix.

But I seem to be confused about what's going on here and will revert my commit.

lionel-:
Source files are prefixed with `standalone-`, and imported files are prefixed with `import-standalone-`, this way you can tell by the name if a file is an original or a copycat.

DavisVaughan (approved):

PR #1553, opened 2023-02-08
Optimize `new_data_mask()` and `r_alloc_environment()`

DavisVaughan:
This PR was motivated by this dplyr issue https://github.com/tidyverse/dplyr/issues/6666

If we are going to create a fresh data mask at each group evaluation, then mask creation needs to be very fast.

This PR helps in two ways:
- Reduce the initial data mask size from 100 down to 10. I don't think that we really need such a large initial mask size, since most of the time we don't use `<-` in `eval_tidy()` calls to poke into the mask. I think most of the time the expressions we pass to `eval_tidy()` are function calls like `eval_tidy(mean(x), mask)` or something similar, so I don't think the mask env needs to be this big.
- Optimize `r_alloc_environment()` to use `R_NewEnv()` on R >=4.1.0, which is a C level version of `new.env()`.

``` r
library(rlang)

env <- new_environment()

# CRAN rlang
bench::mark(new_data_mask(env), iterations = 1000000)
#> # A tibble: 1 × 6
#>   expression              min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>         <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 new_data_mask(env)   2.08µs   2.94µs   299002.    3.48KB     18.2

# Just 100->10
bench::mark(new_data_mask(env), iterations = 1000000)
#> # A tibble: 1 × 6
#>   expression              min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>         <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 new_data_mask(env)      2µs   2.75µs   337047.     4.3KB     20.2

# Both 100->10, and `R_NewEnv()`
bench::mark(new_data_mask(env), iterations = 1000000)
#> # A tibble: 1 × 6
#>   expression              min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>         <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 new_data_mask(env)    980ns   1.39µs   646986.     4.3KB     14.9
```

I am hopeful that this means C calls to `new_data_mask()` in dplyr will be in the nanosecond range

We can probably also use this in vctrs if we wanted

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # rlang (development version)
 
+* `new_data_mask()` is now slightly faster due to a smaller initial mask size
+  and usage of the C level function `R_NewEnv()` on R >=4.1.0 (#1553).
+
 * The C level `r_dyn_*_push_back()` utilities are now faster (#1542).
 
 * rlang is now compliant with `-Wstrict-prototypes` as requested by CRAN
--- a/src/internal/eval-tidy.c
+++ b/src/internal/eval-tidy.c
@@ -160,14 +160,14 @@ r_obj* ffi_new_data_mask(r_obj* bottom, r_obj* top) {
   r_obj* data_mask;
 
   if (bottom == r_null) {
-    bottom = KEEP(r_alloc_environment(100, r_envs.empty));
+    bottom = KEEP(r_alloc_environment(10, r_envs.empty));
     data_mask = bottom;
   } else {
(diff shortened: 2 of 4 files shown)


lionel- (approved):
Cool!

PR #1557, opened 2023-02-10
Support warning backtraces in Rmd documents

### LINKED ISSUE #1551: `sew()` method for warnings and messages
https://github.com/r-lib/rlang/blob/90ef6d1af8256ee0a13f95fad48d8c8fd489e7db/R/cnd-message.R#L309-L326

lionel-:
- Special-case calls to `global_entrace()` inside Rmd documents to register calling handlers with the new `knitr::opts_chunk$set(calling.handlers = )` added in https://github.com/yihui/knitr/pull/2079. 

  cc @jennybc, with this you only need the following to entrace base errors in https://github.com/tidyverse/reprex/issues/230:

  ````md
  ```{r setup}
  rlang::global_entrace()
  options(rlang_backtrace_on_error_report = "full"
  ````

- Resignal warnings enriched by `rlang::entrace()`. As discussed on Slack, this may cause bad interactions in extremely specific cases (a warning is resignalled from a warning calling handler without a corresponding restart - we'll suppress the outside warning in that case).

  Currently, we only support displaying the backtraces via `rlang::last_warnings()` and so we haven't needed resignalling the enriched warnings. We now resignal them in order to support the new `rlang_backtrace_on_warning_report` option (see below). Once (or if) we become in control of warning display in R, resignalling will also be useful to automatically display backtraces on warning via a `rlang_backtrace_on_warning` option.

- Add `rlang_backtrace_on_warning_report` global option with an accompanying `sew()` method. Closes #1551. In conjunction with `global_entrace()`, this allows displaying backtraces for all warnings emitted in an Rmd document.


lionel-:
Last commit fixes `last_warnings()` support.

@jennybc this might be useful to include in reprex as well to get warning backtraces?

lionel-:
> So you're saying I should add rlang_backtrace_on_warning_report = "full" to the options being set here (?):

Exactly.

> And perhaps add rlang::global_entrace() before the options() call?

Yes this is necessary for the global option above to have any effect. Also this will enable backtraces for base errors, if I'm not mistaken they're currently only shown for rlang errors.

lionel-:
@hadley now rebased

R/cnd-entrace.R
```diff
@@ -22,6 +22,15 @@
 #' @param class A character vector of one or several classes of
 #'   conditions to be entraced.
 #'
+#' @section Inside RMarkdown documents:
+#'
+#' Call `global_entrace()` inside an RMarkdown document to cause
```
hadley, line 27: Mention that this needs to be done in a chunk before the error/warning?

R/cnd-entrace.R
```diff
@@ -22,6 +22,15 @@
...
+#' errors and warnings to be promoted to rlang conditions that include
+#' a backtrace.
+#'
+#' This is useful in conjunction with [`rlang_backtrace_on_error`] and
+#' [`rlang_backtrace_on_warning`].
+#'
 #' @section Under the hood:
 #' On R 4.0 and newer, `global_entrace()` installs a global handler
 #' with `globalCallingHandlers()`. On older R versions, `entrace()` is
```
hadley, line 36: I think it would be useful to include this somewhere in a copy-and-pasteable format:

```R
rlang::global_entrace()
options(
  rlang_backtrace_on_error_report = "full",
  rlang_backtrace_on_warning_report = "full"
)
```

hadley (approved):

R/cnd-abort.R
```diff
@@ -1364,6 +1365,30 @@ call_restore <- function(x, to) {
 #' # stop("foo")
 NULL
 
+#' Display backtrace on warning
+#'
+#' @description
+#' Similarly to [`rlang_backtrace_on_error_report`], you can set
```
hadley, line 1371: Should these actually be documented in the same file?

hadley (approved):

R/cnd-abort.R
lionel-: You're right, I merged this inside the Rmd section of backtrace-on-error.

PR #1560, opened 2023-02-10
Add `check_logical()`

DavisVaughan:
For `if_else()` and `case_when()` to validate the condition inputs

### DIFF
--- a/R/standalone-types-check.R
+++ b/R/standalone-types-check.R
@@ -1,7 +1,7 @@
 # ---
 # repo: r-lib/rlang
 # file: standalone-types-check.R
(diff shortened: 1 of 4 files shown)


R/standalone-types-check.R
```diff
@@ -470,4 +473,28 @@ check_character <- function(x,
   )
 }
 
+check_logical <- function(x,
+                          ...,
+                          allow_null = FALSE,
```
lionel-, line 478: I wonder if the vector variants should also get the same parameters as the scalar variants. Not for this PR, just a general thought.

lionel- (approved):

PR #1563, opened 2023-02-15
Implement number checks in C

lionel-:
I was hoping for more of a speed-up, especially given how much effort it is to write these things in C. I guess the R versions are already quite fast to begin with. Will this be enough to fix the perf issues @DavisVaughan?

```r
bench::mark(
  new = check_number_whole(1),
  old = check_number_whole_old(1),
  iterations = 10000
)
#> # A tibble: 2 × 13
#>   express…¹    min median itr/s…² mem_a…³ gc/se…⁴ n_itr  n_gc total…⁵
#>   <bch:exp> <bch:> <bch:>   <dbl> <bch:b>   <dbl> <int> <dbl> <bch:t>
#> 1 new       1.07µs 1.44µs 675435.      0B     0   10000     0  14.8ms
#> 2 old       3.57µs 3.81µs 254095.      0B    25.4  9999     1  39.4ms

bench::mark(
  new = check_number_decimal(1.5),
  old = check_number_decimal_old(1.5),
  iterations = 10000
)
#> # A tibble: 2 × 13
#>   express…¹    min median itr/s…² mem_a…³ gc/se…⁴ n_itr  n_gc total…⁵
#>   <bch:exp> <bch:> <bch:>   <dbl> <bch:b>   <dbl> <int> <dbl> <bch:t>
#> 1 new       1.19µs 1.64µs 572827.      0B       0 10000     0  17.5ms
#> 2 old       3.03µs 3.53µs 259783.      0B       0 10000     0  38.5ms
```

FYI @hadley


lionel-:
oops benchmarks above are with `-g -O0`, the `-O3` benchmarks are better:

```r
bench::mark(
  new = check_number_whole(1),
  old = check_number_whole_old(1),
  iterations = 100000
)
#> # A tibble: 2 × 13
#>   expression      min   median `itr/sec` mem_al…¹ gc/se…² n_itr  n_gc
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:by>   <dbl> <int> <dbl>
#> 1 new        573.93ns 737.84ns  1238480.       0B    61.9 99995     5
#> 2 old          2.67µs   3.03µs   321286.       0B    51.4 99984    16

bench::mark(
  new = check_number_decimal(1.5),
  old = check_number_decimal_old(1.5),
  iterations = 10000
)
#> # A tibble: 2 × 13
#>   expression      min   median `itr/sec` mem_al…¹ gc/se…² n_itr  n_gc
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:by>   <dbl> <int> <dbl>
#> 1 new        614.91ns 738.07ns  1221700.       0B    61.1 99995     5
#> 2 old          2.34µs   2.71µs   359769.       0B    57.6 99984    16

````

**Comment by DavisVaughan** [comment]
I think it will help in `lead()` and `lag()` for sure. If we didn't make this change then after I add in faster vctrs assertions this would probably show up as one of the larger points of slowness

```r
library(dplyr)

x <- 1:5 + 0L

bench::mark(lead(x), iterations = 100000)

# Main (dev dplyr, dev rlang, dev vctrs)
#> # A tibble: 1 × 6
#> 1 lead(x)      32.1µs   43.2µs    22647.     211KB     15.6

# With this PR and updated standalone file
#> # A tibble: 1 × 6
#> 1 lead(x)      21.6µs   28.7µs    34649.     131KB     15.3

# simulate many groups
bench::mark(for(i in 1:10000) lead(x), iterations = 10)

# Main (dev dplyr, dev rlang, dev vctrs)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 1 × 6
#>   expression                      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                 <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 for (i in 1:10000) lead(x)    350ms    358ms      2.73    2.56MB     20.8

# With this PR and updated standalone file
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 1 × 6
#>   expression                      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                 <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 for (i in 1:10000) lead(x)    226ms    264ms      3.82    2.56MB     19.5
```

DavisVaughan:
@lionel- I think your M1 runs R code much faster than my Intel. So the switch to C seems more noticeable here

```r
x <- 1L

bench::mark(check_number_whole(x), iterations = 500000)

# Main
#> # A tibble: 1 × 6
#>   expression                 min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>            <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 check_number_whole(x)   7.01µs   9.17µs   106014.    2.08KB     17.4

# This PR
#> # A tibble: 1 × 6
#>   expression                 min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>            <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 check_number_whole(x)   1.99µs   2.34µs   402779.        0B     12.1
```

src/internal/init.c
```diff
@@ -195,6 +195,7 @@ static const R_CallMethodDef r_callables[] = {
   {"ffi_set_names",                    (DL_FUNC) &ffi_set_names, 4},
   {"ffi_sexp_iterate",                 (DL_FUNC) &ffi_sexp_iterate, 2},
   {"ffi_squash",                       (DL_FUNC) &ffi_squash, 4},
+  {"ffi_standalone_check_number_1.0.7", (DL_FUNC) &ffi_standalone_check_number, 7},
```
DavisVaughan, line 198: Do you care about the indentation of `(DL_FUNC)` being off?

src/internal/standalone-types-check.c
```diff
@@ -17,3 +27,136 @@ r_obj* ffi_standalone_is_bool(r_obj* x,
...
 }
+
+r_obj* ffi_standalone_check_number(r_obj* x,
+                                   r_obj* allow_decimal,
+                                   r_obj* min,
+                                   r_obj* max,
+                                   r_obj* allow_infinite,
+                                   r_obj* allow_na,
+                                   r_obj* allow_null) {
+  int out = IS_NUMBER_false;
```
DavisVaughan, line 38: Minor: The implicit enum->int casting always weirds me out

src/internal/standalone-types-check.c
```diff
@@ -17,3 +27,136 @@ r_obj* ffi_standalone_is_bool(r_obj* x,
...
+  default:
+    break;
+  }
+
+  return r_int(out);
+}
+
+static
+bool is_numeric(r_obj* x) {
+  if (r_attrib_get(x, r_syms.class_) == r_null) {
```
DavisVaughan, line 78:
```suggestion
  if (r_class(x) == r_null) {
```

?

Also did you just want `r_is_object()`?

src/internal/decl/standalone-types-check-decl.h
```diff
@@ -0,0 +1,15 @@
...
+                                           r_obj* ffi_min,
+                                           r_obj* ffi_max,
+                                           r_obj* allow_na,
+                                           r_obj* allow_null);
+
+static
+enum is_number dbl_standalone_check_number(r_obj* x,
+                                           r_obj* allow_decimal,
+                                           r_obj* min,
+                                           r_obj* max,
```
DavisVaughan, line 12:
```suggestion
                                           r_obj* ffi_min,
                                           r_obj* ffi_max,
```

src/internal/standalone-types-check.c
```diff
@@ -17,3 +27,136 @@ r_obj* ffi_standalone_is_bool(r_obj* x,
...
+    return IS_NUMBER_false;
+  }
+  if (!is_numeric(x)) {
+    return IS_NUMBER_false;
+  }
+
+  double value = r_dbl_get(x, 0);
+
+  // `value == r_globals.na_dbl` is always false since it involves NaN
+  if (memcmp(&value, &r_globals.na_dbl, sizeof(double)) == 0) {
```
DavisVaughan, line 138: Do you want `R_IsNA()`?

https://github.com/wch/r-source/blob/a428ba5d44a019b048482afd550c1ea2fc566710/src/include/R_ext/Arith.h#L61-L62

DavisVaughan: You could move this down into the branch of `isnan(value)` to differentiate between `NA_real_` and `NaN` (like, if `isnan(value)` is true, then differentiate between the two types)

If we expect the 99% case to be an actual number and _not_ a non-finite value then that might be worth it

src/rlang/c-utils.h
```diff
@@ -23,6 +23,25 @@
 void* r_shelter_deref(r_obj* x);
 
 
+// Allow integers up to 2^52, same as R_XLEN_T_MAX when long vector
+// support is enabled
+#define RLANG_MAX_DOUBLE_INT 4503599627370496
+
+static inline
+bool r_dbl_is_decimal(double x) {
```
DavisVaughan, line 31: I think this name is wrong? Is this `r_dbl_is_whole()`?

i.e. it returns `true` if the double value is "kind of like an integer"

`r_dbl_is_decimal()` implies that `r_dbl_is_decimal(1.5)` returns `true`, i think

src/internal/standalone-types-check.c
```diff
@@ -17,3 +27,136 @@ r_obj* ffi_standalone_is_bool(r_obj* x,
...
+    if (isnan(value)) {
+      return IS_NUMBER_false;
+    } else {
+      return r_as_bool(allow_infinite) ? IS_NUMBER_true : IS_NUMBER_false;
+    }
+  }
+
+  if (!r_as_bool(allow_decimal) && !r_dbl_is_decimal(value)) {
+    return IS_NUMBER_false;
+  }
```
DavisVaughan, line 152: This check confused me because I think `r_dbl_is_decimal()` isn't the right name

src/internal/standalone-types-check.c
```diff
@@ -17,3 +27,136 @@ r_obj* ffi_standalone_is_bool(r_obj* x,
...
+
+  double min = r_arg_as_double(ffi_min, "min");
+  double max = r_arg_as_double(ffi_max, "max");
```
DavisVaughan, line 155: I guess we rely on these not being missing values. Seems reasonable because that would be a programmer mistake

DavisVaughan: Oh this is more interesting to me now because it would have errored in the past but you changed `r_arg_as_double()` in this PR to also allow non-finite values

DavisVaughan (approved):

src/internal/init.c
lionel-: yes but I didn't feel like reindenting everything in this PR

src/internal/standalone-types-check.c
lionel-: oh yeah good point I forgot about is-object

src/rlang/c-utils.h
lionel-: oops

src/internal/standalone-types-check.c
lionel-: Good point

src/internal/standalone-types-check.c
lionel-: Yes so `NA` is disallowed but `NaN` is allowed, which seems to makes sense: we allow all valid C values but disallow the special R value `na_dbl`.

It does mean that passing `min = NaN` should cause an error and it doesn't currently, I'll add some tests.

hadley:
@DavisVaughan GET A NEW LAPTOP

PR #1566, opened 2023-02-16
Support wrapping `englue()`

### LINKED ISSUE #1565: Support wrapping `englue()`
By taking `env` and `error_call` arguments.

lionel-:
Closes #1565
cc @hadley

### DIFF
--- a/R/nse-inject.R
+++ b/R/nse-inject.R
@@ -390,21 +390,30 @@ NULL
 #'
 #' @section Wrapping `englue()`:
 #'
-#' You can provide englue semantics to a string by supplying `env` and
-#' `error_call`. In this example we create a variant of `englue()`
-#' that supports a special `.qux` pronoun:
+#' You can provide englue semantics to a string by supplying `env`.
+#' In this example we create a variant of `englue()` that supports a
+#' special `.qux` pronoun by:
+#'
+#' - Creating an environment `masked_env` that inherits from the user
+#'   env, the one where their data lives.
+#'
+#' - Overriding the `error_arg` and `error_call` arguments to point to
+#'   our own argument name and call environment. This pattern is
+#'   slightly different from usual error context passing because
+#'   `englue()` is a backend function that uses its own error context
+#'   by default (and not a checking function that uses _your_ error
+#'   context by default).
 #'
 #' ```{r}
 #' my_englue <- function(text) {
-#'   # Create an environment that inherits from the user env, where
-#'   # their data lives:
-#'   caller <- caller_env()
-#'   env <- env(caller, .qux = "QUX")
+#'   masked_env <- env(caller_env(), .qux = "QUX")
 #'
-#'   # Override `error_call` by passing the caller frame. It normally
-#'   # defaults to `env` but here that wouldn't be appropriate because
-#'   # it is not an execution enviromment of a running function.
-#'   englue(text, env = env, error_call = caller)
+#'   englue(
+#'     text,
+#'     env = masked_env,
+#'     error_arg = "text",
+#'     error_call = current_env()
+#'   )
 #' }
 #'
 #' # Users can then use your wrapper as they would use `englue()`:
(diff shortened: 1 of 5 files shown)


R/nse-inject.R
```diff
@@ -368,6 +368,9 @@ NULL
 #' string using the default name operation.
 #'
 #' @param x A string to interpolate with glue operators.
+#' @param env User environment where the interpolation data lives in
+#'   case you're wrapping `englue()` in another function.
+#' @inheritParams args_error_context
 #'
```
DavisVaughan, line 374: You have a section in details like

> Before using `englue()` in a package

It might also be useful to mention that `env` should be passed through there if the usage is exposed to the user somehow. It also comes up when wrapping glue itself https://github.com/tidyverse/glue/issues/281

hadley:
Nice!

R/nse-inject.R
DavisVaughan: oh i see

```
@section Wrapping `englue()`:
```

now. Never mind

R/nse-inject.R
```diff
@@ -385,6 +388,43 @@ NULL
 #' usethis::use_package("glue", "Imports")
 #' ```
 #'
+#' @section Wrapping `englue()`:
+#'
+#' You can provide englue semantics to a string by supplying `env`.
```
DavisVaughan, line 393:
```suggestion
#' You can provide englue semantics to a user provided string by supplying `env`.
```

?

R/nse-inject.R
```diff
@@ -385,6 +388,43 @@ NULL
...
+#'
+#' - Overriding the `error_arg` and `error_call` arguments to point to
+#'   our own argument name and call environment. This pattern is
+#'   slightly different from usual error context passing because
+#'   `englue()` is a backend function that uses its own error context
+#'   by default (and not a checking function that uses _your_ error
+#'   context by default).
+#'
+#' ```{r}
+#' my_englue <- function(text) {
```
DavisVaughan, line 408: I think `my_englue()` should also expose `env`.

I think this is how we will end up documenting `.envir` in glue, like https://github.com/tidyverse/glue/issues/281#issuecomment-1363096413

R/nse-inject.R
```diff
@@ -400,55 +440,62 @@ NULL
 #' as_label(letters)
 #'
 #' @export
-englue <- function(x) {
-  check_string(x)
+englue <- function(x,
+                   env = caller_env(),
+                   error_call = current_env(),
```
DavisVaughan, line 445: I don't have strong opinions, but do you have a reason for using `current_env()` over `caller_env()`?

`englue()` will pretty much always be wrapped inside another function right? Should it default to reporting that function as the location of any issues?

R/nse-inject.R
DavisVaughan: I guess the main point is ensuring that `caller_env()` gets used and passed through somehow, so it isn't that important... but it still might be "best practice"?

Or maybe you can say:

> If you are creating a low level package on top of `englue()`, you should also consider exposing `env`, `error_arg` and `error_call` in your `englue()` wrapper so users can wrap your wrapper.

R/nse-inject.R
```diff
@@ -400,55 +440,62 @@ NULL
...
+                   error_arg = "x") {
```
DavisVaughan, line 446: Also mildly surprised this isn't `error_arg(x)`

R/nse-inject.R
```diff
@@ -400,55 +440,62 @@ NULL
...
+glue_first_pass_eval <- function(text, env) {
   text_expr <- parse_expr(text)
   defused_expr <- eval_bare(call2(enexpr, text_expr), env)
 
-  if (is_symbol(text_expr)) {
+  if (is_symbol(text_expr) && is_missing(defused_expr)) {
+    error_arg <- as_string(text_expr)
+    error_call <- NULL
+
+    # Find the relevant error frame. There are edge cases where this
```
DavisVaughan, line 485: This is the only part im a little iffy on but it seems reasonable and the downsides seem slim

DavisVaughan (approved):

R/nse-inject.R
lionel-: I like your note about exposing these arguments to allow wrappers.

There are cases where this would be unwieldy though, e.g. in the tidyr case.

By the way, maybe we should consider an `enquo()`-based interface for englue? We'd require literal strings as inputs and to wrap englue you'd use `{{`.

lionel-: This would have the advantage of making it easy to wrap `names_glue` if the tidyr function adds englue support, without having to expose `env`.

Doesn't solve the error-call issue though.

DavisVaughan: The idea being that the quosure created through the internal `enquo()`-like call would capture and pass on the `env` for you, right? Seems interesting.

I guess that quosure env becomes the parent env if you want to insert a `.qux` pronoun or something

lionel-: yup we'd have to take `data` instead of `env` in this case. And if multiple layer masks are ever needed, we could extend to accept rlang data masks which make the layer structure explicit so we'd have all the information needed (even though we wouldn't use `eval_tidy()` but `glue()` in the end, I think that should still work).

lionel-: But the compiler would get us in trouble though, since it unwraps strings from promises and then we lose the environment. We can write this idea off.

R/nse-inject.R
lionel-: Most of the time `englue()` will be used as part of tidy eval to program against dplyr or tidyr. In these cases, I think it makes sense to match input errors to the functions that the user used, which is `englue()`.

R/nse-inject.R
lionel-: In most cases (tidyeval prog), `caller_arg(x)` will be a string instead of a symbol.

PR #1568, opened 2023-02-17
Add `.ignore_null` argument for `enquos()` and friends

### LINKED ISSUE #1450: Should `enquos()` get `ignore_nulls` option?
Since we want to do that in so many functions?

lionel-:
Closes #1450

### DIFF
--- a/R/nse-defuse.R
+++ b/R/nse-defuse.R
@@ -117,6 +117,8 @@ expr <- function(expr) {
 #'   of `"trailing"`, `"none"`, `"all"`. If `"trailing"`, only the
 #'   last argument is ignored if it is empty. Named arguments are not
 #'   considered empty.
+#' @param .ignore_null Whether to ignore unnamed null arguments. Can be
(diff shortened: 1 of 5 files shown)


R/nse-defuse.R
```diff
@@ -117,6 +117,8 @@ expr <- function(expr) {
 #'   of `"trailing"`, `"none"`, `"all"`. If `"trailing"`, only the
 #'   last argument is ignored if it is empty. Named arguments are not
 #'   considered empty.
+#' @param .ignore_null Whether to ignore null arguments. Can be
```
hadley, line 120: Would this be clearer if it was "unnamed" or "none" ?

hadley (approved):

R/nse-defuse.R
lionel-: I think so. However it would incorrectly imply that `.ignore_empty = "all"` behaves differently than `.ignore_null = "unnamed"`.

PR #1576, opened 2023-02-23
Deprecate flatten and squash functions

lionel-:
Branched from #1575.

Deprecate `flatten()`, `squash()`, and their variants in favour of purrr functions.

cc @hadley @DavisVaughan


R/lifecycle-deprecated.R
```diff
@@ -1017,3 +1017,191 @@ child_env <- function(.parent, ...) {
...
+#' [unlist()] but they are type-stable so you always know what the
+#' type of the output is.
+#'
+#'
+#' @section Life cycle:
+#'
+#' These functions are in the questioning stage. They have slightly
+#' different semantics than the flattening functions in purrr and we
+#' are currently rethinking our approach to flattening with the new
+#' typing facilities of the vctrs package.
```
hadley, line 1040:
```suggestion
```

?

hadley (approved):

R/nse-defuse.R
```diff
@@ -491,7 +491,10 @@ endots <- function(call,
   })
 
   if (splice_dots) {
-    dots <- flatten_if(dots, is_spliced)
+    with_options(
+      lifecycle_verbosity = "quiet",
+      dots <- flatten_if(dots, is_spliced)
+    )
```
DavisVaughan, line 497: I'm somewhat worried about the performance impact of this in `enquos()`

Historically we've had performance issues with lifecycle warnings in performance sensitive functions, and `with_options()` also has non negligible overhead in setting up exit handlers

lionel-: On the other hand endots is already quite slow.

lionel-: You're right the overhead is bad, I'll change the implementation to something else.

lionel-: I changed the implementation here: https://github.com/r-lib/rlang/commit/2fa538f8bedb58c82293f82f50e620dedd6aef7e

Before:

```r
fn <- function(a, ..., b) enquos(a, ..., b)

bench::mark(
  fn(foo, bar, b = baz),
  iterations = 10000
)
#> # A tibble: 1 × 13
#>   expression                 min median itr/s…¹ mem_a…² gc/se…³ n_itr
#>   <bch:expr>            <bch:tm> <bch:>   <dbl> <bch:b>   <dbl> <int>
#> 1 fn(foo, bar, b = baz)   20.7µs 22.4µs  38632.      0B    7.73  9998
```

After:

```r
#> # A tibble: 1 × 13
#>   expression                 min median itr/s…¹ mem_a…² gc/se…³ n_itr
#>   <bch:expr>            <bch:tm> <bch:>   <dbl> <bch:b>   <dbl> <int>
#> 1 fn(foo, bar, b = baz)   23.6µs   26µs  36994.      0B    11.1  9997
```

The deprecation put this example in the half-ms range.

DavisVaughan: Yea that checks out, I remember the lifecycle infrastructure being around that slow

And yea even `with_options()` has a chunk of overhead

``` r
bench::mark(with_options(lifecycle_verbosity = "quiet", 1))
#> # A tibble: 1 × 6
#>   expression                                          min  median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                                     <bch:tm> <bch:t>     <dbl> <bch:byt>    <dbl>
#> 1 with_options(lifecycle_verbosity = "quiet", 1)   92.6µs   103µs     9206.      688B     12.7
```

echasnovski:
@lionel- , sorry to bother here.

I've just encountered the soft deprecation of `rlang::squash()` which I heavily use in [ruler](https://github.com/echasnovski/ruler). As far as I've researched, there is no immediate replacement for it in `purrr` (and deprecation note doesn't seem to suggest anything in particular). `purrr::list_flatten()` removes only one level while I need to convert whole nested structure into a single list. The only approach I can see is to repeatedly call `purrr::list_flatten()` to achieve same effect.

Could you, please, recommend where I'd better look to resolve the issue?

DavisVaughan:
I did have to add this helper to dplyr to retain squash-like behavior

```r
#' @param x A list
#' @param fn An optional function of 1 argument to be applied to each list
#'   element of `x`. This allows you to further refine what elements should be
#'   flattened. `fn` should return a single `TRUE` or `FALSE`.
#' @param recursive Should `list_flatten()` be applied recursively? If `TRUE`,
#'   it will continue to apply `list_flatten()` as long as at least one element
#'   of `x` was flattened in the previous iteration.
#' @noRd
list_flatten <- function(x, ..., fn = NULL, recursive = FALSE) {
  check_dots_empty0(...)

  obj_check_list(x)
  x <- unclass(x)

  loc <- map_lgl(x, obj_is_list)

  if (!is_null(fn)) {
    loc[loc] <- map_lgl(x[loc], fn)
  }

  not_loc <- !loc

  names <- names(x)
  if (!is_null(names)) {
    # Always prefer inner names, even if inner elements are actually unnamed.
    # This is what `rlang::flatten_if()` did, with a warning. We could also
    # use `name_spec` and `name_repair` for a more complete solution.
    names[loc] <- ""
    names(x) <- names
  }

  x[loc] <- map(x[loc], unclass)
  x[not_loc] <- map(x[not_loc], list)

  out <- list_unchop(x, ptype = list())

  if (recursive && any(loc)) {
    out <- list_flatten(out, fn = fn, recursive = TRUE)
  }

  out
}
```

echasnovski:
> I did have to add this helper to dplyr to retain squash-like behavior

Yeah, so it seems that writing own helper with iterative `purrr::list_flatten()` application is the way to go. Shame to have a deprecation without immediate alternative :(

hadley:
@DavisVaughan can you add that code to a purrr issue? It feels like purrr is the right home for this sort of thing.

PR #1578, opened 2023-02-24
Turning off condition inheritance

lionel-:
Follow up to #1573 (see discussion there).

New `.inherit = <bool>` argument. This controls whether `cnd_inherits()` and `try_fetch()` are allowed to match parent conditions. The default is `TRUE` in `abort()`. In `warn()` and `inform()`, the default is `TRUE` in the general case, and `FALSE` when `parent` is a condition of higher severity. This prevents matching downgraded conditions unexpectedly.

The parameter is stored in `cnd$rlang$inherit`. This is a public field (otherwise it'd be stored in `rlang$internal$`. It's namespaced to avoid name clashes with subclasses.

cc @DavisVaughan @gaborcsardi

### DIFF
--- a/R/cnd-handlers.R
+++ b/R/cnd-handlers.R
@@ -216,7 +216,6 @@ handler_call <- quote(function(cnd) {
     } else {
       except <- ""
     }
-    inherit <- .subset2(.subset2(cnd, "rlang"), "inherit")
   }
 
   while (!is_null(cnd)) {
@@ -225,6 +224,7 @@ handler_call <- quote(function(cnd) {
       if (!inherits(out, "rlang_zap")) throw(out)
     }
 
+    inherit <- .subset2(.subset2(cnd, "rlang"), "inherit")
     if (is_false(inherit)) {
       return()
     }
--- a/R/cnd-handlers.R
+++ b/R/cnd-handlers.R
@@ -132,7 +132,7 @@ environment(hnd_prompt_install) <- baseenv()
 #' insensitive to changes of implementation or context of evaluation
 #' that cause a classed error to suddenly get chained to a contextual
 #' error. Note that some chained conditions are not inherited, see the
-#' `inherit` argument of [abort()] or [warn()]. In particular,
+#' `.inherit` argument of [abort()] or [warn()]. In particular,
 #' downgraded conditions (e.g. from error to warning or from warning
 #' to message) are not matched across parents.
 #'
--- a/R/cnd-handlers.R
+++ b/R/cnd-handlers.R
@@ -131,8 +131,10 @@ environment(hnd_prompt_install) <- baseenv()
 #' [abort()]. This is a useful property that makes `try_fetch()`
 #' insensitive to changes of implementation or context of evaluation
 #' that cause a classed error to suddenly get chained to a contextual
-#' error. Note however that downgraded conditions (e.g. from error to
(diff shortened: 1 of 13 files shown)


hadley (approved):

R/cnd-handlers.R
```diff
@@ -131,8 +131,10 @@ environment(hnd_prompt_install) <- baseenv()
 #' [abort()]. This is a useful property that makes `try_fetch()`
 #' insensitive to changes of implementation or context of evaluation
 #' that cause a classed error to suddenly get chained to a contextual
-#' error. Note however that downgraded conditions (e.g. from error to
-#' warning or from warning to message) are not matched across parents.
+#' error. Note that some chained conditions are not inherited, see the
+#' `inherit` argument of [abort()] or [warn()]. In particular,
```
DavisVaughan, line 135:
```suggestion
#' `.inherit` argument of [abort()] or [warn()]. In particular,
```

R/cnd-handlers.R
```diff
@@ -214,12 +216,19 @@ handler_call <- quote(function(cnd) {
...
+    if (inherits(cnd, CLASS)) {
       out <- handlers[[I]](cnd)
       if (!inherits(out, "rlang_zap")) throw(out)
     }
+
+    if (is_false(inherit)) {
+      return()
+    }
+
     cnd <- .subset2(cnd, "parent")
```
DavisVaughan, line 232: Should `inherit` be updated after each iteration to the `inherit` slot of next parent?

R/cnd.R
```diff
@@ -287,6 +287,9 @@ cnd_type <- function(cnd) {
 #' object is a particular kind of error or has been caused by such an
 #' error.
 #'
+#' Some chained conditions carry parents that are not inherited. See
+#' the `inherit` argument of [abort()], [warn()], and [inform()].
```
DavisVaughan, line 291:
```suggestion
#' the `.inherit` argument of [abort()], [warn()], and [inform()].
```

R/cnd.R
```diff
@@ -345,6 +348,19 @@ cnd_type <- function(cnd) {
 #' class(cnd$parent)
 #' ```
 #'
+#' Note that `try_fetch()` uses `cnd_inherits()` internally. This
+#' makes it very easy to match a parent condition:
+#'
+#' ```{r, comment = "#>", collapse = TRUE}
```
DavisVaughan, line 354: i thought roxygen2 sets these for us automatically now

DavisVaughan (approved):

R/cnd-handlers.R
lionel-: oh good point

R/cnd.R
```diff
@@ -354,11 +370,17 @@ cnd_inherits <- function(cnd, class) {
...
   while (is_condition(cnd)) {
     if (fn(cnd, ...)) {
       return(TRUE)
     }
 
+    if (is_false(inherit)) {
+      return(FALSE)
+    }
+
     cnd <- cnd[["parent"]]
```
DavisVaughan, line 384: Check recursively here too?

lionel-: https://github.com/r-lib/rlang/commit/69e0a5b5d022047144c657e80c8517d7bf8d32c2

PR #1592, opened 2023-03-08
Pass `allow_na = FALSE`

mgirlich:
I stumbled this when working on #1590. As they don't accept `NA` it makes sense to explicitly pass `allow_na = FALSE` so that you cannot pass it accidentally via the `...` arguments (this happened in the tests).
I thought this isn't worth a NEWS entry but if you want I can add one.

### DIFF
--- a/R/standalone-types-check.R
+++ b/R/standalone-types-check.R
@@ -1,7 +1,7 @@
 # ---
 # repo: r-lib/rlang
 # file: standalone-types-check.R
-# last-updated: 2023-03-07
+# last-updated: 2023-03-13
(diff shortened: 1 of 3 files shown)


lionel-:
Thanks! Can you bump the `last-updated` field please? Or maybe I'll do it after release after merging, since there are multiple changes coming up.

lionel-:
It's best not to touch the versions by the way, to prevent conflicts. For instance I'm now merging into 1.0.6.9000 instead of 1.1.0.9000 as the release didn't go through.

PR #1593, opened 2023-03-08
Add `allow_infinite` to `check_number_whole()`

### LINKED ISSUE #1588: Add `allow_infinite` to `check_number_whole()`
This is useful e.g. when wrapping APIs with pagination. I frequently want everything so I pass something like `n = Inf`. In case I don't want everything `n` definitely should be a whole number.

mgirlich:
Closes #1588.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -69,7 +69,10 @@
 
 ## Miscellaneous fixes and features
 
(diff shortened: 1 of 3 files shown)


lionel-:
Thanks for these!

PR #1594, opened 2023-03-08
`check_number_*()` error messages

### LINKED ISSUE #1591: `check_whole_number()` closed interval messaging
Hi rlang team,

I think it would be great if during a check for nonzero positive integers, the output of `check_number_whole(0, min = 1)` would not be:

```r
! `0` must be a number larger than 1, not the number 0.
```

and instead could read: 

```r
`0` must be a whole number larger than or equal to 1, not the number 0
```

In particular, I'd like to confer that 1 is included in the interval. It would also be nice to message that these should be integers/whole numbers.

Please let me know if you'd welcome a PR for this.

teunbrand:
This PR aims to fix #1591.

Briefly, it does two things:
* Messages are about 'a whole number' if `allow_decimal = FALSE`, even when the number is OOB.
* The OOB check uses closed intervals, so for the upper and lower violation (not both), it now signals 'or equal to' as well.

If you'd like me to make an entry in the changelog or NEWS.md as well, please let me know.

### DIFF
--- a/R/standalone-types-check.R
+++ b/R/standalone-types-check.R
@@ -10,8 +10,9 @@
 # ## Changelog
 #
 # 2023-03-13:
-# - Added `allow_infinite` argument to `check_number_whole()`.
-# - Added `check_data_frame()`.
+# - Improved error messages of number checkers (@teunbrand)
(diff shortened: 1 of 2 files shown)


lionel-:
That's better, thanks!

PR #1599, opened 2023-03-10
Introduce `r_wrap_or_clone()` and `r_wrap_or_copy()`

DavisVaughan:
These are aliases for `R_shallow_duplicate_attr()` and `R_duplicate_attr()` respectively

Added in 2018, with R 3.6.0, so we can use them without issues in all R versions we support except for 3.6, which we are dropping soon anyways 
https://github.com/wch/r-source/commit/bbcf6868752fcd5481e2379f93ccbe02911eea63

I've chosen these function names because they either "wrap" the object into an ALTREP wrapper if possible (i.e. if the type is supported and it is "large enough"), or they call their respective `r_clone/copy()` equivalents.

There are 2 places I use these helpers, shown below. They are only useful if you are tweaking attributes and not the main object. Otherwise you should just `r_clone()`.

---

We now use `r_wrap_or_clone_shared()` in the re-encoding helper. If we re-encode an attribute, but not the main object, then we can "wrap" the main object before re-assigning the attributes to it.

``` r
x <- 1:1e6 + 0L
attr(x, "foo") <- `Encoding<-`("fa\xE7ile", "latin1")

vctrs:::is_altrep(x)
#> [1] TRUE

bench::mark(encode = .Call(rlang:::ffi_test_obj_encode_utf8, x))

# Main
#> # A tibble: 1 × 6
#> 1 encode        356µs    606µs     1186.    3.81MB     149.

# This PR
#> # A tibble: 1 × 6
#> 1 encode        1.1µs   1.53µs   499840.        0B        0


out <- .Call(rlang:::ffi_test_obj_encode_utf8, x)
vctrs:::is_altrep(out)

# Main
#> [1] FALSE

# This PR - now an altrep wrapper
#> [1] TRUE
```

---

We now also use `r_wrap_or_clone()` in `set_names()` to avoid having to call `names<-`, which is what we were doing before when I thought that was the only way to get access to the ALTREP wrappers. We now only call `names<-` if `x` is an "object" that might have an S3 method.

This gives us a little performance boost, but `names<-` was already very fast to call (typically in the 350ns range) so it doesn't look huge. It is relatively large percentage wise though, I guess.

I don't show any cases where `x` is an S3 object, that is typically in the 3.5us range because we have to do a `length()` dispatch call and a `names<-` dispatch call in case there is an S3 method registered, and those cases are completely separate from this PR. We could probably improve on this a little by actually checking to see if there is a `length()` or `names<-` method registered in the base/global env and only dispatching if we see one, but I didn't do that here.

``` r
library(rlang)

x <- 1:1e6 + 0L
names <- as.character(x)

# no names before, no names after
bench::mark(set_names(x, NULL), iterations = 1000000)
# before
#> # A tibble: 1 × 6
#>   expression              min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>         <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 set_names(x, NULL)   1.12µs   1.38µs   695965.    3.26KB     16.7

# after
#> # A tibble: 1 × 6
#>   expression              min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>         <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 set_names(x, NULL)    868ns   1.11µs   856415.     7.9KB     15.4

# no names before, names after
bench::mark(set_names(x, names), iterations = 1000000)

# before
#> # A tibble: 1 × 6
#>   expression               min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>          <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 set_names(x, names)   1.51µs   1.81µs   527374.        0B     18.5

# after
#> # A tibble: 1 × 6
#>   expression               min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>          <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 set_names(x, names)    1.2µs   1.52µs   636811.        0B     17.2

# this doesn't make an altrep wrapper, but `names<-` does
attr(x, "names") <- names

# names before, no names after
bench::mark(set_names(x, NULL), iterations = 1000000)

# before
#> # A tibble: 1 × 6
#>   expression              min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>         <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 set_names(x, NULL)   1.32µs   1.61µs   598498.        0B     19.8

# after
#> # A tibble: 1 × 6
#>   expression              min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>         <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 set_names(x, NULL)   1.03µs   1.34µs   720275.        0B     18.0

# names before, names after
bench::mark(set_names(x, names), iterations = 1000000)

# before
#> # A tibble: 1 × 6
#>   expression               min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>          <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 set_names(x, names)   1.59µs   1.92µs   501267.        0B     17.5

# after
#> # A tibble: 1 × 6
#>   expression               min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>          <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 set_names(x, names)   1.33µs   1.65µs   584423.        0B     15.8
```

### DIFF
--- a/src/internal/attr.c
+++ b/src/internal/attr.c
@@ -61,20 +61,26 @@ r_obj* node_names(r_obj* x) {
 }
 
 r_obj* ffi_set_names(r_obj* x, r_obj* mold, r_obj* nm, r_obj* env) {
-  int n_kept = 0;
-
-  r_obj* dots = KEEP_N(rlang_dots(env), &n_kept);
-
   if (!r_is_vector(x, -1)) {
     r_abort("`x` must be a vector");
   }
 
   if (nm == r_null) {
-    x = set_names_dispatch(x, r_null, env);
-
-    FREE(n_kept);
-    return x;
+    // Fast case for dropping names
+    if (r_is_object(x)) {
+      return set_names_dispatch(x, r_null, env);
+    } else if (r_names(x) != r_null) {
+      x = r_wrap_or_clone(x);
+      r_attrib_poke_names(x, r_null);
+      return x;
+    } else {
+      return x;
+    }
   }
+
+  int n_kept = 0;
+
+  r_obj* dots = KEEP_N(rlang_dots(env), &n_kept);
 
   if (r_is_function(nm) || r_is_formula(nm, -1, -1)) {
     if (r_names(mold) == r_null) {
@@ -85,12 +91,16 @@ r_obj* ffi_set_names(r_obj* x, r_obj* mold, r_obj* nm, r_obj* env) {
 
     nm = KEEP_N(rlang_as_function(nm, env), &n_kept);
     nm = KEEP_N(eval_fn_dots(nm, mold, dots, env), &n_kept);
-  } else {
-    if (r_length(dots) > 0) {
-      nm = KEEP_N(eval_fn_dots(c_fn, nm, dots, env), &n_kept);
-    }
+  } else if (r_length(dots) > 0) {
+    nm = KEEP_N(eval_fn_dots(c_fn, nm, dots, env), &n_kept);
+  }
 
+  if (r_typeof(nm) != R_TYPE_character || r_is_object(nm)) {
     nm = KEEP_N(eval_as_character(nm, env), &n_kept);
+
+    if (r_typeof(nm) != R_TYPE_character) {
+      r_abort("`nm` must be `NULL` or a character vector.");
+    }
   }
 
   r_ssize n;
@@ -98,10 +108,6 @@ r_obj* ffi_set_names(r_obj* x, r_obj* mold, r_obj* nm, r_obj* env) {
     n = length_dispatch(x, env);
   } else {
     n = r_length(x);
-  }
-
-  if (r_typeof(nm) != R_TYPE_character) {
-    r_abort("`nm` must be `NULL` or a character vector.");
   }
 
   r_ssize nm_n = r_length(nm);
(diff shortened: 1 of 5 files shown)


src/internal/attr.c
```diff
@@ -61,21 +61,27 @@ r_obj* node_names(r_obj* x) {
 }
 
 r_obj* ffi_set_names(r_obj* x, r_obj* mold, r_obj* nm, r_obj* env) {
-  int n_kept = 0;
-
-  r_obj* dots = KEEP_N(rlang_dots(env), &n_kept);
```
DavisVaughan, line 66: I moved these after the early exit for `nm == NULL` because they aren't needed, just for maximal performance and code clarity

src/internal/attr.c
```diff
@@ -85,12 +91,16 @@ r_obj* ffi_set_names(r_obj* x, r_obj* mold, r_obj* nm, r_obj* env) {
...
+    nm = KEEP_N(eval_fn_dots(c_fn, nm, dots, env), &n_kept);
+  }
 
+  if (r_typeof(nm) != R_TYPE_character || r_is_object(nm)) {
     nm = KEEP_N(eval_as_character(nm, env), &n_kept);
+
+    if (r_typeof(nm) != R_TYPE_character) {
+      r_abort("`nm` must be `NULL` or a character vector.");
+    }
   }
```
DavisVaughan, line 104: I tweaked this section a little

- The `nm == function` case didn't call `as.character()` on the result of the function, but I felt like it was reasonable to do so since we call `as.character()` on the result of `set_names(x, 1, 2)` when `...` are combined with `c()`, and on `set_names(x, 1)` when `nm` just isn't a character vector.

- We also had some redundant "if not character, then error" checks below that I've removed in favor of just this single character check.

PR #1604, opened 2023-03-20
Never keep sources in `parse_*()`

### LINKED ISSUE #1603: Add `keep_source` to `parse_exprs()` and friends
These call `parse()` which defaults to `keep.source = TRUE`, but you typically don't need that and it is super slow. It ends up calling `srcfilecopy()` which even calls `getwd()` internally!

```r
bench::mark(parse(text = "x"), parse(text = "x", keep.source = F), check = F)
#> # A tibble: 2 × 6
#>   expression                              min   median `itr/sec` mem_a…¹ gc/se…²
#>   <bch:expr>                         <bch:tm> <bch:tm>     <dbl> <bch:b>   <dbl>
#> 1 parse(text = "x")                   47.53µs   51.3µs    17793.    280B    4.04
#> 2 parse(text = "x", keep.source = F)   3.01µs   3.87µs   256832.      0B   25.7 
#> # … with abbreviated variable names ¹​mem_alloc, ²​`gc/sec`
```

Consider defaulting `keep_source` to `FALSE`? Not sure how important the sources are in typical usage of this function.

DavisVaughan:
Closes #1603 

It turns out that in both rlang 1.0.6 and 1.1.0, source information was already being dropped in all cases (I wanted to ensure that removing usage of `flatten()` was not the result of this change, and it doesn't seem to be).

I think this makes sense as the source reference information is typically attached to the expression object, and:
- `parse_expr()` returns the individual expression element, not the expression vector
- `parse_exprs()` can parse and flatten _multiple_ expressions, and since each expression can return >1 element from `parse()`, you don't have a good 1:1 mapping between the source information and the flattened parsed expressions, so it makes sense to just drop the source information here too

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # rlang (development version)
 
+* `parse_expr()` and `parse_exprs()` are now faster (#1603).
+
 # rlang 1.1.0
 
 ## Life cycle changes
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # rlang (development version)
 
-* `parse_expr()` and `parse_exprs()` are now faster (#1603).
+* `parse_expr()` and `parse_exprs()` are now faster when
+  `getOption("keep.source")` is `TRUE` (#1603).
 
 # rlang 1.1.0
 
--- a/R/parse.R
+++ b/R/parse.R
@@ -35,6 +35,10 @@
 #'   the latter is more convenient as you don't need to extract `expr`
 #'   and `env`.
 #'
+#' @details
+#' Unlike [base::parse()], these functions never retain source reference
+#' information, as doing so is slow and rarely necessary.
+#'
 #' @param x Text containing expressions to parse_expr for
 #'   `parse_expr()` and `parse_exprs()`. Can also be an R connection,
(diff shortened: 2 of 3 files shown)


lionel- (approved):
> you don't have a good 1:1 mapping between the source information and the flattened parsed expressions, so it makes sense to just drop the source information here too

Unless I'm missing something I think the mapping is fine but we'd have to return an expression vector to preserve the srcrefs.

DavisVaughan:
I suppose you can manually merge the `srcref` list, but I'm not sure what you'd do with `$wholeSrcref` here

```r
attributes(parse(text = "foo(); bar"))
#> $srcref
#> $srcref[[1]]
#> foo()
#> 
#> $srcref[[2]]
#> bar
#> 
#> 
#> $srcfile
#> <text> 
#> 
#> $wholeSrcref
#> foo(); bar

attributes(parse(text = "baz"))
#> $srcref
#> $srcref[[1]]
#> baz
#> 
#> 
#> $srcfile
#> <text> 
#> 
#> $wholeSrcref
#> baz
```

lionel-:
hmm I think this attribute is not required.

PR #1605, opened 2023-03-21
Add `check_list()`

DavisVaughan:
Makes sense to me to have this lower level version too even if we have `vctrs::obj_check_list()`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # rlang (development version)
 
+* New `check_list()` in `standalone-types-check.R` (#1605).
+
 * `parse_expr()` and `parse_exprs()` are now faster when
   `getOption("keep.source")` is `TRUE` (#1603).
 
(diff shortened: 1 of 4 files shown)


lionel- (commented):
I'm just not sure about the predicate. I think we should probably use `inherits(x, "list")`, cf Slack discussion with Julia.

Thought: What would a (non-vctrs) list-of check look like? Would it take a type predicate for the elements and use chained errors?

DavisVaughan:
Regarding list-of, yea kind of like the C++ `all_of` that we also modeled `r_list_all_of()` after https://en.cppreference.com/w/cpp/algorithm/all_any_none_of

But yea ok i forgot about `inherits(x, "list")`. I'm happy to close this for now since we aren't really sure about that yet, and we can make the decision about that when we really need it. It isn't critical for me to have this.

PR #1611, opened 2023-03-30
handle `I()` input literally in `stop_input_type(arg)`

### LINKED ISSUE #1607: Consider allowing `arg = I("custom arg name")` in `stop_input_type()`
So you can do something like:

```r
stop_input_type(x, "a numeric vector", arg = I(glue("Element {i} of `x`"))
#> Element 5 of `x` must be a numeric vector, not a factor.
```

Useful if you are mapping over a user supplied list (like if the user supplied `...`) and you want to check that each one is a particular type. Otherwise I'd end up calling `cli_abort()` directly to completely control the error message

simonpcouch:
Closes #1607.🍄

### DIFF
--- a/R/standalone-obj-type.R
+++ b/R/standalone-obj-type.R
@@ -1,12 +1,15 @@
 # ---
 # repo: r-lib/rlang
 # file: standalone-obj-type.R
-# last-updated: 2022-10-04
+# last-updated: 2023-03-30
 # license: https://unlicense.org
 # imports: rlang (>= 1.1.0)
 # ---
 #
 # ## Changelog
+#
+# 2023-03-30:
+# - `stop_input_type()` now handles `I()` input literally in `arg`.
 #
 # 2022-10-04:
 # - `obj_type_friendly(value = TRUE)` now shows numeric scalars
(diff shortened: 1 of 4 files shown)


R/standalone-obj-type.R
```diff
@@ -311,10 +314,15 @@ stop_input_type <- function(x,
   if (length(what)) {
     what <- oxford_comma(what)
   }
+  if (inherits(arg, "AsIs")) {
```
DavisVaughan, line 317: We are assuming that the `abort(arg = arg)` call down below can handle an `arg` that doesn't actually match any argument in the call by just ignoring it - because presumably if you supply an AsIs `arg` then it is going to look completely customized and won't match anything in the call.

Alternatively we could set something like `arg_highlight <- NULL` and pass that through to `abort()` in this case instead, but I will let Lionel make that decision

R/standalone-obj-type.R
```diff
@@ -311,10 +314,15 @@ stop_input_type <- function(x,
...
+    format_arg <- identity
```
lionel-, line 318: I'm not sure if this should be `unclass()`. `abort()` does ignore unmatching `arg` but maybe the `AsIs` class could be detected to ignore it unconditionally.

lionel-:
Thanks! 🌳

PR #1616, opened 2023-04-14
Fix for dev testthat

hadley:
Which now correctly checks the overall class of the underlying condition when you set `class`

(And can we plan a patch release in the next two weeks?)


lionel-:
Thanks, and sure!

lionel-:
On CRAN.

PR #1631, opened 2023-06-02
Remove speculation about the end of the walrus.

jonthegeek:
Issue 1296 was closed saying the walrus must remain, so this section of this doc doesn't make sense anymore.


lionel-:
Good catch, thanks for this. I think next time I'm working on rlang I'll update your PR to include some notes about why we can't use `=`.

PR #1651, opened 2023-08-16
Update _pkgdown.yml

olivroy:
Tweak, so that rlang's site looks more like other tidyverse sites. This puts news to the left of the search bar.

https://pkgdown.r-lib.org/articles/customise.html#navbar

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -13,8 +13,7 @@ template:
 
 navbar:
   structure:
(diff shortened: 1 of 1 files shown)


olivroy:
@lionel- The search bar has disappeared from the pkgdown site with the recent 2.1 update. This PR will resolve that. I can also make some tweaks if you'd rather keep the news on the right.

lionel-:
Thanks!

PR #1657, opened 2023-10-06
speed up package version check in `.rlang_cli_compat()`

simonpcouch:
I recently came across a slowdown when profiling some tidymodels code that I thought might be worth raising. The `utils::packageVersion("rlang") >= "0.4.2"` check in `standalone-cli.R` adds up for us downstream.

With rlang 1.1.1:

``` r
bench::mark(
  signal = rlang::signal(class = "boop"),
  group_by = dplyr::group_by(mtcars, cyl),
  check = FALSE
)
#> # A tibble: 2 × 6
#> 1 signal        276µs    288µs     3393.  296.68KB     16.8
#> 2 group_by      860µs    900µs     1080.    3.49MB     16.8
```

With this PR:

``` r
#> # A tibble: 2 × 6
#> 1 signal       46.1µs   47.8µs    20399.  294.26KB     32.0
#> 2 group_by    572.5µs  604.7µs     1603.    3.48MB     23.2
```

I believe this should be compatible with the "standalone" approach but feels a bit more hacky than I'd usually be comfortable submitting. Very much open to iterate. :)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@
 
 * `expr_label()` now has back-compatility with respect to changes made by R version 4.4 and `is.atomic(NULL)` (#1655)
 
+* Performance improvement in `.rlang_cli_compat()` (#1657).
+
 # rlang 1.1.1
 
 * `englue()` now allows omitting `{{`. This is to make it easier to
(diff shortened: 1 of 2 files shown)


lionel-:
Thanks! The other thing you could do is cache the information in a top-level object namespaced with the compat prefix.

Issue #1426, opened 2022-06-19
Find the OO system used for an object


hadley:
Something like this:

```R
obj_oo_type <- function(x) {
  if (is.object(x)) {
    if (inherits(x, "R7_object")) {
      "R7"
    } else if (inherits(x, "R6")) {
      "R6"
    } else if (isS4(x)) {
      "S4"
    } else {
      "S3"
    }
  } else {
    "base"
  }
}
```

This might possibly live in shim file, but that partly depends on the result of https://github.com/r-lib/cli/issues/463 and https://github.com/r-lib/vctrs/issues/139 — I think we need to host a consistent way of describing types somewhere centrally (either in rlang or maybe in a new package).

lionel-:
Maybe we should add this to `compat-friendly-type.R`?

hadley:
Yeah, that's probably a reasonable place for now. 

I can't quite figure if there's enough here for a separate package or not. When you consider the full scope of abbreviated vs full vs user facing type names, there is potentially a lot of code that we'd want to use in a lot of places.

PR #1688, opened 2024-02-14
`friendly_type_of()` recognizes S7 objects

olivroy:
Adjust obj-type to reflect this. Follow-up to #1426


olivroy:
@lionel- I think this PR is low risk. R7 objects have been renamed to S7 objects almost a year ago now

lionel-:
Thanks!

PR #1693, opened 2024-02-22
Allow reticulate errors in cnd_type() (fixes #1664)

### LINKED ISSUE #1664: Support reticulate conditions
e.g. https://github.com/tidyverse/purrr/issues/1104

gergness:
Not sure if this is the right way to fix, but was useful enough for me for that I thought I'd share.

reticulate stores it's conditions with the right classes to work with rlang, but uses an environment instead of a list.

``` r
err <- try(reticulate::py_run_string("oops"), silent = TRUE) |> attr("condition")

is.environment(err)
#> [1] TRUE
class(err)
#> [1] "python.builtin.NameError"     "python.builtin.Exception"    
#> [3] "python.builtin.BaseException" "python.builtin.object"       
#> [5] "error"                        "condition"
```

<sup>Created on 2024-02-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/src/rlang/cnd.c
+++ b/src/rlang/cnd.c
@@ -100,7 +100,8 @@ void r_interrupt(void) {
 
 enum r_cnd_type r_cnd_type(r_obj* cnd) {
   r_obj* classes = r_class(cnd);
-  if (r_typeof(cnd) != R_TYPE_list ||
+  if (
+      (r_typeof(cnd) != R_TYPE_list && r_typeof(cnd) != R_TYPE_environment) ||
       r_typeof(classes) != R_TYPE_character) {
     goto error;
   }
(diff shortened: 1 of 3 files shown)


src/rlang/cnd.c
```diff
@@ -100,7 +100,8 @@ void r_interrupt(void) {
 
 enum r_cnd_type r_cnd_type(r_obj* cnd) {
   r_obj* classes = r_class(cnd);
-  if (r_typeof(cnd) != R_TYPE_list ||
+  if (
```
lionel-, line 103: Changing this will cause R to crash if the caller assumes (rightfully IMO) that a condition is a list.

PR #1696, opened 2024-03-12
docs: Document `on_load()` detail

### LINKED ISSUE #1695: `on_load()` contents not analyzed by `R CMD check`
Learned painfully today. Is it worth mentioning this in `?on_load` ?

Reference: https://github.com/duckdblabs/duckplyr/pull/122.

krlmlr:
Closes #1695.

### DIFF
--- a/R/aaa.R
+++ b/R/aaa.R
@@ -69,6 +69,10 @@ the <- new.env(parent = emptyenv())
 #' `on_load()` is implemented via `.onLoad()` and requires
 #' `run_on_load()` to be called from that hook.
 #'
+#' The expressions inside `on_load()` do not undergo static analysis
+#' by `R CMD check`. Therefore, it is advisable to only use
+#' simple function calls inside `on_load()`.
+#'
 #' @examples
 #' quote({  # Not run
 #' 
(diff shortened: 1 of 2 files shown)


R/aaa.R
```diff
@@ -69,6 +69,10 @@ the <- new.env(parent = emptyenv())
 #' `on_load()` is implemented via `.onLoad()` and requires
 #' `run_on_load()` to be called from that hook.
 #'
+#' The expressions inside `on_load()` do not undergo static analysis
+#' by `R CMD check`. Therefore, it is advisable to only use
+#' simple function calls inside `on_load()`.
```
lionel-, line 74: I think we could add support for on_load to other tools so I'd remove the latter part:

```suggestion
#' Note that the expressions inside `on_load()` do not undergo static analysis
#' by `R CMD check`.
```

krlmlr: I don't understand "support ... to other tools", I'm fine with the change if it makes sense to you.

lionel-: On the other hand extracting to a function also makes sense in terms of style. Let's just merge as is.

lionel-:
Thanks!

PR #1697, opened 2024-03-12
docs: Wrap `on_load()` contents in examples with braces

krlmlr:
This gives better tracebacks, I suppose.


lionel-:
How does it give better tracebacks?

krlmlr:
I've noticed that srcrefs are available only for braced expressions, not for naked ones. Can't vouch for this specific case, but observed with `expect_snapshot()` . @hadley: would you like to share your perspective?

lionel-:
oh yeah that's definitely the case

PR #1699, opened 2024-03-28
Add `r_dyn_raw_push_back()` and `r_dyn_chr_push_back()`

DavisVaughan:
Added 2 missing push-back helpers.

In https://github.com/r-lib/rlang/pull/1542 I optimized the push-back API by separating it from the generic `r_dyn_push_back()` helper. This was nice, but it left the _typed_ push back API untested. In addition to adding those 2 missing helpers, this PR also adds the missing tests for the typed API.


DavisVaughan:
Test failures seem unrelated but definitely worth taking a look at, as it seems important?

lionel- (approved):
I've fixed the failures on main!

PR #1701, opened 2024-04-17
Fix bug in `vec_recycle_common()`

### LINKED ISSUE #1700: Missing error in `vec_recycle_common()` shim
I believe the following is supposed to throw an error, as the output is not a list where every element is of length 1.
``` r
rlang:::vec_recycle_common(list(a = 1, b = 1:2), size = 1)
#> $a
#> [1] 1
#> 
#> $b
#> [1] 1 2
```

<sup>Created on 2024-04-12 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

The function has a specific branch to throw an appropriate error in the part below.

https://github.com/r-lib/rlang/blob/65018f9d6212610a518f6e38d87a427d98422824/R/standalone-vctrs.R#L121-L123

I think `ns != size` should be `n != size` for the error to trigger correctly.
I'd be happy to prepare a PR if welcome.

teunbrand:
This PR aims to fix #1700.

It restores an error that should be thrown when `size = 1` is requested, but input has larger vectors.
Reprex from issue with this PR:

``` r
rlang:::vec_recycle_common(list(a = 1, b = 1:2), size = 1)
#> Error: Inputs can't be recycled to `size`.
```

<sup>Created on 2024-04-17 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/standalone-vctrs.R
+++ b/R/standalone-vctrs.R
@@ -13,6 +13,10 @@
 # of data frames without having to depend on tibble or vctrs. The
 # embedded type system is minimal and not extensible.
 
+# 2024-04-17:
+# * `vec_recycle_common()` throws intended error when `size = 1` but input
+#   is larger.
+
 # 2021-08-27:
 # * `vec_slice()` now preserves attributes of data frames and vectors.
 # * `vec_ptype2()` detects unspecified columns of data frames.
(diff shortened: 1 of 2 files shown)


lionel-:
Can you add a test please?

teunbrand:
Sure thing!

lionel-:
Thanks!

PR #1704, opened 2024-04-26
Show how to use call = `base::call()` in `abort()`

olivroy:
I discovered yesterday that if you wanted  to hard-code the `call` argument, it was possible, but never found this in rlang's `abort()` documentation,

I thought adding an example could be useful? Sometimes, I think it can be useful to hard-code it? May be easier to learn to use this than all of rlang's infrastructure.


lionel-:
Thanks!

PR #1709, opened 2024-05-07
Add cross-ref between the ellipsis checking functions

olivroy:
Currently, there is no reference to other dot checking functions. So adding a `@family` adds the following to see also

image

Maybe worth adding `dots_n()`?


lionel-:
oh good idea, thanks!

PR #1728, opened 2024-07-02
Remove dependency on `LEVELS()`

### LINKED ISSUE #1726: Remove `LEVELS()` accessor
Part of https://github.com/r-lib/rlang/issues/1706

`LEVELS()` was added in https://github.com/r-lib/rlang/pull/1187

Used to detect non-utf8 strings in `r_obj_encode_utf8()` and optimise the default case of no encoding conversion needed:

- Avoid cloning a STRSXP if no translation is needed.
- Avoid allocating a CHARSXP if no translation is needed (in case of mixed encodings in the vector, less important I guess?).

We could just call `Encoding(x) <- "UTF8"` from R. This will unconditionally clone the input vector unless it doesn't have any references. To preserve our optimisation, either of these would work:

- Add a predicate to the C API to determine if a `CHARSXP` is encoded in UTF8

- Improve `do_setencoding()` to only duplicate if needed so that it's a noop in the common case. Bonus points if exported on C side, e.g. as `Rf_EnsureUtf8()`?

My sense is that these new unconditional allocs would be bad for performance in dplyr. @DavisVaughan you added these optimisations for vctrs, could you confirm please?

lionel-:
Closes #1726


DavisVaughan (approved):
We determined the one place this is actually used right now is in rlang's `chr_detect_dups()` function here
https://github.com/r-lib/rlang/blob/23f0f306aac07a5b9d8807c1d60e776f0c890ce0/src/internal/utils.c#L228

Which is used by `names_as_unique()`

https://github.com/r-lib/rlang/blob/23f0f306aac07a5b9d8807c1d60e776f0c890ce0/src/internal/names.c#L49

Which is a common name repair strategy in the tidyverse.

Since no tests failed we determined this is probably going to work fine.

---

It is also worth noting that this is not currently used in vctrs right now. We did do the work to expose this in the rlang C API as an rlang callable, but we never went back and updated vctrs to use it (instead it uses its own local copy of a similar function).

---

Lastly, here is the main change:

https://github.com/wch/r-source/blob/7f80ed7ae28959026e071a7c58cd38fa3a3a8f9f/src/main/sysutils.c#L962

The one case `Rf_charIsUTF8` seems to be different for us is when:
* String is not "marked" as UTF8
* But `utf8locale` is set to `true`

In this case we previously assumed it was _not_ UTF8 because it was "unknown" to us and we can't access `utf8locale`

But R can access `utf8locale` so I guess it assumes that if the string is not marked as anything and not `NA` and `utf8locale` is set, then it must be UTF8

This case is probably quite rare, and the only difference in behavior is that we used to call  `Rf_translateCharUTF8()` on these "unknown" strings, but R seems to think they are UTF8 anyways, so that was probably already a no-op in terms of actual behavior (_possibly_ it would add the UTF8 marking, but it probably didn't change the underlying bytes).

PR #1731, opened 2024-07-09
Make rlang clangd-okay

lionel-:
- Make C headers and sources self-contained.

- Add IWYU pragmas (see https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md and https://github.com/clangd/clangd/issues/1085#issuecomment-1082060994 for context) in rlang header files so that clangd completions don't try to include private headers.

- Disable diagnostics in the `decl` folder since these headers are not self-contained.


DavisVaughan (approved):
Woo!

Issue #1618, opened 2023-04-18
add `check_length()` to `standalone-types-check.R`


ateucher:
It would be nice to be able to check if a vector is of a required length (exact, min or max). The scalar checkers `check_name()`, `check_string()` etc., ensure an object is of length 1, but being able to check length constraints > 1 would be helpful. I could see this being implemented as something like:

`check_length(x, ..., length, arg = caller_arg(x), call = caller_env())`

where `length` could be either a scalar integer value to specify an exact required length, or a length 2 vector supplying an allowed range of lengths (which could include `-Inf` or `Inf` if only a minimum or maximum length is required).

Alternatively, a `length` argument could be added to the vector checkers such as `check_character()` and `check_logical()`.

lionel-:
I think that makes sense. I'd just put `length` on the left of `...` to avoid the repetition in calls, e.g. `check_length(x, length = 3)` vs `check_length(x, 3)`. I'm not sure we need to allow for ranges of length though. But sometimes we want to allow multiple lengths when recycling is allowed, e.g. length 1 or length n. So maybe it should rather be `length = c(1, n)`?

olivroy:
I think `check_character()` should have a way to disallow `character(0)`, i.e. length 1 or more.

the use case is that I want to accept `NULL` (length 0), and character of length 1 or more (not `character(0)`)

PR #1734, opened 2024-07-14
Add `min_length` to `check_character()`

olivroy:
Sometimes you want something non-empty. `allow_zero = FALSE` will disallow `character(0)`. Addresses https://github.com/r-lib/rlang/issues/1618#issuecomment-2127898009


lionel- (commented):
I think this should be `allow_empty` and apply to all vector types?

Could also be something like `length = c("any", "non_empty", "scalar")`? It seems like this change deserves some discussion about general vector checks.

olivroy:
I agree. However 'allow_empty` would be confusing since for `check_string()` `allow_empty = FALSE is meant to disallow `""`.

R/standalone-types-check.R
```diff
@@ -1,13 +1,15 @@
...
-# last-updated: 2023-03-13
+# last-updated: 2024-07-14
 # license: https://unlicense.org
 # dependencies: standalone-obj-type.R
 # imports: rlang (>= 1.1.0)
 # ---
 #
 # ## Changelog
+# 2024-07-14:
+# - `check_character()` gains `allow_zero` to allow prohibiting `character(0)` (@olivroy)
```
olivroy, line 12:
```suggestion
# - `check_character()` gains `min_length` to allow prohibiting `character(0)` (@olivroy)
```

R/standalone-types-check.R
```diff
@@ -460,10 +462,11 @@ check_formula <- function(x,
 check_character <- function(x,
                             ...,
                             allow_null = FALSE,
+                            allow_zero = TRUE,
```
olivroy, line 465:
```suggestion
                            min_length = 0,
```

R/standalone-types-check.R
```diff
@@ -460,10 +462,11 @@ check_formula <- function(x,
...
                             arg = caller_arg(x),
                             call = caller_env()) {
   if (!missing(x)) {
-    if (is_character(x)) {
+    if (is_character(x) && (allow_zero || length(x) > 0)) {
```
olivroy, line 469:
```suggestion
    if (is_character(x) && (length(x) >= min_length)) {
```

olivroy (commented):
Modify the `stop_input_type()` call for:

```r
  what <- "character vector"  
  if (min_length > 0) {
    what <- paste0(what, "of length greater or equal to", min_length) 
  }
```

R/standalone-types-check.R
```diff
@@ -460,20 +462,26 @@ check_formula <- function(x,
...
+    if (is_character(x) && (length(x) >= min_length)) {
       return(invisible(NULL))
     }
     if (allow_null && is_null(x)) {
       return(invisible(NULL))
     }
   }
 
+  what <- "a character vector"  
+  if (min_length > 0) {
```
olivroy, line 478:
```suggestion
  if (min_length > 1) {
```

Actually, the error message will be clear enough for length 1

R/standalone-types-check.R
```diff
@@ -460,20 +462,26 @@ check_formula <- function(x,
...
       return(invisible(NULL))
     }
     if (allow_null && is_null(x)) {
       return(invisible(NULL))
     }
   }
 
+  what <- "a character vector"  
+  if (min_length > 1) {
+    what <- paste0(what, "of length greater or equal to", min_length)
```
olivroy, line 479:
```suggestion
    what <- paste(what, "of length greater or equal to", min_length) 
```

Issue #1712, opened 2024-05-17
`is_dictionarish(NULL)` should return `TRUE`


hadley:
I think, because there are no names, hence the names are necessarily unique.

lionel-:
On the other hand `NULL` is not a vector.

```r
vctrs::vec_is(NULL)
#> [1] FALSE

is_vector(NULL)
#> [1] FALSE
```

On the other other hand, you can take the names and length of `NULL`:

```r
names(NULL)
#> NULL
```

I'm just wondering what kind of invariants we can attribute to dictionaryish objects. Is it just that they have a length and names, and the names are unique?

This would be consistent with:

```r
is_dictionaryish(env())
#> [1] TRUE
```

Do you remember what you needed this for? Having some examples of patterns that would be simplified by adding this special-case for NULL would be helpful to understand the relevant invariants.

hadley:
I don't, but I suspect it was an argument where the default was `NULL` and the expected type was a list.

PR #1741, opened 2024-08-15
dictionaryish(NULL) return true

ilovemane:
#1712

### DIFF
--- a/R/attr.R
+++ b/R/attr.R
@@ -110,13 +110,15 @@ detect_void_name <- function(x) {
 is_dictionaryish <- function(x) {
   # 2022-01: Used in many packages. Don't deprecate without a
   # replacement.
-  if (!length(x)) {
-    return(!is.null(x))
-  }
-  if (is.null(names(x))) {
+  if (is.null(x)) {
     TRUE
   } else {
+    if (!length(x)) {
+      return(!is.null(x))
+    }
+
     is_named(x) && !any(duplicated(names(x)))
+
   }
 }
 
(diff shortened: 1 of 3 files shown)


R/attr.R
```diff
@@ -110,11 +110,16 @@ detect_void_name <- function(x) {
 is_dictionaryish <- function(x) {
   # 2022-01: Used in many packages. Don't deprecate without a
   # replacement.
-  if (!length(x)) {
-    return(!is.null(x))
-  }
+  if (is.null(x)) {
+    TRUE
```
lionel-, line 114: I'd do this with an early return so that the rest of the function is not nested in a branch. So something like:

```r
if (is.null(x)) {
  return(TRUE)
}
if (!length(x)) {
  return(!is.null(x))
}
```

R/attr.R
```diff
@@ -110,11 +110,16 @@ detect_void_name <- function(x) {
...
+  } else {
+    if (!length(x)) {
+      return(!is.null(x))
+    }
 
-  is_named(x) && !any(duplicated(names(x)))
+    is_named(x) && !any(duplicated(names(x)))
+
```
lionel-, line 121: Can you remove this new unnecessary whitespace please?

lionel- (commented):
Looking good! Just a couple of nitpicks.

Also can you please add a NEWS bullet referencing the issue number and tagging your username?

R/attr.R
ilovemane: done!

R/attr.R
ilovemane: done!

lionel-:
Thank you!

PR #1742, opened 2024-08-15
add allow_na argument to check character()

### LINKED ISSUE #1724: check_character() always permits NA values
i would like to ensure that `check_character()` does not permit NA values. However, it _always_ allows them. 

``` r
source("https://raw.githubusercontent.com/r-lib/rlang/main/R/standalone-obj-type.R")
source("https://raw.githubusercontent.com/r-lib/rlang/main/R/standalone-types-check.R")

# confirm that i have greater than 1.1
packageVersion("rlang")
#> [1] '1.1.3'
```

``` r

# import rlang
library(rlang)

check_character(c("a", NA, "c"), allow_na = FALSE)
```

martaalcalde:
Addresses part of #1724


martaalcalde:
Although the error message that it appears now may not be very intuitive:
Screenshot 2024-08-15 at 12 04 28

R/standalone-types-check.R
```diff
@@ -459,11 +459,16 @@ check_formula <- function(x,
...
                             allow_null = FALSE,
                             arg = caller_arg(x),
                             call = caller_env()) {
+
   if (!missing(x)) {
-    if (is_character(x)) {
+    if (is_character(x) & allow_na) {
+      return(invisible(NULL))
+    }
+    if(is.character(x) & !(TRUE %in% is.na(x)) & !allow_na){
```
lionel-, line 471: These checks should use the scalar boolean operators, e.g. `&&`. (Also missing a space between `if` and `(`.)

lionel-: A simpler way of expressing `TRUE %in% is.na(x)` is `any(is.na(x))`.

lionel-: To solve the error message I think I'd structure this code like this:

```r
if (is_character(x)) {
  if (!allow_na && any(is.na(x))) {
    abort(...)
  }
  return(invisible(NULL))
}
```

With a specially crafted error message in the `abort()` call.

tests/testthat/test-standalone-types-check.R
```diff
@@ -152,11 +152,16 @@ test_that("`check_environment()` checks", {
...
-  expect_null(check_character(na_chr))
+  expect_error(check_character(na_chr))
   expect_null(check_character(chr()))
   expect_null(check_character("foo"))
   expect_null(check_character(letters))
+
   expect_null(check_character(NULL, allow_null = TRUE))
+  expect_error(check_character(NULL, allow_null = FALSE))
+
+  expect_error(check_character(c("a",NA)))
```
lionel-, line 163: This should be tested in the `expect_snapshot()` block below so the error message is tracked.

R/standalone-types-check.R
```diff
@@ -459,23 +459,34 @@ check_formula <- function(x,
 
 check_character <- function(x,
                             ...,
+                            allow_na = FALSE,
```
olivroy, line 462: This would be a breaking change here?

lionel-:
Thanks @martaalcalde!

R/standalone-types-check.R
lionel-: Now changed to `TRUE`.

PR #1743, opened 2024-08-15
start check_numbers_decimal

catalamarti:
does this looks like it is going in the right direction? @lionel-

### DIFF
--- a/R/standalone-types-check.R
+++ b/R/standalone-types-check.R
@@ -190,6 +190,7 @@ check_number_decimal <- function(x,
     ...,
     exit_code = exit_code,
     allow_decimal = TRUE,
+    vector = FALSE,
     min = min,
     max = max,
     allow_na = allow_na,
@@ -228,6 +229,7 @@ check_number_whole <- function(x,
     ...,
     exit_code = exit_code,
     allow_decimal = FALSE,
+    vector = FALSE,
     min = min,
     max = max,
     allow_na = allow_na,
@@ -241,6 +243,7 @@ check_number_whole <- function(x,
                              ...,
                              exit_code,
                              allow_decimal,
+                             vector,
                              min,
                              max,
                              allow_na,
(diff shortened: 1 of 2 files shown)


R/standalone-types-check.R
```diff
@@ -253,6 +256,10 @@ check_number_whole <- function(x,
     what <- "a whole number"
   }
 
+  if (vector) {
+    what <- sprintf("%s vector")
```
lionel-, line 260: missing an argument?

R/standalone-types-check.R
```diff
@@ -531,4 +538,43 @@ check_data_frame <- function(x,
...
+  ))))) {
+    return(invisible(NULL))
+  }
+
+  .stop_not_number(
+    x,
+    ...,
+    exit_code = exit_code,
+    allow_decimal = TRUE,
+    vector = TRUE,
```
lionel-, line 570: Potential upgrade: Should probably pass the indices of the failing elements, as well as the reasons, so they can be reported in the error message. These should be truncated if there are too many.

But in a first pass it's fine to just fail saying all elements in the vector should follow the required invariants.

tests/testthat/test-standalone-types-check.R
```diff
@@ -201,3 +201,13 @@ test_that("`check_data_frame()` checks", {
     err(checker(list(data.frame(), data.frame()), check_data_frame, allow_null = TRUE))
   })
 })
+
+test_that("`check_numbers_decimal` checks", {
+  expect_null(check_numbers_decimal(c(10, 5.5)))
+  expect_null(check_numbers_decimal(c(10L, 50L)))
+
+  expect_snapshot({
```
lionel-, line 209: It looks like you still need to generate and commit this snapshot.

lionel- (commented):
A few comments but looking good. Also need to add a news bullet at the top of the standalone file.

PR #1758, opened 2024-10-24
Simplify tests by using `cnd_class = TRUE` from `expect_snapshot()`

olivroy:
Not 100% passed through, but thought I'd send this.

Makes snapshot results more readable in general


lionel-:
Thanks!

PR #1762, opened 2024-11-25
Prefer skip_if_not() to skip_if(!)

MichaelChirico:
It will give more readable skip messages, currently we get:

```
• !compiled_by_gcc() is TRUE (1): test-c-api.R:1113:3
```


lionel-:
Thanks!

PR #1767, opened 2024-12-12
fixing some call_modify examples w/ NULL -> zap()

jmobrien:
Some examples in call_modify() predate the introduction of zap() and use the older `[arg] = NULL` syntax.  In particular, this made line 560's example, `call_modify(call, bar = NULL, bar = missing_arg())`, no longer illustrate its key behavior.  Changed that, plus 2 other (likely less important) examples where argument removal seems to have been intended.


lionel-:
Thanks!

PR #1769, opened 2025-01-08
Simplify reporting of bracktraces during knitting

lionel-:
For https://github.com/r-lib/evaluate/pull/232#discussion_r1906023821


hadley:
I think it would be worth doing a patch release of rlang for this.

Issue #1608, opened 2023-03-28
Cache result of `rlang:::is_on_disk()`


DavisVaughan:
Due to it calling `system.file()`, which is typically around half the time of `is_installed()` or `check_installed()`.

Would also be good to look into what the rest of the time is taken up by, but this is a good first step.

``` r
library(rlang)

bench::mark(check_installed("dplyr"))
#> # A tibble: 1 × 6
#>   expression                    min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>               <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 check_installed("dplyr")    760µs    877µs     1119.    8.11MB     10.4

bench::mark(is_installed("dplyr"))
#> # A tibble: 1 × 6
#>   expression                 min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>            <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 is_installed("dplyr")    755µs    854µs     1139.    1.95KB     12.6

bench::mark(requireNamespace("dplyr", quietly = TRUE))
#> # A tibble: 1 × 6
#>   expression                                  min median itr/s…¹ mem_a…² gc/se…³
#>   <bch:expr>                                <bch> <bch:>   <dbl> <bch:b>   <dbl>
#> 1 requireNamespace("dplyr", quietly = TRUE) 956ns 1.22µs 791422.      0B       0
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`

# Often around half the time
bench::mark(system.file(package = "dplyr"))
#> # A tibble: 1 × 6
#>   expression                          min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                     <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 system.file(package = "dplyr")    334µs    352µs     2434.    1.95KB     4.06
```

<sup>Created on 2023-03-28 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

Issue #1776, opened 2025-02-12
Propose a faster version of `check_installed()`


shikokuchuo:
This function provides a much better user experience for code using 'suggests' packages over some formulation of
`if (requireNamespace("x", quietly = TRUE))`.

However, it's much more expensive to use (a full 3 orders of magnitude):

``` r
bench::mark(rlang::check_installed("later"), requireNamespace("later", quietly = TRUE), check = FALSE)
#> # A tibble: 2 × 6
#>   expression                             min median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                           <bch> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 "rlang::check_installed(\"later\")"  397µs  443µs     2189.    5.54MB     17.0
#> 2 "requireNamespace(\"later\", quietl… 558ns  613ns  1294407.        0B      0
```

<sup>Created on 2025-02-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

In purrr, I employ caching to ensure it's not run time every time a function is called e.g. https://github.com/tidyverse/purrr/blob/main/R/map.R#L214-L217 but I notice in other codebases it's often used unconditionally at the top of a function body (FYI @hadley).

Rather than require the user to implement these optimizations, it might be nice to handle this at source by short-cutting detection if the package is already loaded. This can be easily done using `.getNamespace()`, a really cheap function which returns either the namespace environment if already loaded or else `NULL`.

Note that this would be different to current behaviour as `check_installed()` actually checks if the package is installed on disk (which contributes to the slowness). Hence a question is: do we hide this behind an argument, or would it be cleaner to introduce another function e.g. `check_available()`?

hadley:
I think we could assume that if the package is loaded, then it's installed. 

It might just be that `check_installed()` could use a different method than `detect_installed()`; looking at the logic there I suspect we had to adapt it to the "partly" loaded case that you sometimes get when loading dependencies, and that distinction _might_ not matter for `check_installed()`. (But presumably matters elsewhere)

shikokuchuo:
> I think we could assume that if the package is loaded, then it's installed.

If you load the package e.g. call `pkg::function()`, then uninstall the package, the package is still loaded and usable, but technically no longer installed. I can't think of a situation where this distinction matters though.

[...]

PR #1777, opened 2025-02-13
Speed up `check_installed()` - faster check for whether a package is on disk

shikokuchuo:
As suggested by @hadley in https://github.com/r-lib/rlang/issues/1776#issuecomment-2654610545. This PR leaves that issue open for further improvements.

Additionally closes https://github.com/r-lib/rlang/issues/1608 by replacing the slow call to `system.file()`.

### DIFF
--- a/R/session.R
+++ b/R/session.R
@@ -101,10 +101,7 @@ detect_installed <- function(info) {
 }
 
 is_on_disk <- function(pkg) {
-  # A warning is emitted if the package was removed from disk
-  suppressWarnings(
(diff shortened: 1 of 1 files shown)


lionel- (approved):

lionel-:
This new implementation was causing failures in pkgdown checks because it didn't support packages loaded with pkgload. I fixed this here: https://github.com/r-lib/rlang/commit/830745e8798b7f2ef7c4afd96ed40d4adfe2fdf0

PR #1793, opened 2025-04-16
Don't memcpy a 0-length vector

MichaelChirico:
The same issue as here:

https://github.com/r-lib/vctrs/pull/1968

MRE traced back to:

```r
vctrs::vec_set_union(integer(), 12L)
```

I added that as a regression test in {vctrs}, if you know of a way to test that directly in {rlang} please advise.

r-devel thread:

https://stat.ethz.ch/pipermail/r-devel/2024-June/083456.html


DavisVaughan:
I vote for https://github.com/r-lib/rlang/pull/1797 instead!

Issue #393, opened 2018-01-27
`call_modify` fails on `arg = NULL` when `arg` is not in the call


vspinu:
On deletion, the lookup of non-existing args is not consistent with `list[["arg"]] <- NULL` or `list$arg <- NULL`:

```R
> call_modify(quote(foo(a = 1)), ... = NULL)
Error in expr[[nm]] <- args[[nm]] : subscript out of bounds
```

This makes using deletion programatically inconvenient, as an extra check for arg presence is required.

vspinu:
Conversely, how would you add `...` to a call consistently?

```R
> call_modify(quote(foo(a = 1)), quote(...)) # OK
foo(a = 1, ...)
> call_modify(quote(foo(a = 1, ...)), quote(...)) # Not OK
foo(a = 1, ..., ...)
```

Thanks.

lionel-:
> Conversely, how would you add ... to a call consistently?

I'm not sure, but I think it wouldn't feel right to special-case `...` here.

lionel-:
Hmm... Thinking more about it, it does make sense to special case it. In a way it is similar to a named argument.

PR #1795, opened 2025-04-20
news typos

musvaage:
the following was left as is

unconditionally would be more common in modern English

in `#393` had been used the wording non-existing args rather than unexisting arguments

```
$ grep -nr inconditionally rlang
rlang/NEWS.md:2201:Deprecated functions and arguments issue a warning inconditionally,
$ grep -nr unexisting rlang
rlang/NEWS.md:2052:* `call_modify()` now supports removing unexisting arguments (#393)
rlang/tests/testthat/test-call.R:209:test_that("can remove unexisting arguments (#393)", {
$

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -49,7 +49,7 @@
 
 * `obj_type_friendly()` now only displays the first class of S3 objects (#1622).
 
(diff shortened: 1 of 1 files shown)


lionel-:
Thanks!

PR #1802, opened 2025-05-29
Format with autobracing version of Air

DavisVaughan:
Everything looks good but there are two notes:
- A snapshot test that captured a function definition body had to be updated
- Some sensitive deparse related tests were marked with `# fmt: skip` to avoid reformatting them. We could update these tests to expect the new result if you'd like. I'm not sure if the exact form the test expects is meaningful or not.

rlang is a particularly low level package, so it's not totally surprising that running Air on rlang broke a few very sensitive tests. But it's worth keeping in mind that we can't 100% guarantee that you won't have to change your tests after using Air.

### DIFF
--- a/man/notes/handling-introspection.R
+++ b/man/notes/handling-introspection.R
@@ -36,7 +36,6 @@
 # currently no way of finding the setup frame with 100% reliability
 # (there might be several on the stack).
 
-
 f <- function() {
   throw()
 }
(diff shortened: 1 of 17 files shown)


man/notes/handling-introspection.R
```diff
@@ -73,7 +72,6 @@ foo(error = handle)
 #> >7.   └─global h(simpleError(msg, call))
 #>  8.     └─global handler_helper()
 
-
```
lionel-, line 76: oh yeah I gave up on this issue when I implemented double spacing... Double lines between comments are squeezed down to one.

lionel- (approved):
Looks good but can you please omit `man/` from the formatting and only do `R/` and `tests/`?

PR #1804, opened 2025-05-31
Speed up `is_installed()` - shortcut when package is already loaded

eitsupi:
Same as #1780

### DIFF
--- a/R/session.R
+++ b/R/session.R
@@ -63,9 +63,21 @@
 is_installed <- function(pkg, ..., version = NULL, compare = NULL) {
   check_dots_empty0(...)
 
+  # Fast path. Note that if `version` is supplied, this always falls through to the
+  # more expensive check.
+  if (is.null(version)) {
+    loaded <- lapply(pkg, function(x) {
+      is.character(x) && nzchar(x) && is.environment(.getNamespace(x))
+    })
+    if (all(as.logical(loaded))) {
+      return(TRUE)
+    }
+  }
+
   info <- pkg_version_info(pkg, version = version, compare = compare)
   all(detect_installed(info))
 }
+
 detect_installed <- function(info) {
   # Internal mechanism for unit tests
   hook <- peek_option("rlang:::is_installed_hook")


shikokuchuo (commented):
Thanks @eitsupi looks good to me!

R/session.R
```diff
@@ -63,9 +63,21 @@
...
+  # Fast path. Note that if `version` is supplied, this always falls through to the
+  # more expensive check.
+  if (is.null(version)) {
+    loaded <- lapply(pkg, function(x) {
+      is.character(x) && nzchar(x) && is.environment(.getNamespace(x))
+    })
+    if (all(as.logical(loaded))) {
+      return(TRUE)
+    }
+  }
```
MichaelChirico, line 75: Is this a typical intended interface? `is_installed(list("dplyr", "dbplyr"))`

If not, we can move the checks "outside the loop" toget early exits for any of the `!is.character(pkg)` and `!all(nzchar(pkg))` cases.

Next, why use `is.environment(.getNamespace(x))` instead of `isNamespaceLoaded(x)`? The former is described as an "internal" (hence the leading `.`), the latter is more readable (and nanosecond-scale-faster).

```suggestion
  if (is.null(version) && is.character(pkg) && all(nzchar(pkg)) && all(vapply(pkg, isNamespaceLoaded, NA))) {
    return(TRUE)
  }
```

shikokuchuo: `is_installed(c("dplyr", "dbplyr"))` is a typical call that we'd want to speed up.

The `is.character(x) && nzchar(x)` just ensures that the call to `is.environment(.getNamespace(x))` doesn't error out as that's not the intention here. Rather if the fast path fails, we run the full logic.

Thanks for the suggestion of `isNamespaceLoaded(x)`. It is indeed faster as it shares the same code path as `.getNamespace(x)` but just returns a logical value rather than the environment itself.

shikokuchuo: @MichaelChirico sorry somehow I missed your entire code suggestion, so my inital reply was totally off the mark (now amended above).

eitsupi: Would it be better to refactor this with `check_installed` after merging this since my intention of this PR is just to bring the same changes here as `check_installed` in #1780?

shikokuchuo: Yes that's fine thanks! If Lionel adopts the suggestion here, I will follow up with the corresponding change to `check_installed()`.

Issue #1806, opened 2025-06-11
last_trace() advice not helpful when missing {crayon} [aka odd mix of {crayon} & {cli}]


MichaelChirico:
I got an error and saw the expected advice:

```
Error in `FUN()`:
! `arg` must be a symbol
Run `rlang::last_trace()` to see where the error occurred.
```

However, the current environment doesn't have {crayon}, so that advice turned out to be less than helpful:

```r
rlang::last_trace()
# Error in loadNamespace(x) : there is no package called ‘crayon’
```

I peeked around and AFAICT ultimately this is the issue:

https://github.com/r-lib/rlang/blob/7371bac25ffe9b91097e0dad424cec5b2ee3c288/R/utils.R#L165-L166

I see three options:

 1. Use {cli} completely and avoid {crayon}
 2. Don't give the unhelpful advice if {crayon} is missing
 3. Add a `check_installed("crayon")` branch in `style_dim_soft()`.

This last is very easy and I'll file a quick PR, but I wonder if the others will be worth the higher investment.

PR #1807, opened 2025-06-11
sanity check for crayon being installed before using it

MichaelChirico:
See #1806 

I checked that the only other place using `crayon::` already has a similar check:

https://github.com/r-lib/rlang/blob/7371bac25ffe9b91097e0dad424cec5b2ee3c288/src/rlang/session.c#L20-L22


lionel-:
Thanks!

PR #1813, opened 2025-07-23
Update cnd.R: `is_error()` to also check for `try-error`

avhz:
`is_error()` to also check for `try-error` such that it can check for errors resulting from expressions like:

```R
try( 1 / "1" )
```

### Before:

```R
> rlang::is_error(try(1/"1"))
Error in 1/"1" : non-numeric argument to binary operator
[1] FALSE
```

### After:

```R
> rlang::is_error(try(1/"1"))
Error in 1/"1" : non-numeric argument to binary operator
[1] TRUE
```

### DIFF
--- a/R/cnd.R
+++ b/R/cnd.R
@@ -248,7 +248,7 @@ is_condition <- function(x) {
 #' @rdname is_condition
 #' @export
 is_error <- function(x) {
-  inherits(x, "error")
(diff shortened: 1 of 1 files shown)


lionel-:
Thank you for your proposal. I think the invariant of `is_error()` is that it only returns true for _condition objects_ of class `"error"`, so I can't accept the PR.

Issue #1684, opened 2024-01-23
is_syntactic_literal() returns TRUE in the presence of attributes.


moodymudskipper:
``` r
x <- 1
attr(x, "foo") <- 42
rlang::is_syntactic_literal(x)
#> [1] TRUE
```

<sup>Created on 2024-01-23 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I believe this contradicts the given definition, though the definition is a bit ambiguous :

> is_syntactic_literal() is a predicate that returns TRUE for the subset of literals that are created by R when parsing text (see [parse_expr()](http://127.0.0.1:9743/help/library/rlang/help/parse_expr)): numbers, strings and NULL. Along with symbols, these literals are the terminating nodes in an AST.

moodymudskipper:
it also uses `length()` which might dispatch on the class

PR #1826, opened 2025-09-19
More `is_syntactic_literal()` strictness

### LINKED ISSUE #1799: Negative numbers are not syntactic literals
Since they use `-`.

We dismiss complex numbers with a real part, presumably because of the `+`, so it would make sense to return FALSE for negative integer, double, and imaginary numbers.

``` r
x <- quote((-1))
y <- substitute((X), list(X = -1))
x
#> (-1)
y
#> (-1)
typeof(x[[2]])
#> [1] "language"
typeof(y[[2]])
#> [1] "double"
rlang::is_expression(x) # correct
#> [1] TRUE
rlang::is_expression(y) # incorrect
#> [1] TRUE
```

jonthegeek:
Reject negative numbers and restructure `is_syntactic_literal()` to hopefully make it easier to deal with any other strangeness we encounter.

Fixes #1799.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@
 * C code no longer calls `memcpy()` and `memset()` on 0-length R object memory
   (#1797).
 * `is_syntactic_literal()` returns `FALSE` for objects with attributes, such as `array(1)` or `factor("x")` (#1817, @jonthegeek).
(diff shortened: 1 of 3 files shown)


jonthegeek:
Also closes #1684

PR #1829, opened 2025-09-22
NA + 1i is not syntactic literal

### LINKED ISSUE #1828: NA + 1i is not a syntactic literal
Extracted from #1799.

`rlang::is_syntactic_literal(NA + 1i)` prints as `NA`, but it's actually `complex(real=NA, imaginary=1)` and thus `is_syntactic_literal(NA + 1i)` should return `FALSE`.

I'll add a PR for this shortly to wrap up the changes from Tidyverse Dev Day.

jonthegeek:
Fixes #1828.

### DIFF
--- a/R/expr.R
+++ b/R/expr.R
@@ -133,7 +133,12 @@ is_syntactic_literal <- function(x) {
       length(x) == 1 && is.null(attributes(x)) && (is.na(x) || x >= 0)
     },
     complex = {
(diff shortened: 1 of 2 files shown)


jonthegeek:
Evidently the NA + 1i behavior changed in R 4.4. I'll take a closer look when I get a minute.

PR #1832, opened 2025-09-30
Report array dimensionality in `obj_type_friendly()`

DavisVaughan:
In vctrs, we are trying to be fairly strict about the type of `condition` in `vec_if_else()`. In particular, we disallow arrays from being passed as `condition`, even 1D arrays (we do this consistently across `vec_if_else()`, `list_combine()`, vec_case_when()`, and `vec_replace_when()`, all of which can take logical "condition" indices).

https://github.com/r-lib/vctrs/blob/f27ab24c7eb0b1d78d48bd1d39175a1ecf7d4411/src/slice-assign.c#L964-L971

Unfortunately, `obj_type_friendly()` currently reports 1D arrays as identical to vectors resulting in the following not-useful error message

```r
     Error in `if_else()`:
     ! `condition` must be a logical vector, not a logical vector.
```

I'd argue that 1D arrays should be reported _as arrays_ by `obj_type_friendly()`. In other words, we'd have:

- `n_dim = 0` gives `vector`
- `n_dim = 1` gives `array`
- `n_dim = 2` gives `matrix`
- `n_dim = 3+` gives `array`

This feels right to me and allows us to report a useful distinction between vectors and 1D arrays.

### DIFF
--- a/R/standalone-obj-type.R
+++ b/R/standalone-obj-type.R
@@ -188,12 +188,12 @@ vec_type_friendly <- function(x, length = FALSE) {
   if (type == "list") {
     if (n_dim == 0) {
       return(add_length("a list"))
-    } else if (n_dim == 1) {
-      return("a list 1D array")
-    } else if (is.data.frame(x)) {
-      return("a data frame")
     } else if (n_dim == 2) {
-      return("a list matrix")
+      if (is.data.frame(x)) {
+        return("a data frame")
+      } else {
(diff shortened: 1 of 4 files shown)


lionel- (commented):
Since array vectors print the same way as vectors, it seems useful to distinguish them from arrays (which print differently) in a more explicit way. Some suggestions inside.

PR #1843, opened 2025-10-21
Require R >=4.0.0

DavisVaughan:
We are requiring R >=4.0.0 for a reason I will outline below. I took this chance to purge some dead code that switched on the R version, which we no longer need. It felt great to do so!

## Why require R >= 4.0.0?

The main reason we require R >=4.0.0 now is to never have to deal with `NAMEDNESS()` again, and instead fully embrace the reference counting semantics introduced in that version. Having to hold both mental models in our heads at once has been very difficult over the years.

In particular, https://github.com/r-lib/vctrs/pull/2076 hit a problem where the following is different on 3.6 vs 4.0:

```r
x <- list(a = 1, b = 2)
list2(!!!x)
```

On 3.6

```
.Internal(inspect(list2(!!!x)))> .Internal(inspect(list2(!!!x)))
@7f9fa8fb3c48 19 VECSXP g0c2 [NAM(7),ATT] (len=2, tl=0)
  @7f9fa8cefea8 14 REALSXP g0c1 [NAM(7)] (len=1, tl=0) 1
  @7f9fa8cefee0 14 REALSXP g0c1 [NAM(7)] (len=1, tl=0) 2
ATTRIB:
  @7f9fb8f91c78 02 LISTSXP g0c0 [] 
    TAG: @7f9fe9810100 01 SYMSXP g1c0 [MARK,NAM(7),LCK,gp=0x4000] "names" (has value)
    @7f9fa8fb3c88 16 STRSXP g0c2 [] (len=2, tl=0)
      @7f9fe9b016f8 09 CHARSXP g1c1 [MARK,gp=0x61] [ASCII] [cached] "a"
      @7f9fe9ef02c8 09 CHARSXP g1c1 [MARK,gp=0x61] [ASCII] [cached] "b"
```

On 4.0+

```
> x <- list(a = 1, b = 2)
> .Internal(inspect(list2(!!!x)))
@35489fb08 19 VECSXP g0c2 [REF(65535),ATT] (len=2, tl=0)
  @3568ad3c0 14 REALSXP g0c1 [REF(3)] (len=1, tl=0) 1
  @3568ad3f8 14 REALSXP g0c1 [REF(3)] (len=1, tl=0) 2
ATTRIB:
  @346754828 02 LISTSXP g0c0 [REF(1)] 
    TAG: @125831420 01 SYMSXP g1c0 [MARK,REF(65535),LCK,gp=0x4000] "names" (has value)
    @35489fac8 16 STRSXP g0c2 [REF(65535)] (len=2, tl=0)
      @130185300 09 CHARSXP g1c1 [MARK,REF(1609),gp=0x61,ATT] [ASCII] [cached] "a"
      @127937b08 09 CHARSXP g1c1 [MARK,REF(277),gp=0x61,ATT] [ASCII] [cached] "b"
```

The thing to note is that on 3.6 for the `names` attribute we get `STRSXP g0c2 []` but on 4.0+ we get `STRSXP g0c2 [REF(65535)]`. In other words, on 3.6 the `names` attribute does not look referenced by anyone, it looks completely "fresh" and mutable even though the result of `list2()` really owns it.

This is a problem because vctrs gets this list and thinks that it can mutate the names attribute in place, even though we check if we can ahead of time with `names = r_clone_referenced(names)`.

The reason this didn't work on 3.6 is that rlang uses `r_attrib_poke_names()`, which just calls `Rf_setAttrib()`. On 3.6, `Rf_setAttrib()` with a "fresh" `val` will let that value be assigned as an attribute without bumping its namedness.

- `setAttrib()` https://github.com/wch/r-source/blob/32190bec3f6417add7716be23c4eaed9bd40bc1b/src/main/attrib.c#L230
- `namesgets()` https://github.com/wch/r-source/blob/32190bec3f6417add7716be23c4eaed9bd40bc1b/src/main/attrib.c#L960
- `installAttrib()` https://github.com/wch/r-source/blob/32190bec3f6417add7716be23c4eaed9bd40bc1b/src/main/attrib.c#L352
- `cons()` https://github.com/wch/r-source/blob/78f3b069d6269ddde40d76515a1eef67df674435/src/main/memory.c#L2359
- `INCREMENT_REFCNT()` which was a no-op on 3.6 https://github.com/wch/r-source/blob/78f3b069d6269ddde40d76515a1eef67df674435/src/include/Rinternals.h#L356

Note that you'll see those `MAYBE_REFERENCED(val)` calls in there, but in this case `val` is _not_ referenced by anything yet, so those all return `false` and `val`'s namedness is never bumped.

On 4.0+, `INCREMENT_REFCNT()` will actually increment the refcount of `val`, making it look referenced, and vctrs's call to `r_clone_referenced()` will correctly clone it.

### DIFF
--- a/src/internal/attr.c
+++ b/src/internal/attr.c
@@ -240,11 +240,14 @@ r_obj* expr_vec_zap_srcref(r_obj* x) {
 
   attrib_zap_srcref(x);
 
+  // `VECTOR_PTR_RO()` (i.e. `r_list_cbegin()`) doesn't allow `EXPRSXP`, so we
+  // use rlang equivalents of `Rf_xlength()`, `VECTOR_ELT()`, and
+  // `SET_VECTOR_ELT()` instead
   r_ssize n = r_length(x);
-  r_obj* const * v_x = r_list_cbegin(x);
 
   for (r_ssize i = 0; i < n; ++i) {
-    r_list_poke(x, i, zap_srcref(v_x[i]));
+    r_obj* elt = r_list_get(x, i);
+    r_list_poke(x, i, zap_srcref(elt));
   }
 
   FREE(1);


tanho63:
curious: is [this](https://tidyverse.org/blog/2019/04/r-version-support/) still the best place to find docs on R version compatibility for tidyverse and r-lib? I notice it hasn't been consistently updated, but the doc seems to suggest that rlang could (or should?) depend on >=4.1 now

tanho63:
> It is, note how the news bullet here says rlang is still _more_ permissive than is required
> 
> ```
> * R >=4.0.0 is now required. This is still more permissive than the general tidyverse policy of supporting the [5 most recent versions of R](https://www.tidyverse.org/blog/2019/04/r-version-support/).
> ```

Thanks! (I 100% only read the PR body/comments and not the actual file changes :sweat_smile:)

PR #1855, opened 2025-11-16
fix embrace operator printing; add test

### LINKED ISSUE #1834: `rlang::as_label()` can't handle calls to `{{`
``` r
rlang::as_label(quote({{x}}))
#> Error in switch(type, call = "prefix", control = , delim = , subset = "special", : EXPR must be a length 1 vector

sessionInfo()
#> R version 4.3.3 (2024-02-29)
#> Platform: x86_64-pc-linux-gnu (64-bit)
#> Running under: Ubuntu 24.04.2 LTS
#> 
#> Matrix products: default
#> BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.12.0 
#> LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.12.0
#> 
#> locale:
#>  [1] LC_CTYPE=C.UTF-8       LC_NUMERIC=C           LC_TIME=C.UTF-8       
#>  [4] LC_COLLATE=C.UTF-8     LC_MONETARY=C.UTF-8    LC_MESSAGES=C.UTF-8   
#>  [7] LC_PAPER=C.UTF-8       LC_NAME=C              LC_ADDRESS=C          
#> [10] LC_TELEPHONE=C         LC_MEASUREMENT=C.UTF-8 LC_IDENTIFICATION=C   
#> 
#> time zone: America/New_York
#> tzcode source: system (glibc)
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> loaded via a namespace (and not attached):
#>  [1] digest_0.6.37     fastmap_1.2.0     xfun_0.52         glue_1.8.0       
#>  [5] knitr_1.50        htmltools_0.5.8.1 rmarkdown_2.29    lifecycle_1.0.4  
#>  [9] cli_3.6.5         reprex_2.1.1      withr_3.0.2       compiler_4.3.3   
#> [13] rstudioapi_0.17.1 tools_4.3.3       evaluate_1.0.5    yaml_2.3.10      
#> [17] rlang_1.1.6       fs_1.6.6
```
<sup>Created on 2025-10-01 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I believe the offending function is `rlang:::call_print_fine_type()`, which does not account for the possibility of `{{`. It returns the result of a `switch()` statement that does not anticipate `{{` and has no "default". Therefore, `rlang:::call_print_fine_type()` returns `NULL`, which is passed to another `switch()` within the calling function `rlang::call_print_type()`. This is where the error occurs, as seen above.

NikKrieger:
Fixes #1834 

`as_label()` used to error out if supplied an embraced symbol (e.g., `{{x}}`) because `call_print_fine_type()` did not anticipate `{{` as an option.

`call_print_fine_type()` now returns `"delim"` for calls to `{{` instead of null. The result is that `as_label()` will now always yield `{ ... }` if it is supplied an embraced symbol, which is what `as_label()` returns for all other calls to `{`. A test was added to ensure that this would be the result.

### DIFF
--- a/tests/testthat/test-deparse.R
+++ b/tests/testthat/test-deparse.R
@@ -740,7 +740,7 @@ test_that("as_label() supports special objects", {
   expect_identical(as_label(quo(foo)), "foo")
   expect_identical(as_label(quo(foo(!!quo(bar)))), "foo(bar)")
   expect_identical(as_label(~foo), "~foo")
-  expect_identical(as_label(quote({{foo}})), "{{foo}}")
+  expect_identical(as_label(call("{", quote({x}))), "{ ... }")
   expect_identical(as_label(NULL), "NULL")
 })
 


lionel-:
Thanks you!

PR #1857, opened 2025-12-10
Remove usage of `SET_OBJECT()`

lionel-:
Set class attibrute instead.

I've kept `r_mark_object()` for convenience for the case where an object is constructed by assigning a whole attribute structure at once that includes a class. What it now does is get the class and set it again, so that R can mark as object.


DavisVaughan:
Looks like we used `r_mark_object()` in vctrs in `new_data_frame()` after a `SET_ATTRIB()` which does include the `class`, so that's fine
https://github.com/r-lib/vctrs/blob/d0239de6c0140d807efcb0faab03bfcb6064a958/src/type-data-frame.c#L128-L129

Similar with dyn-array and splice box
https://github.com/r-lib/rlang/blob/0beac08e9d00a90e6bc38e46632c9b822c7e0edb/src/rlang/dyn-array.c#L14
https://github.com/r-lib/rlang/blob/0beac08e9d00a90e6bc38e46632c9b822c7e0edb/src/internal/dots.c#L57

PR #1871, opened 2026-02-04
Update environment binding API with planned ctors and accessors

lionel-:
So we can write algorithms that can be easily switched to the API discussed at https://bugs.r-project.org/show_bug.cgi?id=18928

### DIFF
--- a/src/rlang/env.c
+++ b/src/rlang/env.c
@@ -75,10 +75,11 @@ void r_env_coalesce(r_obj* env, r_obj* from) {
   }
 
   r_ssize n = r_length(syms);
+  r_obj* const * v_syms = r_list_cbegin(syms);
   enum r_env_binding_type* v_types = (enum r_env_binding_type*) r_int_begin(types);
 
   for (r_ssize i = 0; i < n; ++i) {
-    r_obj* sym = r_list_get(syms, i);
+    r_obj* sym = v_syms[i];
 
     if (r_env_has(env, sym)) {
       continue;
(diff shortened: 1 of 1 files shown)


src/internal/env-binding.c
DavisVaughan:
```suggestion
void env_poke_delayed(r_obj* env, r_obj* sym, r_obj* expr, r_obj* eval_env) {
```

?

src/rlang/env.c
```diff
@@ -65,56 +65,77 @@ r_obj* r_env_clone(r_obj* env, r_obj* parent) {
...
+        r_env_binding_delayed_env(from, sym)
+      );
+      break;
+
+    case R_ENV_BINDING_TYPE_forced:
+      r_env_bind_forced(
+        env,
+        sym,
+        r_env_binding_forced_expr(from, sym),
+        r_env_binding_forced_value(from, sym)
```
DavisVaughan: I guess this relies on the `_expr`, `_env`, and `_value` helpers never allocating, right? Because if `r_env_binding_delayed_expr()` and `r_env_binding_delayed_env()` allocate then we'd have a protection issue. But I guess we'd say that the promise is protecting them?

DavisVaughan (approved):
Looks like a nice API to me!

src/rlang/env-binding.c
```diff
@@ -95,3 +99,194 @@ r_obj* r_env_binding_types(r_obj* env, r_obj* bindings) {
...
+  r_obj* value = r_env_find(env, sym);
+
+  if (r_typeof(value) != R_TYPE_promise) {
+    r_abort("Not a promise binding.");
+  }
+  if (PRVALUE(value) == r_syms.unbound) {
+    r_abort("Not a forced binding.");
+  }
+
+  return PRVALUE(value);
```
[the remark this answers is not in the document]
lionel-: We have the promise already so doing a separate lookup might be surprising in other ways

src/rlang/env.c
lionel-: I think you're right, this relies on the assumption that the accessors do pure access and don't allocate. Otherwise rchk will flag us down.

PR #1872, opened 2026-02-05
Remove `r_env_find()` and `r_env_find_anywhere()`

lionel-:
Follow up to #1871

### DIFF
--- a/src/rlang/env.c
+++ b/src/rlang/env.c
@@ -90,25 +90,28 @@ void r_env_coalesce(r_obj* env, r_obj* from) {
       break;
 
     case R_ENV_BINDING_TYPE_value:
-      r_env_bind(env, sym, r_env_get(from, sym));
+      r_env_bind(env, sym, KEEP(r_env_get(from, sym)));
+      FREE(1);
       break;
 
     case R_ENV_BINDING_TYPE_delayed:
       r_env_bind_delayed(
         env,
         sym,
-        r_env_binding_delayed_expr(from, sym),
-        r_env_binding_delayed_env(from, sym)
+        KEEP(r_env_binding_delayed_expr(from, sym)),
+        KEEP(r_env_binding_delayed_env(from, sym))
       );
+      FREE(2);
       break;
 
     case R_ENV_BINDING_TYPE_forced:
       r_env_bind_forced(
         env,
         sym,
-        r_env_binding_forced_expr(from, sym),
-        r_env_get(from, sym)
+        KEEP(r_env_binding_forced_expr(from, sym)),
+        KEEP(r_env_get(from, sym))
       );
+      FREE(2);
       break;
 
     case R_ENV_BINDING_TYPE_missing:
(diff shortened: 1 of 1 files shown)


src/rlang/env-binding.c
```diff
@@ -160,28 +166,25 @@ void r_env_bind_delayed(r_obj* env, r_obj* sym, r_obj* expr, r_obj* eval_env) {
...
 #if RLANG_HAS_R_BINDING_API
   R_MakeForcedBinding(sym, expr, value, env);
 #else
-  // Creating an evaluated promise requires internal R API (`R_mkEVPROMISE`).
-  // Create a delayed binding and force it manually.
-  r_env_bind_delayed(env, sym, expr, r_envs.empty);
-
-  r_obj* promise = r_env_find(env, sym);
-  SET_PRVALUE(promise, value);
+  r_obj* promise = KEEP(Rf_allocSExp(PROMSXP));
```
DavisVaughan: Same comment about `KEEP(expr)`, and idk if we want to `KEEP(value)` too or not?

I don't remember what we made `R_MakeForcedBinding` did

DavisVaughan (approved):
A few notes from our offline conversation for future us:

- `r_env_get()` has always triggered active bindings, due to how `Rf_findVarInFrame3()` also triggers active bindings. This feels right to us. If you need to specially handle active, delayed, or forced bindings then you should switch on the binding type up front.

- An emerging pattern is moving from a single `r_env_find()` call to `r_env_has()` + `r_env_get()`

- It's unfortunate that `R_getVar()` errors on `R_MissingArg` bindings. This is a valid symbol to bind into an environment, and it feels like a shame that you can't pluck it out of there as well. It's certainly not "unbound". It feels like only `Rf_eval()` should care about erroring on `R_MissingArg`. If `R_getVar()` allowed `R_MissingArg`, then `r_env_get()` could have just used that directly, but without that it feels like we will never be able to make this switch.

src/rlang/env-binding.c
lionel-: I think an assignment operation can't protect two values at the same time because that would mean that the first temporary object would be unprotected while the second is created in:

```c
bind(x, foo(), bar())
```

src/rlang/env.c
```diff
@@ -137,7 +141,7 @@ void env_coalesce_plain(r_obj* env, r_obj* from, r_obj* syms) {
       continue;
     }
 
-    r_env_bind(env, sym, r_env_find(from, sym));
+    r_env_bind(env, sym, r_env_get(from, sym));
```
DavisVaughan: Bah, `KEEP()`? We did above with a similar line

lionel-:
We decided to embrace the explicit protection pattern when assigning.

Before:

```c
obj_set(x, new());
```

After:

```c
obj_set(x, KEEP(new());
FREE(1);
```

The need to free after the assignment is especially annoying, but there's no way to implement a fully general assignment principle without getting into rchk complications. And the principle breaks down if more than one object are assigned. And without the explicit protection, there's always a small doubt in the back of the head when reading code compelling you to double check what the assignment does.

PR #1876, opened 2026-03-06
Implement quosure capture on top of planned R core API for dots

lionel-:
This PR reimplements quosure capture on top of the API discussed with R core in https://gist.github.com/lionel-/1ebcbd5ec69c0775d514c329522408a3 and drafted in https://github.com/r-devel/r-svn/pull/209.

The implementation of the dots API will serve as polyfills for old versions once it makes it into R core. I've also added a bunch of tests to make sure the proposed implementation works as expected.

The main issue I found in the draft implementation is that dots accessors need to unwrap chains of promises created by forwarding `...` across layers of calls. This needs to happen both in the type getter and field getters. It's possible for an upstream promise to be forced, and both the type and field getters need to account for this.

The same bug actually exists in our proposed implementation for environment accessors we sent to https://bugs.r-project.org/show_bug.cgi?id=18928. Currently the accessors may return inner PROMSXP objects when accessing promise chains. The PR fixes the rlang polyfills implemented in https://github.com/r-lib/rlang/pull/1871 using the same promise unwrapping as for dots, and we should send an updated patch to the bugzilla.

I've ran revdeps for rlang, dplyr, tidyr, ggplot2, and purrr. No failures!

### DIFF
--- a/src/capture.c
+++ b/src/capture.c
@@ -60,13 +60,15 @@ r_obj* capture_delayed(r_obj* expr, r_obj* expr_env) {
             FREE(1);
             return result;
         }
-        case DOT_TYPE_delayed:
+        case DOT_TYPE_delayed: {
+            r_obj* new_env = r_env_dot_delayed_env(expr_env, dd - 1);
+            expr = r_env_dot_delayed_expr(expr_env, dd - 1);
+            expr_env = new_env;
             break;
         }
-
-        r_obj* new_env = r_env_dot_delayed_env(expr_env, dd - 1);
-        expr = r_env_dot_delayed_expr(expr_env, dd - 1);
-        expr_env = new_env;
+        default:
+            r_stop_unreachable();
+        }
     }
 
     MARK_NOT_MUTABLE(expr);
@@ -209,6 +211,9 @@ r_obj* capturedots(r_obj* frame) {
         case DOT_TYPE_delayed:
             dot = env_dot_delayed_capture(frame, i);
             break;
+
+        default:
+            r_stop_unreachable();
         }
 
         KEEP_AT(dot, dot_pi);


src/internal/dots-api.c
```diff
@@ -0,0 +1,248 @@
...
+        dots = CDR(dots);
+        if (dots == R_NilValue)
+            Rf_error("the ... list contains fewer than %d elements", i);
+    }
+
+    return CAR(dots);
+}
+
+// R API: R_DotsElt
+SEXP r_env_dot_get(SEXP env, int i) {
```
lionel-: This one felt better as a singular variant, so I wonder if we should update the name in the API proposal too.

DavisVaughan: Comparing to the gist, maybe

```
Rboolean R_DotsExist(SEXP env);
R_xlen_t R_DotsLength(SEXP env); // ...length()
SEXP R_DotsNames(SEXP env);  // ...names()
SEXP R_DotsElt(R_xlen_t i, SEXP env); // ...elt(i)

typedef enum {
    R_DotTypeValue = 0,     
    R_DotTypeMissing = 1, 
    R_DotTypeDelayed = 2,    
    R_DotTypeForced = 3    
} R_DotType;

R_DotType R_GetDotType(R_xlen_t i, SEXP env);

SEXP R_DotDelayedExpression(R_xlen_t i, SEXP env);
SEXP R_DotDelayedEnvironment(R_xlen_t i, SEXP env);

SEXP R_DotForcedExpression(R_xlen_t i, SEXP env);
```

Should become

```
Rboolean R_DotsExist(SEXP env);
R_xlen_t R_DotsLength(SEXP env); // ...length()
SEXP R_DotsNames(SEXP env);  // ...names()

typedef enum {
    R_DotTypeValue = 0,     
    R_DotTypeMissing = 1, 
    R_DotTypeDelayed = 2,    
    R_DotTypeForced = 3    
} R_DotType;

R_DotType R_DotType(R_xlen_t i, SEXP env);

SEXP R_Dot(R_xlen_t i, SEXP env); // ...elt(i)

SEXP R_DotDelayedExpression(R_xlen_t i, SEXP env);
SEXP R_DotDelayedEnvironment(R_xlen_t i, SEXP env);

SEXP R_DotForcedExpression(R_xlen_t i, SEXP env);
```

So concretely:

- `R_GetDotType` -> `R_DotType`
- `R_DotsElt` -> `R_Dot` (or `R_DotValue`?)

I absolutely see the appeal of `SEXP env` functions having a `Dots` prefix and `R_xlen_t i, SEXP env` functions having a `Dot` prefix.

src/rlang/env-binding.c
```diff
@@ -141,11 +167,17 @@ enum r_env_binding_type r_env_binding_type(r_obj* env, r_obj* sym) {
...
   if (r_typeof(value) == R_TYPE_promise) {
-    if (PRVALUE(value) == r_syms.unbound) {
-      return R_ENV_BINDING_TYPE_delayed;
-    } else {
+    if (PRVALUE(value) != r_syms.unbound) {
+      return R_ENV_BINDING_TYPE_forced;
+    }
+
+    bool forced;
+    delayed_promise_unwrap(value, &forced);
```
DavisVaughan: I think this would all be a bit more elegant if this function was just `promise_unwrap(value, &forced)` which didn't care about REQUIRING a delayed promise, just a promise.

So then this branch could drop the `PRVALUE(value) != r_syms.unbound` check entirely. That would just fall out in the `bool forced` result nicely.

tests/testthat/test-nse-defuse.R
```diff
@@ -701,6 +701,45 @@ test_that("can capture forced numbered dot", {
...
+  wrapper <- compiler::cmpfun(function() outer("hello"))
+  expect_equal(wrapper(), quo("hello"))
+
+  inner <- function(...) enquos(...)
+  outer <- function(...) inner(..1)
+  wrapper <- compiler::cmpfun(function() outer("hello"))
+  expect_equal(wrapper(), quos("hello"))
+})
+
+test_that("enquo() returns literal when ..N chain hits missing arg", {
```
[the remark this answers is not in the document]
lionel-: Not technically wrong, it means forced quosure, i.e. one that does not have an env. But we can also say it returns the missing arg here

src/capture.c
```diff
@@ -27,119 +30,194 @@ SEXP attribute_hidden new_captured_arg(SEXP x, SEXP env) {
...
     } else {
-	arg = findVar(sym, frame);
-	if (arg == R_UnboundValue)
-	    error(_("object '%s' not found"), CHAR(PRINTNAME(sym)));
+        SEXP found = r_env_until(frame, sym, R_EmptyEnv);
+
+        if (found == R_EmptyEnv)
+            error(_("object '%s' not found"), CHAR(PRINTNAME(sym)));
+
+        enum r_env_binding_type type = r_env_binding_type(found, sym);
```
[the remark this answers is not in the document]
lionel-: but we look up the type in the frame returned by `r_env_until()` here

src/internal/dots-info.h
```diff
@@ -0,0 +1,38 @@
...
+
+typedef enum {
+    DOT_TYPE_value = 0,
+    DOT_TYPE_missing = 1,
+    DOT_TYPE_delayed = 2,
+    DOT_TYPE_forced = 3
+} r_dot_type_t;
+
+Rboolean r_env_dots_exist(SEXP env);
+int r_env_dots_length(SEXP env);
```
DavisVaughan: Should be `r_ssize` return value for consistency with getters

src/internal/dots-info.h
```diff
@@ -0,0 +1,38 @@
...
+SEXP r_env_dots_names(SEXP env);
+SEXP r_env_dot_get(SEXP env, r_ssize i);
+
+r_dot_type_t r_env_dot_type(SEXP env, r_ssize i);
+
+SEXP r_env_dot_delayed_expr(SEXP env, r_ssize i);
+SEXP r_env_dot_delayed_env(SEXP env, r_ssize i);
+SEXP r_env_dot_forced_expr(SEXP env, r_ssize i);
```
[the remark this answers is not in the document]
lionel-: I'll do another pass to switch to fully switch to rlang types in dots-info and capture.c

PR #1895, opened 2026-04-16
Stable `hash()`

### LINKED ISSUE #1681: `hash` generates different results on identical objects (even with same memory address)
I thought `hash` is supposed to generate the same results for identical objects. Could you help me with the following cases?

``` r
options(keep.source = TRUE)
a <-   function(){}
rlang::hash(a)
#> [1] "eca7e650f5be54ba7c122fbc88ed0811"
a <- function(){}
rlang::hash(a)
#> [1] "7bcbcbd5583248d94607c79bab4b70f0"
a()
#> NULL
rlang::hash(a)
#> [1] "a70206d89b1b3cc96363d3413aea8ed6"
a()
#> NULL
rlang::hash(a)
#> [1] "c3b887fbb758842048691f04406c130c"
```

<sup>Created on 2024-01-16 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Also 

``` r
memF <- memoise::memoise(function(f){ f() })

a <- function(){
  message("a is evaluated")
}

memF(a)
#> a is evaluated
memF(a)
#> a is evaluated
memF(a)
#> a is evaluated
memF(a)
```

<sup>Created on 2024-01-16 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

lionel-:
Branched from #1894 

Closes #1681

### DIFF
--- a/R/hash.R
+++ b/R/hash.R
@@ -13,6 +13,11 @@
 #' within a session. Closures hash their formals, body, and environment
 #' identity.
 #'
+#' By default, source references are stripped before hashing so that
+#' closures and calls that are textually identical produce the same
+#' hash regardless of where they were parsed. Set `zap_srcref` to
+#' `FALSE` to include source references in the hash.
+#'
 #' @details
 #' These hashers use the XXH128 hash algorithm of the xxHash library, which
 #' generates a 128-bit hash. Both are implemented as streaming hashes, which
@@ -24,6 +29,11 @@
 #' representation details (e.g. ALTREP compact forms, the growable vector bit).
 #'
 #' @param x An object.
+#'
+#' @param zap_srcref Whether to ignore source references when hashing
+#'   (default `TRUE`). Source references depend on parse location, so
+#'   including them makes hashes of closures and calls
+#'   non-reproducible across sessions.
 #'
 #' @param path A character vector of paths to the files to be hashed.
 #'
@@ -45,8 +55,8 @@
 #' # If you need a single hash for multiple files,
 #' # hash the result of `hash_file()`
 #' hash(hashes)
-hash <- function(x) {
-  .Call(ffi_hash, x)
+hash <- function(x, zap_srcref = TRUE) {
+  .Call(ffi_hash, x, zap_srcref)
 }
 
 # Keep this alias for a while
--- a/R/hash.R
+++ b/R/hash.R
@@ -5,17 +5,23 @@
 #'
 #' - `hash_file()` hashes the data contained in a file.
 #'
-#' The generated hash is guaranteed to be reproducible across platforms that
-#' have the same endianness and are using the same R version.
+#' For ordinary data objects (vectors, lists, data frames, etc.), the generated
+#' hash is reproducible across sessions and R versions on platforms that have
+#' the same endianness. However, hash values may change between rlang versions,
+#' although that should be rare. Reference-like objects (environments, external
+#' pointers, builtins) are hashed by identity, so their hashes are only stable
+#' within a session. Closures hash their formals, body, and environment
+#' identity.
 #'
 #' @details
 #' These hashers use the XXH128 hash algorithm of the xxHash library, which
 #' generates a 128-bit hash. Both are implemented as streaming hashes, which
 #' generate the hash with minimal extra memory usage.
 #'
-#' For `hash()`, objects are converted to binary using R's native serialization
-#' tools. Serialization version 3 is used. See [serialize()] for more
-#' information about the serialization version.
+#' For `hash()`, a custom object walker feeds the object's type, length, data
+#' bytes, and attributes directly into the hash algorithm. This avoids
+#' dependency on R's serialization format, making the hash immune to internal
+#' representation details (e.g. ALTREP compact forms, the growable vector bit).
 #'
 #' @param x An object.
 #'
--- a/R/hash.R
+++ b/R/hash.R
@@ -11,7 +11,8 @@
 #' although that should be rare. Reference-like objects (environments, external
 #' pointers, builtins) are hashed by identity, so their hashes are only stable
 #' within a session. Closures hash their formals, body, and environment
-#' identity.
+#' identity. Byte-compiled and uncompiled closures hash identically
+#' because the body is always hashed from the original language tree.
 #'
 #' By default, source references are stripped before hashing so that
 #' closures and calls that are textually identical produce the same
--- a/src/internal/hash.c
+++ b/src/internal/hash.c
@@ -188,9 +188,9 @@ static void hash_feed_vector_data(
         break;
 
     case R_TYPE_double: {
-        const double* p_x = r_dbl_cbegin(x);
+        const double* v_x = r_dbl_cbegin(x);
         for (r_ssize i = 0; i < n; ++i) {
-            double val = hash_normalise_dbl(p_x[i]);
+            double val = hash_normalise_dbl(v_x[i]);
             hash_feed(ctx->p_state, &val, sizeof(double));
         }
         break;
@@ -229,7 +229,14 @@ static void hash_feed_vector_data(
         break;
     }
 
-    case R_TYPE_list:
+    case R_TYPE_list: {
+        r_obj* const* v_x = r_list_cbegin(x);
+        for (r_ssize i = 0; i < n; ++i) {
+            hash_object(ctx, v_x[i]);
+        }
+        break;
+    }
+
     case R_TYPE_expression:
         for (r_ssize i = 0; i < n; ++i) {
             hash_object(ctx, r_list_get(x, i));
--- a/src/internal/hash.c
+++ b/src/internal/hash.c
@@ -13,6 +13,16 @@
 #include <stdio.h>    // sprintf()
 #include <inttypes.h> // PRIx64
 
+struct hash_ctx {
+    XXH3_state_t* p_state;
+    bool zap_srcref;
+};
+
+static inline bool is_srcref_tag(r_obj* tag) {
+    return tag == r_syms.srcref || tag == r_syms.srcfile ||
+        tag == r_syms.wholeSrcref;
+}
+
 #include "decl/hash-decl.h"
 
 // Finalize the hash state into a 128-bit digest and format it as a
@@ -50,10 +60,12 @@ static inline void hash_feed_ptr(XXH3_state_t* p_state, const void* ptr) {
     hash_feed(p_state, &ptr, sizeof(ptr));
 }
 
+// -----------------------------------------------------------------------------
+
 // Feed the type tag, then user-visible data, then attributes
-static void hash_object(XXH3_state_t* p_state, r_obj* x) {
+static void hash_object(struct hash_ctx* ctx, r_obj* x) {
     int type = r_typeof(x);
-    hash_feed_int(p_state, type);
+    hash_feed_int(ctx->p_state, type);
 
     switch (type) {
     case R_TYPE_null:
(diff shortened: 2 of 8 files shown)


tests/testthat/test-hash.R
```diff
@@ -1,8 +1,289 @@
-test_that("simple hashes with no ALTREP and no attributes are reproducible", {
```
DavisVaughan, line 1: since ALTREP was an explicit weakness before, maybe a simple explicit test that it works with ALTREP objects stably?

```
test_that("hashes are stable across R versions with ALTREP objects", {
  skip_if_big_endian()
  expect_snapshot(hash(1:5))
})
```

R/hash.R
```diff
@@ -5,20 +5,37 @@
...
+#' bytes, and attributes directly into the hash algorithm. This avoids
+#' dependency on R's serialization format, making the hash immune to internal
+#' representation details (e.g. ALTREP compact forms, the growable vector bit).
 #'
 #' @param x An object.
 #'
+#' @param zap_srcref Whether to ignore source references when hashing
+#'   (default `TRUE`). Source references depend on parse location, so
+#'   including them makes hashes of closures and calls
+#'   non-reproducible across sessions.
```
DavisVaughan, line 37: I would feel comfortable not exposing this in any way

src/internal/hash.c
```diff
@@ -13,205 +13,289 @@
...
+    case R_TYPE_logical:
+    case R_TYPE_integer:
+    case R_TYPE_double:
+    case R_TYPE_complex:
+    case R_TYPE_raw:
+    case R_TYPE_character:
+    case R_TYPE_list:
+    case R_TYPE_expression: {
+        r_ssize n = r_length(x);
+        hash_feed_ssize(ctx->p_state, n);
```
DavisVaughan, line 89: Why include the size?

I get including the type so INTSXP and LGLSXP look different, but why size?

src/internal/hash.c
```diff
@@ -13,205 +13,289 @@
...
+        hash_feed_vector_data(ctx, x, type, n);
+        break;
```
DavisVaughan, line 91: Speaking of, we should probably have a test that `hash(1L)` and `hash(TRUE)` are different, if we dont already

src/internal/hash.c
```diff
@@ -13,205 +13,289 @@
...
+        // When zapping srcrefs, stop after the 3rd node (formals, body, env).
+        bool is_fn_call = ctx->zap_srcref && type == R_TYPE_call &&
+            r_node_car(x) == r_syms.function;
+        r_ssize n = r_length(x);
+        if (is_fn_call && n > 3) {
+            n = 3;
+        }
+        hash_feed_ssize(ctx->p_state, n);
+        r_ssize i = 0;
+        for (r_obj* node = x; node != r_null && i < n;
```
DavisVaughan: i dont think you need both `node != r_null && i < n` since you got `n` from length?

src/internal/hash.c
```diff
@@ -13,205 +13,289 @@
...
+        // so compiled and uncompiled closures hash identically.
+        hash_object(ctx, r_fn_body(x));
+        hash_feed_ptr(ctx->p_state, (const void*) r_fn_env(x));
+        break;
+    }
 
-    struct R_outpstream_st stream;
+    // S4 slots are stored as attributes, which are walked below
+    case R_TYPE_s4:
+        break;
```
DavisVaughan: hmmmmmm, even ones that wrap a vector that is accessed via `@.Data`? i dont think so

```r
lubridate::seconds(1:5)
```

this wraps an actual integer vector and treats that as "seconds" and then "hours", "minutes", etc are stored in the slots. You access the "seconds" via `@.Data` but i dont think its an attribute.

so i think for lubridate objects youd miss any seconds in the hash

src/internal/hash.c
```diff
@@ -13,205 +13,289 @@
...
+    case R_TYPE_builtin:
+    case R_TYPE_special:
+    case R_TYPE_pointer:
+        hash_feed_ptr(ctx->p_state, (const void*) x);
+        break;
 
-    r_obj* value = KEEP(hash_value(p_xx_state));
-    r_obj* out = r_str_as_character(value);
+    default:
+        hash_feed_ptr(ctx->p_state, (const void*) x);
```
DavisVaughan: i feel like its better to have "hash not implemented" as an error here, which i think is what we do in vctrs.

it would certainly make me feel more comfortable about it, and immediately surface types we dont support but should

src/internal/hash.c
```diff
@@ -13,205 +13,289 @@
...
+            hash_feed(ctx->p_state, &re, sizeof(double));
+            hash_feed(ctx->p_state, &im, sizeof(double));
+        }
+        break;
+    }
 
-    // 32 for hash, 1 for terminating null added by `snprintf()`
-    char out[32 + 1];
+    case R_TYPE_raw:
+        hash_feed(ctx->p_state, r_raw_cbegin(x), (size_t) n);
```
DavisVaughan, line 213: To be super safe should probably be `n * sizeof(Rbyte)`?

src/internal/hash.c
```diff
@@ -13,205 +13,289 @@
...
+    r_ssize n
+) {
+    switch (type) {
+    case R_TYPE_logical:
+    case R_TYPE_integer:
+        hash_feed(ctx->p_state, r_vec_cbegin(x), (size_t) n * sizeof(int));
+        break;
+
+    case R_TYPE_double: {
+        const double* p_x = r_dbl_cbegin(x);
```
DavisVaughan: `p_x` vs our normal `v_x`?

src/internal/hash.c
```diff
@@ -13,205 +13,289 @@
...
+            if (elt == NA_STRING) {
+                hash_feed_int(ctx->p_state, 1);
+            } else {
+                hash_feed_int(ctx->p_state, 0);
+                int enc = (int) Rf_getCharCE(elt);
+                int len = LENGTH(elt);
+                hash_feed_int(ctx->p_state, enc);
+                hash_feed_int(ctx->p_state, len);
+                hash_feed(ctx->p_state, r_str_c_string(elt), (size_t) len);
+            }
```
DavisVaughan, line 229: Is this to differentiate `NA_STRING` from `""`?

Do we have a test for that?

src/internal/hash.c
```diff
@@ -13,205 +13,289 @@
...
+        for (r_ssize i = 0; i < n; ++i) {
+            r_obj* elt = p_x[i];
+            if (elt == NA_STRING) {
+                hash_feed_int(ctx->p_state, 1);
+            } else {
+                hash_feed_int(ctx->p_state, 0);
+                int enc = (int) Rf_getCharCE(elt);
+                int len = LENGTH(elt);
+                hash_feed_int(ctx->p_state, enc);
+                hash_feed_int(ctx->p_state, len);
```
DavisVaughan, line 227:
```suggestion
                r_ssize len = r_length(elt);
                hash_feed_int(ctx->p_state, enc);
                hash_feed_ssize(ctx->p_state, len);
```

probably best to be consistent? i know CHARSXP cant be long vectors right now, but still

src/internal/hash.c
```diff
@@ -13,205 +13,289 @@
...
+            }
+        }
+        break;
+    }
 
-    snprintf(out, sizeof(out), "%016" PRIx64 "%016" PRIx64, high, low);
+    case R_TYPE_list:
+    case R_TYPE_expression:
+        for (r_ssize i = 0; i < n; ++i) {
+            hash_object(ctx, r_list_get(x, i));
```
DavisVaughan, line 244: could use `r_list_cbegin()` here for iteration

DavisVaughan: but youd have to split the list vs expression case to be able to do so (expressions dont support cbegin)

still worth it imo

tests/testthat/test-hash.R
```diff
@@ -1,8 +1,289 @@
...
+    hash(list(2))
+
+    hash(quote(x))
+    hash(quote(y))
+  })
+})
+
+test_that("different types produce different hashes", {
+  expect_false(hash(1L) == hash(1))
+  expect_false(hash(TRUE) == hash(1L))
```
DavisVaughan, line 69: ✅

tests/testthat/test-hash.R
```diff
@@ -1,8 +1,289 @@
...
+  x2 <- parse(text = "1 + 2; 3 + 4", keep.source = TRUE)
+  expect_true("srcref" %in% names(attributes(x1)))
+  expect_identical(hash(x1), hash(x2))
+  expect_false(hash(x1, zap_srcref = FALSE) == hash(x2, zap_srcref = FALSE))
+})
+
+test_that("NA_character_ hashes distinctly from the string \"NA\"", {
+  expect_false(hash(NA_character_) == hash("NA"))
+  expect_false(hash(NA_character_) == hash(""))
+})
```
DavisVaughan, line 215: ✅

tests/testthat/test-hash.R
```diff
@@ -1,8 +1,289 @@
...
+
+test_that("S4 objects hash structurally, not by identity", {
+  on.exit(removeClass("HashTestS4"))
+  setClass("HashTestS4", slots = list(x = "numeric", y = "character"))
+  a <- new("HashTestS4", x = 1, y = "hello")
+  b <- new("HashTestS4", x = 1, y = "hello")
+  c <- new("HashTestS4", x = 2, y = "hello")
+  expect_identical(hash(a), hash(b))
+  expect_false(hash(a) == hash(c))
+})
```
DavisVaughan, line 255: i think you want a test like

```
> Foo <- setClass("Foo", contains = "numeric")
> Foo(1)
An object of class "Foo"
[1] 1
> unclass(Foo(1))
[1] 1
```

tests/testthat/test-hash.R
```diff
@@ -1,8 +1,289 @@
...
+
+test_that("resizable vectors hash the same as regular vectors (#1681)", {
+  skip_if_not_installed("vctrs")
+  expect_identical(hash(vctrs::vec_slice(1:3, 1:3)), hash(1:3))
+  expect_identical(hash(vctrs::vec_c(1L, 2L, 3L)), hash(1:3))
+
+  x <- c("a", "b", "c")
+  expect_identical(hash(vctrs::vec_slice(x, 1:3)), hash(x))
```
DavisVaughan, line 278: im not the biggest fan of making rlang rely on vctrs in this way in case something in vctrs changes. is there an rlang only way to do this? maybe its not easy

DavisVaughan (approved):

src/internal/hash.c
lionel-: To bound the object, otherwise this introduces ambiguity with the next object we hash if it matches a type hash?

src/internal/hash.c
lionel-: I think `Rbyte` can't legally be different than `char`

src/internal/hash.c
lionel-: Agreed, worth it

src/internal/hash.c
lionel-: yep

tests/testthat/test-hash.R
lionel-: It seemed like an invariant to uphold?

R/hash.R
lionel-: rlang is sufficiently low-level that I'm comfortable exposing it

tests/testthat/test-hash.R
DavisVaughan: sure

src/internal/hash.c
lionel-: good catch!

lionel-: oops nope, this kind of S4 objects doesn't have type S4SXP.

What we should do though is hash the S4 bit, so that these vector-like S4 objects hash differently than regular vectors mimicking the S4 shape (same class and data).

lionel-:
```c
    int type = r_typeof(x);
    hash_feed_int(ctx->p_state, type);

    // Vector-like S4 objects only differ from regular objects by the S4 bit. We
    // hash it here to disambiguate. This can be extended to a bitfield if needed.
    hash_feed_int(ctx->p_state, Rf_isS4(x));
```

src/internal/hash.c
lionel-: It would feel a bit wasteful to feed ssize instead of int for each element of a character vector

PR #1896, opened 2026-04-16
Fix double evaluation of active bindings in `env_get()` on old R

### LINKED ISSUE #1893: `rlang` 1.2.0: `env_get()` evaluates active bindings twice in R <= 4.4
This can lead to duplicated messages or warnings. Reprex:

```r
getRversion()
packageVersion("rlang")

e <- new.env()
makeActiveBinding(
  "x",
  function() {
    message("forced")
    1
  },
  e
)

rlang::env_get(e, "x")
get("x", envir = e, inherits = FALSE)
```

#### R 4.4, `rlang` 1.1.7: works

``` r
getRversion()
#> [1] '4.4.1'
packageVersion("rlang")
#> [1] '1.1.7'

rlang::env_get(e, "x")
#> forced
#> [1] 1
get("x", envir = e, inherits = FALSE)
#> forced
#> [1] 1
```

#### R 4.4, `rlang` 1.2.0: duplicated message
``` r
getRversion()
#> [1] '4.4.1'
packageVersion("rlang")
#> [1] '1.2.0'

rlang::env_get(e, "x")
#> forced
#> forced
#> [1] 1
get("x", envir = e, inherits = FALSE)
#> forced
#> [1] 1
```

#### R 4.5, `rlang` 1.1.7: works

``` r
getRversion()
#> [1] '4.5.0'
packageVersion("rlang")
#> [1] '1.1.7'

rlang::env_get(e, "x", inherit = FALSE)
#> forced
#> [1] 1
get("x", envir = e, inherits = FALSE)
#> forced
#> [1] 1
```

#### R 4.5, `rlang` 1.2.0: works
``` r
getRversion()
#> [1] '4.5.0'
packageVersion("rlang")
#> [1] '1.2.0'

rlang::env_get(e, "x", inherit = FALSE)
#> forced
#> [1] 1
get("x", envir = e, inherits = FALSE)
#> forced
#> [1] 1
```

lionel-:
Closes #1893

### DIFF
--- a/src/rlang/env-binding.c
+++ b/src/rlang/env-binding.c
@@ -186,16 +186,11 @@ r_obj* r_env_get(r_obj* env, r_obj* sym) {
 #if R_VERSION >= R_Version(4, 5, 0)
     return R_getVar(sym, env, FALSE);
 #else
-    // `Rf_findVarInFrame3()` evaluates active bindings via `BINDING_VALUE()`
-    // on R < 4.5, so skip `env_find()` to avoid evaluating them twice (#1893)
-    if (type != R_ENV_BINDING_TYPE_active) {
-        r_obj* value = env_find(env, sym);
-        if (r_typeof(value) == R_TYPE_dots) {
-            return value;
-        }
+    // `...` bindings must be returned as-is without `Rf_eval()`
+    if (sym == r_syms.dots) {
+        return env_find(env, sym);
     }
 
-    // Handles value, delayed, forced, and active bindings
     return Rf_eval(sym, env);
 #endif
 }
--- a/src/rlang/env-binding.c
+++ b/src/rlang/env-binding.c
@@ -186,9 +186,13 @@ r_obj* r_env_get(r_obj* env, r_obj* sym) {
 #if R_VERSION >= R_Version(4, 5, 0)
     return R_getVar(sym, env, FALSE);
 #else
-    r_obj* value = env_find(env, sym);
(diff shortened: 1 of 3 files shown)


src/rlang/env-binding.c
```diff
@@ -185,9 +185,13 @@ r_obj* r_env_get(r_obj* env, r_obj* sym) {
...
-  r_obj* value = env_find(env, sym);
-  if (r_typeof(value) == R_TYPE_dots) {
-    return value;
+  // `Rf_findVarInFrame3()` evaluates active bindings via `BINDING_VALUE()`
+  // on R < 4.5, so skip `env_find()` to avoid evaluating them twice (#1893)
+  if (type != R_ENV_BINDING_TYPE_active) {
+    r_obj* value = env_find(env, sym);
+    if (r_typeof(value) == R_TYPE_dots) {
+      return value;
+    }
```
DavisVaughan: Would it also be better to do a pointer comparison of `sym` against `Rf_install("...")` to see if we even need to do the `env_find()`? Or maybe you can have a DOTSXP in an env even if the name isn't `...` so that might not catch every case?

I'm just mildly annoyed from a performance perspective that you have to kind of "get" the thing twice, but i guess its also for R < 4.5 so its probably fine as is i suppose...

DavisVaughan (approved):

src/rlang/env-binding.c
lionel-: oh good idea!

PR #1912, opened 2026-08-05
protect rotated ast root during injection

### LINKED ISSUE #1910: use-after-free in `!!` injection during AST rotation
`expr()` (and other injection entry points) can return corrupted results — or in principle crash — when an expression requires the operator-precedence fixup (AST rotation) in `src/internal/ast-rotate.c` and a GC runs while a later `!!` operand is evaluated.

## Reproducible example

Reproduces deterministically on CRAN rlang 1.3.0 and on `main` (R 4.6.1, Linux). The `gc()` inside `f()` runs exactly in the unprotected window; the `pairlist()` churn then forces reuse of the freed cons cells:

``` r
f <- function() {
  gc()
  for (i in 1:200000) pairlist(i)
  2
}
a <- 1
res <- rlang::expr(1 + !!a + !!f())
deparse(res)
#> [1] "<environment>"     # expected "1 + 1 + 2"; also seen: pairlist(NULL), etc.
identical(res, quote(1 + 1 + 2))
#> [1] FALSE
```

Any GC triggered inside the evaluation of the second `!!` operand (e.g. a large allocation in user code) puts you in the same state; the explicit `gc()` just makes it deterministic.

## Analysis

The only protection for the expression during injection is the `KEEP()` on the original root (e.g. `ffi_interp()` in `src/internal/nse-inject.c:426`); `ast-rotate.c` contains no `KEEP` at all. In `maybe_rotate()`'s rotation branch:

- `ast-rotate.c:302` detaches the upper pivot from the protected tree (`r_node_poke_car(info->lower_root, r_node_cadr(info->lower_pivot))`).
- `ast-rotate.c:303` makes the old (protected) root a *child* of the pivot. Protection is transitive downward only, so the new root and every not-yet-expanded RHS operand hanging off it are now unreachable from any GC root — they're referenced only by C locals.
- `ast-rotate.c:319` then recurses via `node_list_interp_fixup()`, which for a remaining `!!` operand reaches `bang_bang()` → `r_eval()` (`nse-inject.c:281`), i.e. arbitrary user code, in that unprotected state.

A GC there collects the new root's cons cells and the `!!` subtree in active use; the subsequent `r_node_poke_car()` writes into a freed cell and the dangling pointer is returned to R.

Verified empirically with an instrumented build: an `R_gc()` inserted between the rotation and the recursion frees the new root's cell (the allocator hands the same address back ~15k cons allocations later).

## Suggested fix

Protect the current root across the recursion in `maybe_rotate()`:

``` c
KEEP(op);
node_list_interp_fixup(op, NULL, env, info, false);
r_obj* out = maybe_rotate(op, env, info);
FREE(1);
return out;
```

With this change the reproducible example returns `1 + 1 + 2` intact and `test-nse-inject.R` passes. The other rotation branch (`ast-rotate.c:307-311`) is already safe: it reattaches the pivot into the protected tree before recursing. Happy to submit a PR.

kevinushey:
Fixes #1910.

After a left rotation in `maybe_rotate()` (`src/internal/ast-rotate.c`), the new expression root is referenced only from a C local: the protect stack pins the old root, which has just become a *child* of the new one, and protection is only transitive downwards. The recursion that follows evaluates remaining `!!` operands via `r_eval()`, so a GC triggered by user code collects the new root's cons cells and the not-yet-expanded operands. See the issue for a deterministic reproducible example on CRAN rlang 1.3.0 and full analysis.

This PR protects the new root across the recursion. Each recursion level pins its current root, and the old root plus the unexpanded RHS chain are reachable from it. The second rotation branch is unchanged: it reattaches the pivot into the protected tree before recursing.

The regression test interpolates before calling `expect_identical()`: with the injection inline in the expectation, the interpolation would be performed by the rlang imported by testthat, which in `load_all()` sessions can be a different (installed) rlang than the package under test.

Verified: the new test fails against unfixed rlang 1.3.0 and passes with the fix; all other `test-nse-inject.R` expectations pass; an instrumented build confirms the rotated root survives a forced `R_gc()` in the previously unprotected window.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # rlang (development version)
 
+* Fixed a use-after-free that could corrupt the result of `expr()` and other injection functions when `!!` injection required an operator precedence fixup of the AST and a garbage collection occurred during injection (#1910).
+
 * Fixed backtrace links for `file://` URLs in srcrefs.
 
 * Backtraces no longer emit file hyperlinks for source files that don't exist on disk, such as the `R CMD INSTALL` staging directory recorded when a package is installed kept srcrefs. The location is shown as plain text instead (#1908).
--- a/src/internal/ast-rotate.c
+++ b/src/internal/ast-rotate.c
@@ -315,9 +315,17 @@ static r_obj* maybe_rotate(
     // reuse it in the recursion
     initialise_rotation_info(info);
 
+    // After a rotation `op` is the detached upper pivot whose only
+    // reference is this C local, so protect it while the recursion
+    // below evaluates `!!` operands (#1910)
+    KEEP(op);
+
     // Recurse on the RHS of the upper pivot (which is now the new root)
     node_list_interp_fixup(op, NULL, env, info, false);
-    return maybe_rotate(op, env, info);
+    r_obj* out = maybe_rotate(op, env, info);
+
+    FREE(1);
+    return out;
 }
 
 /**
--- a/tests/testthat/test-nse-inject.R
+++ b/tests/testthat/test-nse-inject.R
@@ -148,6 +148,20 @@ test_that("lower pivot is correctly found (#1125)", {
   expect_equal_(expr(1 + !!2 * 3 * 4 + 5), expr(1 + 2 * 3 * 4 + 5))
 })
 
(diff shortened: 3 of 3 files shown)


lionel-:
Thanks!

PR #1913, opened 2026-08-05
bump iterator state before pushing on sexp stack

### LINKED ISSUE #1911: sexp iterator revisits nodes after stack resize (stale pointer)
`r_sexp_next()` in `src/rlang/walk.c` misbehaves once traversal depth exceeds `SEXP_STACK_INIT_SIZE` (256): part of the structure is visited twice. The cause is a write through a dangling pointer into the traversal stack's pre-resize buffer.

This code is only compiled under `RLANG_USE_PRIVATE_ACCESSORS` (its sole caller is `ffi_sexp_iterate()`), so shipped builds are unaffected — but `walk.h` is part of the librlang API. Relatedly, `walk.c` no longer compiles on R >= 4.5 even with the flag, since the declarations for `FRAME`, `HASHTAB`, `PRENV`, `PREXPR`, `PRVALUE`, `EXTPTR_PROT`, and `EXTPTR_TAG` were removed from R's API headers.

## Reproducible example

Requires a dev build with `PKG_CPPFLAGS += -DRLANG_USE_PRIVATE_ACCESSORS` in `src/Makevars` (on R >= 4.5, plus declarations/stubs for the accessors above; a nested-list traversal never reaches them).

``` r
count_visits <- function(depth) {
  x <- 1
  for (i in 1:depth) x <- list(x)
  visits <- 0L
  addrs <- character()
  invisible(rlang:::sexp_iterate(x, function(...) {
    args <- list(...)
    visits <<- visits + 1L
    if (args$dir == "incoming") addrs <<- c(addrs, args$addr)
    NULL
  }))
  c(visits = visits, dup_incoming = sum(duplicated(addrs)))
}
count_visits(250)  # visits: 501, dup_incoming: 0   (2n + 1, correct)
count_visits(300)  # visits: 690, dup_incoming: 44  (expected 601, 0)
```

## Analysis

- `r_sexp_next()` (`walk.c:119-120`) obtains `p_info = r_dyn_last(p_stack)` — a raw pointer into the dyn-array's data buffer — and passes it to `sexp_next_incoming()`.
- The child push (`walk.c:221`) goes through `r__dyn_increment()` (`dyn-array.h:149`), which past capacity calls `r_dyn_resize()` (`dyn-array.c:78`): a **new** vector is allocated, poked into the shelter in place of the old one (leaving the old buffer unprotected), and `p_arr->data`/`v_data` are updated.
- The parent's state bump then writes through the stale pointer (`walk.c:226-239`: `++p_info->v_arr` / `++p_info->p_state` and the incoming→outgoing flip), i.e. into the dead buffer.

Consequences: the copied entry in the live stack keeps its pre-bump state, so when that parent resurfaces the iterator revisits the same edge — the duplicate visits above. It is also a latent use-after-free: the old buffer is unprotected while still being written (currently benign only because no allocation happens between the resize and the writes).

## Suggested fix

Bump the parent's state *before* pushing the child (the bump only touches `p_info`, and all reads of `p_info` used to build `child` happen earlier), so the updated state is what gets copied on resize. With that reorder, depth 300 yields exactly 601 visits and 0 duplicates. Alternatively, re-fetch `p_info` from the stack after the push. Happy to submit a PR; whether to modernize the non-API accessors is a separate question.

kevinushey:
Fixes #1911.

`sexp_next_incoming()` (`src/rlang/walk.c`) receives `p_info` as a raw pointer into the traversal stack's dyn-array buffer. The child push could resize the stack — reallocating the buffer and dropping the old one's protection — after which the parent's state bump and incoming→outgoing flip were written through the stale pointer into the dead buffer. The live (copied) entry kept its pre-bump state, so the iterator revisited the same edge and re-traversed entire subtrees once depth exceeded the initial stack capacity of 256. It was also a latent use-after-free write, benign today only because no allocation happens between the resize and the writes.

This PR moves the state bump before the push. All reads of `p_info` used to build `child` happen earlier, so the reorder is behavior-preserving apart from the fix.

Since this code is only compiled under `RLANG_USE_PRIVATE_ACCESSORS` (and currently doesn't build on R >= 4.5, see the issue), there's no regression test in the default configuration. Verified in a flagged dev build with accessor stubs: before the fix, a depth-300 nested list yields 690 `sexp_iterate()` callback visits with 44 nodes visited twice as incoming; after the fix, exactly 601 visits (2n + 1) with no duplicates, and the depth-250 control is unchanged.

### DIFF
--- a/src/rlang/walk.c
+++ b/src/rlang/walk.c
@@ -209,19 +209,9 @@ static bool sexp_next_incoming(
     bool has_attrib = sexp_has_attrib(child.type, child.x);
     enum sexp_iterator_type it_type = sexp_iterator_type(child.type, child.x);
 
-    if (it_type == SEXP_ITERATOR_TYPE_atomic && !has_attrib) {
-        child.p_state = NULL;
-        child.dir = R_SEXP_IT_DIRECTION_leaf;
-    } else {
-        init_incoming_stack_info(&child, it_type, has_attrib);
-
-        // Push incoming node on the stack so it can be visited again,
-        // either to descend its children or to visit it again on the
-        // outgoing trip
-        r_dyn_push_back(p_it->p_stack, &child);
-    }
-
-    // Bump state for next iteration
+    // Bump state for next iteration. This must be done before pushing
+    // `child` on the stack: the push may resize the stack and `p_info`
+    // points into the pre-resize buffer.
     if (state == SEXP_ITERATOR_STATE_elt) {
         ++p_info->v_arr;
         if (p_info->v_arr == p_info->v_arr_end) {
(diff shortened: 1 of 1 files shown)


lionel-:
Thanks!

PR #1917, opened 2026-08-27
Delete dead tests

MichaelChirico:
These have been here untouched for 8+ years, and were marked 'dead' 5+ years ago c1387559faca2189f89ce127579baf2753a86e73 e2686134713cf54457bfd67732bff7f730d115de

Nevertheless CRAN is still getting a zip file, which is a bit unusual: https://github.com/cran/rlang/blob/92706ea9bbea75e0cac1ef41b5e4e5d943a1aae3/tests/testthat/fixtures/lib.zip

The 40KiB here is about 5% of the total size; if these are still needed, I recommend at least hiding them from CRAN in .Rbuildignore.


lionel-:
Thank you!

PR #1918, opened 2026-09-04
stats is Imports, not Suggests

MichaelChirico:
https://github.com/r-lib/rlang/blob/d0e9d81baa937ba7a4cf981dd027229b0229c096/NAMESPACE#L545-L548


lionel-:
Thanks!

